PR #3, opened 2020-08-24
Configure Terraform S3 backend

jakemulley:
This PR:

- Defines some local variables for tagging
- Defines an S3 bucket to use for a Terraform backend
- Defines the Terraform backend

This allows remote state sharing in Terraform to ensure that local references are up-to-date and that `terraform apply` doesn't create a conflicting state set if run at the same time by different people.

### DIFF
--- a/terraform/global-resources/locals.tf
+++ b/terraform/global-resources/locals.tf
@@ -1,0 +1,9 @@
+locals {
+  global_resources {
+    application-name    = "Modernisation Platform - Global Resources"
+    business-unit       = "Ministry of Justice - Hosting"
+    tech-contact-email  = "modernisation-platform@digital.justice.gov.uk"
+    budget-holder-email = "steve.marshall@digital.justice.gov.uk"
+    stage               = "production"
+  }
+}
--- a/terraform/global-resources/locals.tf
+++ b/terraform/global-resources/locals.tf
@@ -1,9 +1,8 @@
 locals {
   global_resources {
-    application-name    = "Modernisation Platform - Global Resources"
-    business-unit       = "Ministry of Justice - Hosting"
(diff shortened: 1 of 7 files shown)


terraform/global-resources/locals.tf
```diff
@@ -0,0 +1,9 @@
+locals {
+  global_resources {
+    application-name    = "Modernisation Platform - Global Resources"
+    business-unit       = "Ministry of Justice - Hosting"
+    tech-contact-email  = "modernisation-platform@digital.justice.gov.uk"
+    budget-holder-email = "steve.marshall@digital.justice.gov.uk"
+    stage               = "production"
+  }
+}
```
SteveMarshall, line 9: Some of these look like things that would be covered by [our tagging guidance](https://ministryofjustice.github.io/technical-guidance/documentation/standards/documenting-infrastructure-owners.html).

terraform/global-resources/locals.tf
```diff
@@ -0,0 +1,9 @@
+locals {
```
davidkelliott, line 1: MoJ have tagging guidelines that we should follow - https://ministryofjustice.github.io/technical-guidance/documentation/standards/documenting-infrastructure-owners.html#tags-you-should-use

davidkelliott (commented):
Looks good, I didn't realise you can set up the S3 bucket that you are using for the state?  Will it not give an error trying to access the state on the first run? My only comment is to update the tags to match the MoJ tagging guidelines. 👍

SteveMarshall (changes requested):
😎

One meta thing I'd like to see is that the PR/commit talk more about the _why_: I can see that it does all the things you list, but why do we need those? (Also, “Set up”, not “Setup” 😬)

SteveMarshall (approved):

PR #5, opened 2020-09-01
Fix syntax error with local variables

jakemulley:
This PR:

- Fixes a syntax issue for defining and using local variables


davidkelliott (approved):
LGTM

PR #6, opened 2020-09-02
Add a password policy for IAM users

jakemulley:
This PR:
- Creates a password policy for IAM users

This forces:
- 30 day rotation periods
- minimum 12-character passwords
- lowercase and uppercase characters, numbers, symbols
- prevents reusing any of the 5 previously used passwords

It also allows a user to change their own password without enforcing an administrator reset.

### DIFF
--- a/terraform/global-resources/iam.tf
+++ b/terraform/global-resources/iam.tf
@@ -1,0 +1,12 @@
+# Set an IAM account password policy to ensure rotation of passwords, with some sensible defaults
+resource "aws_iam_account_password_policy" "strict" {
+  allow_users_to_change_password = true
+  hard_expiry                    = false
+  max_password_age               = 30
+  minimum_password_length        = 12
+  password_reuse_prevention      = 5
+  require_lowercase_characters   = true
+  require_numbers                = true
+  require_symbols                = true
+  require_uppercase_characters   = true
+}


davidkelliott (approved):
Generally I'm not a fan of forcing password resets but if this is purely for the admin users and we are all using a password manager it shouldn't be an issue.  I thought we had some password guidance somewhere but I can't find it now :/ have you seen the AWS baseline requirements - https://ministryofjustice.github.io/security-guidance/baseline-aws-accounts/#identity--access-management
We'll need to have MFA as well but can be done in a later PR.

jakemulley:
> Generally I'm not a fan of forcing password resets but if this is purely for the admin users and we are all using a password manager it shouldn't be an issue. I thought we had some password guidance somewhere but I can't find it now :/ have you seen the AWS baseline requirements - https://ministryofjustice.github.io/security-guidance/baseline-aws-accounts/#identity--access-management
> We'll need to have MFA as well but can be done in a later PR.

@davidkelliott If we do decide to use traditional IAM users as well as or rather than federated IAM users, I thought this would be good to have in place before it gets forgotten about or overlooked in the future, regardless of if it's actually used at all.

I'll align this with the password standard, thanks for letting me know!

As MFA involves an IAM policy and is a bit more involved, I thought I'd separate that out into a separate PR 👍

davidkelliott:
Posting the link to the guidance you found here in case I lose it again - https://intranet.justice.gov.uk/guidance/security/it-computer-security/ict-security-policy-framework/password-standard/

PR #7, opened 2020-09-03
Use superadmin Terraform module

jakemulley:
This PR:
- Moves to using the superadmin Terraform module to allow superadmin access to accounts
- Devolves password policy management to the Terraform module

This allows us to reuse the superadmin and password management policy across accounts.


davidkelliott (approved):
Looks good, chatted through with Jake 👍

PR #21, opened 2020-09-15
Add a public access block and enable S3 encryption

jakemulley:
This PR:
- Encrypts the `terraform.tfstate` file stored in S3
- Encrypts new S3 objects in the bucket
- Explicitly blocks public access to the S3 bucket

This is to ensure our S3 bucket is as secure as possible.


davidkelliott (approved):
Discussed pros and cons of SSE-S3 vs SSE-KMS with Jake 👍

PR #22, opened 2020-09-15
Add an ADR for IAM Federated Access

jakemulley:
This PR:
- Adds a proposed ADR for the use of IAM Federated Access


davidkelliott (approved):

davidkelliott:
Happy that we have a working POC, we can always revisit when we decide the account structure and how that will work if this doesn't fit in with that. It's good to align with CP 👍

jakemulley:
Sure, I'll merge it in as proposed so we can reject it if needed 👍🏼

PR #23, opened 2020-09-25
Allow automated environment creation from a JSON file

### LINKED ISSUE #11: Decide on the approach of a multi-account management structure
We need to decide on the approach of a multi-account management structure.

We've previously spoken about:
- AWS Organisations and organisation accounts
- Landing Zones
- Control Tower
...and more.

We should record a decision in the ADR.

jakemulley:
This PR resolves #11, and is reliant on ministryofjustice/aws-root-account#5.

This PR does the following manually:
- Creates two organisational units for `non-production` and `production` application environments

This PR automates the following:
- The creation of child organisation units in either `non-production` and `production` named after applications defined in the `environments` folder
- The creation of attached OU accounts for applications in their respective child organisation units, with some sensible defaults such as allowing access to billing, a standardised email address, and tags.

It does this by assuming a role as defined in ministryofjustice/aws-root-account#5.

### DIFF
--- a/environments/non-production/shared-services.json
+++ b/environments/non-production/shared-services.json
@@ -1,0 +1,10 @@
+{
+  "name": "shared-services",
+  "environments": ["logs", "security"],
+  "tags": {
+    "business-unit": "Platforms",
+    "application": "modernisation-platform-shared-services",
+    "is-production": false,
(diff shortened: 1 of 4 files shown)


jakemulley:
Closing, as I'll move this to a module.

PR #24, opened 2020-09-25
Add environment tag checking with a sample environment

jakemulley:
This PR:
- Adds some sample data to this repository
- Adds a script that will warn about missing key value pairs that are needed to create an OU and account for an application to sit on the Modernisation Platform, and will throw an error if required tags from the MoJ Technical Guidance are missing
- Creates a GitHub workflow to run the script on new PRs, and comment the outcome

It has been extrapolated from ministryofjustice/modernisation-platform-environments#1.


jakemulley:
Marked this as draft whilst I sort out the message formatting.

PR #28, opened 2020-09-29
Move environments management to a separate Terraform state

jakemulley:
As our environments grow, `terraform plan` and `apply` will take longer, so this PR moves environments management to a separate Terraform state. This also allows us a separation of concerns between setting up environments dynamically, and managing other resources that are part of the Modernisation Platform.


davidkelliott (approved):
👍

PR #33, opened 2020-10-06
Add a provisioner to generate providers for each region enabled in our AWS account

jakemulley:
This adds a way to automatically generate provider aliases based on what regions are enabled in the Modernisation Platform account.


jakemulley:
I've moved this into a separate module at ministryofjustice/modernisation-platform-terraform-baselines#1.

PR #34, opened 2020-10-06
Enable SecurityHub in all regions with all standards

jakemulley:
Requires ministryofjustice/modernisation-platform#33.

This uses AWS SSM to [query what AWS regions are available for services](https://aws.amazon.com/blogs/aws/new-query-for-aws-regions-endpoints-and-more-using-aws-systems-manager-parameter-store/), which can be reused for services that should be enabled in each region.

This automatically generates `securityhub.tf`, which by using AWS SSM, allows us to automatically generate (and therefore enable) SecurityHub in all regions it's available, alongside the default standard subscriptions for: AWS Foundational Security Best Practices, CIS AWS Foundations, PCI DSS v3.2.1.

This only enables them for the Modernisation Platform account as this will be the "root" reporting account.


jakemulley:
I've moved this into a separate module at ministryofjustice/modernisation-platform-terraform-baselines#2.

PR #35, opened 2020-10-06
Enable GuardDuty in all regions

jakemulley:
Requires ministryofjustice/modernisation-platform#34.

This enables GuardDuty in all available regions.


jakemulley:
I've moved this into a separate module at ministryofjustice/modernisation-platform-terraform-baselines#3.

PR #37, opened 2020-10-07
Implement a fix for the cyclical dependency on a generated file

jakemulley:
- Add a Makefile to download generated files
- Update README and fix a link
- Stop Terraform from managing the local version of the generated file

As Terraform doesn't support for_each for providers, we generate static blocks via a template file instead. If you're getting started on the MP, it will throw an error because you don't have that file. This fixes it by providing a way to download the file without using Terraform.


davidkelliott (approved):
Nice one 👍

jakemulley:
Closing in favour of #38.

PR #38, opened 2020-10-07
Fix the cyclical dependency on generated files

jakemulley:
This PR supersedes #37.

Instead of having to download the file from S3, we can store the account IDs generated via the environments module in AWS Secrets Manager, and use variables in `providers-generated.tf` to assume a role within the account. The secret will now update with new account IDs whenever a new account is created.

It removes the need to use S3 for this file, and we should commit it every time it changes.


ewastempel (approved):
LGTM

PR #39, opened 2020-10-08
Add the dynamically generated baseline files

jakemulley:
Requires ministryofjustice/modernisation-platform-terraform-baselines#5.

This commits the dynamically generated baseline files into a subfolder, which is invoked as a module.


ewastempel (approved):
LGTM

PR #40, opened 2020-10-08
Pin IAM to a hash due to change to AWS SSO

jakemulley:
This PR pins the Auth0 module to the working hash whilst we move toward AWS SSO.


ewastempel (approved):
LGTM

PR #41, opened 2020-10-08
Enable baselines for accounts

jakemulley:
This PR generates and creates the resources from the [ministryofjustice/modernisation-platform-terraform-baselines](https://github.com/ministryofjustice/modernisation-platform-terraform-baselines) module, for each account that is configured through the Modernisation Platform.


ewastempel (approved):
I can view security services in an assumable role. Looks good!

PR #42, opened 2020-10-12
Tag GuardDuty resources and update baseline module output

jakemulley:
This tags GuardDuty resources and updates the baseline module output to fix some spacing issues.


SteveMarshall (approved):
✅


But only for `FIFTEEN_MINUTES` 😂

PR #44, opened 2020-10-13
AWS Config remediation: tag default VPC and update S3 bucket permissions

jakemulley:
This resolves the VPC configuration being non-compliant in AWS Config due to not having tags, and tightens up public restrictions on the Terraform state S3 bucket.


davidkelliott (approved):
👍

PR #46, opened 2020-10-14
Require SSL for the modernisation-platform-terraform-state bucket

jakemulley:
This resolves an AWS Config non-compliancy on S3 buckets requiring SSL.


davidkelliott (approved):
👍

PR #48, opened 2020-10-16
Remove deprecated files and configuration from the old baselines module

jakemulley:
This brings `global-resources` in line with the with the new, non-generated files provided by the [modernisation-platform-terraform-baselines](https://github.com/ministryofjustice/modernisation-platform-terraform-baselines) module.


terraform/templates/providers.tmpl
```diff
@@ -0,0 +1,26 @@
...
+}
+
+%{ endfor ~}
+
+module "baselines" {
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-baselines"
+  providers = {
+    aws                = aws
+    %{~ for region in account_regions ~}
+    aws.${region} ${substr("              ", 0, 14-length(region))}= aws.${provider_key}-${region}
```
jakemulley, line 21: This spacing substr is to output the generated file with the correct spacing, so `terraform fmt` doesn't try to reformat the file every time, and also so `terraform plan` doesn't output that the file needs updating to revert `terraform fmt` each time.

davidkelliott (approved):
Looks good

PR #50, opened 2020-10-21
Set aws provider to use eu-west-2

jakemulley:
This explicitly sets the `aws` default provider to use the Modernisation Platform `eu-west-2` provider, rather than relying on the caller's `AWS_REGION`.


davidkelliott (approved):
👍

PR #51, opened 2020-10-21
Add a proposed ADR for AWS SSO

jakemulley:
This is a proposed ADR for AWS SSO so we can record our decision on access management.


ewastempel (approved):
LGTM

PR #52, opened 2020-10-21
Remove default VPC configuration

jakemulley:
This removes default VPC tagging and configuration which is now done in [modernisation-platform-terraform-baselines](https://github.com/ministryofjustice/modernisation-platform-terraform-baselines/pull/27) so we can centrally manage it.

It also renames the file as `defaults.tf` wasn't very descriptive.

Requires ministryofjustice/modernisation-platform-terraform-baselines#27.


ewastempel (approved):
LGTM

PR #53, opened 2020-10-21
Move IAM assumable role to a module

jakemulley:
This moves an IAM assumable role definition for the Modernisation Platform to assume a role in an account to a Terraform module, and removes redundant files to be in line with how baselines are created in an account.

It also uses `baselines-${provider_key}` by default to name the baseline module to keep the definition unique.

Requires ministryofjustice/modernisation-platform-terraform-cross-account-access#1.


ewastempel (approved):
LGTM

PR #54, opened 2020-10-21
Update environments README

jakemulley:
This removes redundant information and makes it clearer as to how provider files are generated and what the providers files configure.


ewastempel (approved):
LGTM

PR #55, opened 2020-10-22
Restrict bucket policy for terraform-state

jakemulley:
This restricts the bucket policy for `modernisation-platform-terraform-state` to reduce the actions that can be performed by the root account, to only those [required by Terraform](https://www.terraform.io/docs/backends/types/s3.html#s3-bucket-permissions).


ewastempel (approved):
LGTM

PR #57, opened 2020-10-22
Add missing tags to Route53 hosted zone

jakemulley:
This adds the missing tags to the Route53 Hosted Zone.


ewastempel (approved):
tagged it :)

PR #58, opened 2020-10-22
Add local provider to versions.tf for Terraform 0.13

jakemulley:
This adds the `local` provider to `versions.tf`, as recommended by Terraform 0.13.


ewastempel (approved):
fine

PR #60, opened 2020-10-28
Bring providers.tf file up to date

jakemulley:
Brings the `providers.tf` file up to date with what's generated in the template.


ewastempel (approved):
cool

PR #61, opened 2020-10-28
Add DynamoDB state locking for Terraform

jakemulley:
This configures `terraform` state locking via DynamoDB.


ewastempel (approved):
cool

PR #62, opened 2020-10-29
Specify replication_region for S3 buckets

jakemulley:
This sets the region for S3 buckets to be replicated into as `eu-west-1`.


ewastempel (approved):
LGTM

PR #74, opened 2020-11-03
Update required Terraform and Terraform providers versions

jakemulley:
This specifies that the minimum Terraform version required is `0.13.5` and also specifies to use the following provider versions:

- `aws` to use at least `3.13.0`
- `local` to use at least `2.0.0`


zuriguardiola (approved):
lgtm

Issue #75, opened 2020-11-03
Write a Lambda function to refresh and report on AWS Trusted Advisor


jakemulley:
We should utilise AWS Trusted Advisor to provide us with some monitoring for the following, across other AWS services:

- Cost optimisation
- Performance
- Security
- Fault tolerance
- Service limits

Trusted Advisor allows you to refresh metrics via the console and Support API to get up-to-date information. We should look at sending this information to CloudWatch, Slack, etc.

jakemulley:
I've setup a [repository](https://github.com/ministryofjustice/modernisation-platform-terraform-trusted-advisor) as we should create this as a reusable module across accounts.

jakemulley:
I've got a first draft/initial implementation of this over in ministryofjustice/modernisation-platform-terraform-trusted-advisor#1.

PR #80, opened 2020-11-04
Add Trusted Advisor refresh Lambda

jakemulley:
This is the first part of #75, which refreshes AWS Trusted Advisor every 60 minutes to get the most up-to-date check results.

### DIFF
--- a/terraform/global-resources/providers.tf
+++ b/terraform/global-resources/providers.tf
@@ -110,3 +110,7 @@ module "baselines-modernisation-platform" {
   root_account_id = local.root_account.master_account_id
   tags            = local.global_resources
 }
+
+module "trusted-advisor-modernisation-platform" {
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-trusted-advisor"
+}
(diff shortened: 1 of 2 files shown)


davidkelliott (approved):
Looks good to me, the only thing I would ask is do we know if there are a limit on the number of refreshes per account / org?

jakemulley:
> Looks good to me, the only thing I would ask is do we know if there are a limit on the number of refreshes per account / org?

@davidkelliott There's nothing noted in the documentation or console as far as I can see, apart from the minimum time between refreshes is _mostly_ 60 minutes. Some are longer, but they don't return an error if you try to refresh them before the time interval.

Just a note that Trusted Advisor is _free at point of use_ though it requires a Premium Support plan.

davidkelliott:
> > Looks good to me, the only thing I would ask is do we know if there are a limit on the number of refreshes per account / org?
> 
> @davidkelliott There's nothing noted in the documentation or console as far as I can see, apart from the minimum time between refreshes is _mostly_ 60 minutes. Some are longer, but they don't return an error if you try to refresh them before the time interval.
> 
> Just a note that Trusted Advisor is _free at point of use_ though it requires a Premium Support plan.

Thanks! Thought I'd ask as I couldn't find anything either 👍

PR #117, opened 2020-11-19
ADR for using GitHub Actions as our CI/CD runner

### LINKED ISSUE #12: Discuss tooling for CI/CD
We should define what tooling we'll be using for CI/CD and any applicable setup. We've previously talked and thought about:
- GitHub Actions
- CircleCI
- Concourse
...and other CI/CD providers.

We also need to think if we're going for a self-hosted approach or not. We should record a decision in the ADR.

jakemulley:
ADR for using GitHub Actions as our CI/CD runner, resolves #12


davidkelliott (approved):
Sounds good to me 👍

PR #138, opened 2020-11-20
Rename Modernisation Platform core to core

jakemulley:
Resolves an issue where the concat of the application name and environment names can't exceed 36 characters.


philhorrocks (approved):
all good

PR #150, opened 2020-11-23
Implement local and remote workspace creation for applications

jakemulley:
This creates directories and the base Terraform configuration for all applications individually, with workspaces for each environment stored in S3, with some sample data output.


philhorrocks (dismissed):
approved

davidkelliott (approved):
Looks good 👍

PR #155, opened 2020-11-24
Split out core accounts and rename shared-services-dev to sandbox

jakemulley:
This splits out the Modernisation Platform core accounts into separate environment files, and renames shared-services-dev to core-sandbox for reuse.

This required some pretty hefty state surgery for `terraform workspaces` which doesn't have anything in code, that is committable, to show for it, but it was generally moving the Terraform state between workspaces.


zuriguardiola (approved):

philhorrocks (approved):
approved

Issue #163, opened 2020-12-01
Extend JSON files to include ranges for CIDR allocations


jakemulley:
We need to extend our environment definition JSON files to be able to include ranges for CIDR allocations.

PR #165, opened 2020-12-01
Remove name key in env definitions and update create_accounts to use filename instead

jakemulley:
Removes the name key from environment definitions in favour of using the filename, and updates the `create-accounts.sh` script to use the input_filename provided by `jq`.

This is in preparation for #163.


zuriguardiola (dismissed):

zuriguardiola (approved):

jakemulley:
Also includes a fix for `undefined` in the [comment](https://github.com/ministryofjustice/modernisation-platform/pull/165#issuecomment-736509398) from check-environment-definitions.js

PR #167, opened 2020-12-01
Fix create-per-account-local-workspaces.sh by removing reliance on .name

jakemulley:
Fixes an issue resulting from #165, in preparation for #163.


donmasters (approved):
LGTM

PR #168, opened 2020-12-01
Add a team script for an easier way to see differences between bootstrapped environments

jakemulley:
This script will `cd` into the `terraform/environments/bootstrap` folder, get a list of workspaces, and run `terraform plan` in all of them synchronously.

Previously you would have had to do this manually for each terraform workspace, whereas this does it for you so you can start running it and leave it until it's finished. It outputs something similar to the below in your shell:

```Running terraform init for prepare_terraform, in the bootstrap directory
Running terraform validate for prepare_terraform, in the bootstrap directory
Switched to workspace "default".
Switched to workspace "bichard7-dev-current".
Running terraform refresh for bichard7-dev-current (this may take a while)
Running terraform plan for bichard7-dev-current (this may take a while)

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.

Switched to workspace "bichard7-dev-next".
Running terraform refresh for bichard7-dev-next (this may take a while)
```
etc.

This doesn't actually apply any changes, and shouldn't be run as part of a CI/CD process as the `terraform plan` may contain secrets that are output to shell.

### DIFF
--- a/scripts/internal/loop-through-bootstrap-workspaces-and-plan.sh
+++ b/scripts/internal/loop-through-bootstrap-workspaces-and-plan.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+prepare_terraform () {
+  # We can buffer output (stdout and stderr) unless terraform exits with a non-zero code, so we don't unneccessarily
+  # log the output of these commands in our CI/CD runner
+  echo "Running terraform init for $called_function"
+  output=$(terraform init -input=false 2>&1) || (echo "$output" && false)
+  echo "Running terraform validate for $called_function"
+  output=$(terraform validate -no-color 2>&1) || (echo "$output" && false)
+}
+
+run_terraform_plan () {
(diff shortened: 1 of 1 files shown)


davidkelliott (approved):
Looks good, quick question about language choice for my own learning:
Why did you choose to name the variable `called_function` rather than something like `calling_function` ?  Is this a convention? It just seems more logical to me to pass the name of the function that called it to the called function, rather than it's own name?  Just curious!

jakemulley:
> Looks good, quick question about language choice for my own learning:
> Why did you choose to name the variable `called_function` rather than something like `calling_function` ? Is this a convention? It just seems more logical to me to pass the name of the function that called it to the called function, rather than it's own name? Just curious!

Good spot - using `called_function` was totally arbitrary and I should change it to `calling_function` as it makes more sense - I'll submit a separate PR for those changes 👍🏼

PR #174, opened 2020-12-04
Clarification of application support item: remove solutions

jakemulley:
Broadens the scope by removing the extreme specificity of an example thing we wouldn't support as mentioned by @SteveMarshall in the previous PR.


SteveMarshall (approved):
😍

PR #207, opened 2021-01-08
Move the RAM principal association into the EC2 TGW attachment module

jakemulley:
This PR moves the RAM principal association for our shared Transit Gateway into our attachment module.

Marked as draft as it requires a `terraform state mv`.


donmasters (approved):
LGTM

PR #212, opened 2021-01-12
Fixup core-network-services to use new VPC module model

jakemulley:
The core-vpc module model has changed to deprecate the need to manually work out CIDR block ranges for subnets, as we can generate them using Terraform's `cidrsubnets` function, which is part of the core-vpc module.

This PR brings `core-network-services-production` into line with the new model in `core-vpc`.


donmasters (approved):
LGTM

PR #213, opened 2021-01-12
Enable VPC Flow Logs for the VPC modules

jakemulley:
Marked as draft as it requires #212.

Creates a CloudWatch Log Group for VPC Flow Logs and enables VPC Flow Logging for core-network-services, core-shared-services, core-vpc accounts.


donmasters (approved):
LGTM

PR #227, opened 2021-01-14
Associate subnets with a RAM resource share

jakemulley:
This PR creates a RAM resource share for each VPC set up in `core-vpc`, and associates the non-Transit Gateway and non-protected subnets to it, ready for sharing with a principal.

This PR also creates principal associations with the resource share, and retags the VPC and subnets in shared accounts.


donmasters (approved):
LGTM

PR #229, opened 2021-01-18
Update tech-docs publishing to new docker image & remove superfluous files

jakemulley:
This PR updates how tech-docs are published and removes Ruby-specific files that are on the tech-docs Docker image.


donmasters (approved):
LGTM

PR #234, opened 2021-01-19
Split out bootstrap steps: delegate access, and baselines

jakemulley:
This PR splits out `delegate-access` (to create the `ModernisationPlatformAccess` IAM role in an AWS account) and `secure-baselines` into separate directories, so they can be run independently of each other.

The reason for doing this is that `delegate-access` requires privileged permissions at the AWS Organizations level, and once that access is given, all other bootstrap steps (or Terraform actions) can be run through a different role (`ModernisationPlatformAccess`).

This PR also explicitly differentiates between `terraform workspace` creation, specifically for the bootstrap steps, and removes the execution of `terraform apply` in the same step. This will allow us to generalise a `terraform init/plan/apply` script to do two things: loop over `terraform workspaces`, and run `init/plan/apply` as separate scripts where required.

This PR will still create environments (AWS accounts) when definitions are merged into `./environments` but will no longer automatically delegate access or run the bootstrapping scripts. The next PR will re-introduce this functionality.


donmasters (dismissed):
LGTM

donmasters (approved):
LGTM

PR #239, opened 2021-01-21
Fix syntax issues with provision-terraform-workspaces.sh

jakemulley:
This fixes a few syntax issues in provision-terraform-workspaces.sh. It now:

- checks for `tmp/` before trying to delete it
- uses `sed` with multiple replacements and `sort -u -o` to output into a file, rather than piping `cat` into `grep`, `tr`, `tee`
- replaces the incorrect syntax for comparators in bash from `==` to `=`
- quotes the main $1 argument to stop whitespace breaking it
- uses `-z` for the null value comparator to check if `$1` is set
- reorders the `find` arguments as `mindepth` and `maxdepth` are not positional, which suppresses this warning:

```bash
find: warning: you have specified the -mindepth option after a non-option argument -type, but options are not positional (-mindepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments
```

### DIFF
--- a/scripts/provision-terraform-workspaces.sh
+++ b/scripts/provision-terraform-workspaces.sh
@@ -4,7 +4,7 @@
 #
 # You need to pass through an argument to this script:
 #
-# bootstrap (` sh scripts/provision-terraform-workspaces.sh bootstrap`)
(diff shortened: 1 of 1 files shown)


donmasters (approved):
LGTM

PR #245, opened 2021-01-21
ANPL-252 Add AWS account for analytical platform management tools

andylightfoot:
Add a new AWS account for the Analytical Platform to use for management tools.


donmasters (approved):
LGTM

PR #246, opened 2021-01-21
Add base Terraform scripts for workflows

jakemulley:
This adds some base scripts to run Terraform as part of a CI/CD pipeline.

`plan` and `apply` are piped to `redact-output.sh`, to redact `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID` if exposed, alongside IDs that may be sensitive.


donmasters (approved):
LGTM

PR #247, opened 2021-01-21
Add script to loop through Terraform workspaces and run plan, or apply

jakemulley:
This script runs Terraform in a specified directory and runs either `terraform plan` or `terraform apply` across all workspaces configured within the backend in the directory.

It's a prerequisite for our GitHub workflow that will create new environments, delegate access to the Modernisation Platform, and configure secure baselines within all of the environments.

Requires #246.


donmasters (approved):
LGTM

PR #258, opened 2021-01-22
Separate template files into logical files

jakemulley:
This PR splits Terraform into different, more logical files, rather than just `backend.tf`. It also updates `provision-environment-directories.sh` to use a `$`-prefixed string for replacements, so it's clearer they're variables (in this case, `$application_name` rather than `application_name`.

Instead of `backend.tf`, this splits out:

`provider` definitions into `providers.tf`
`locals` definitions into `locals.tf`

This makes it easier to namespace files into their respective resource within `terraform/environments/*`, such as `vpc.tf` for VPC-related things, allowing you to remove the file to destroy everything related to a VPC, for instance.


zuriguardiola (approved):

donmasters (approved):
LGTM

PR #277, opened 2021-01-28
Transit Gateway attachments: correctly set providers and allow tagging of attachments

jakemulley:
This PR corrects some provider declarations for data sources to enable you to call it from a different account to what it needs to be run in.

It also introduces a `vpc_name` variable. This allows you to tag Transit Gateway attachments to appear as `${vpc-name}-attachment`, rather than the previously set `shared-transit-gateway-attachment`.

It means our Transit Gateway attachments in `core-vpc-production` list as:
`hmpps-production-attachment`, and
`laa-production-attachment`


donmasters (approved):
LGTM

PR #287, opened 2021-02-01
Rename shared-transit-gateway to transit-gateway

jakemulley:
This renames references to `shared-transit-gateway` (which is shared via Resource Access Manager) to `transit-gateway`, as it's inferred that it's shared as it's part of a resource share.


zuriguardiola (approved):

donmasters (approved):
LGTM

PR #288, opened 2021-02-01
Expand acronyms and add missing tags

jakemulley:
This expands acronyms from:
- IG -> Internet Gateway
- TGW -> Transit Gateway
and  rekeys the Internet Gateway to `default` (rather than `ig`, this module only configures 1).

It adds some missing tags on `aws_cloudwatch_log_group`, and updates the default NACLs to be use `=` rather than `:` as key, value separators to match Terraform and show the correct in-file syntax highlighting for IDEs.


zuriguardiola (approved):

donmasters (approved):
LGTM

PR #299, opened 2021-02-04
Update required providers to always specify a minimum AWS provider and Terraform version

jakemulley:
This PR updates the templates directory to use a minimum version for:

- Terraform itself, with a minimum version of v0.14.2
- the AWS provider, with a minimum version of v3.20.0

This is to ensure older Terraform versions and AWS providers aren't used, which may cause issues with our configuration. For instance, Terraform 0.14.0 [introduced a bug](https://github.com/hashicorp/terraform/blob/v0.14/CHANGELOG.md#0141-december-08-2020) where ignore_changes doesn't appear to work, forcing some resources to be created, which was fixed in Terraform 0.14.1. Terraform 0.14.1 [also introduced a bug](https://github.com/hashicorp/terraform/blob/v0.14/CHANGELOG.md#0142-december-08-2020) where remote backends were checked for compatibility when they were only used for read rather than write.

By providing an explicit AWS provider version, this change also resolves a bug whereby if you're using the [plugin_cache](https://www.terraform.io/docs/cli/config/config-file.html#provider-plugin-cache) for providers, and you hadn't installed a AWS provider from version 3 yet, it'd continue to use your version 2 provider.

Finally, this PR also removes some unneeded required_provider declarations, where the required_provider is actually inferred from the child module using the provider, rather than the root module.

### DIFF
--- a/terraform/environments/analytical-platform-management/versions.tf
+++ b/terraform/environments/analytical-platform-management/versions.tf
@@ -1,8 +1,9 @@
 terraform {
   required_providers {
     aws = {
-      source = "hashicorp/aws"
+      version = ">= 3.20.0"
+      source  = "hashicorp/aws"
     }
   }
-  required_version = ">= 0.14.0"
+  required_version = ">= 0.14.2"
 }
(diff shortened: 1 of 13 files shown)


zuriguardiola (approved):
lgtm

PR #300, opened 2021-02-04
Cleanup VPC Transit Gateway routing module

jakemulley:
This PR cleans up the VPC TGW routing module:

- Removes the need to specify both a default and aliases for the provider as the module only uses one provider (core-network-services), so you can use the module's default provider, mapped to the correct root provider
- adds suitable descriptions and types to variables
- adds a versions.tf with a minimum Terraform and AWS provider version


zuriguardiola (approved):
lgtm

PR #317, opened 2021-02-05
Split out backend and providers declarations, import orphaned repository resource

jakemulley:
This PR:

- splits out the `backend` and `provider` declarations into separate files, to mimic other parts of this repository
- imports the `modernisation-platform-cidr-allocation` repository which was manually created, so we can manage it here
- fixes version declarations to use Terraform 0.14.2 or above, and the correct minimum provider version for GitHub
- adds a `visibility` variable for the `modernisation-platform-cidr-allocation` repository to be internal


seanprivett (approved):
Pair reviewed with Jake

PR #322, opened 2021-02-05
Add validation testing for core-vpc json files

donmasters:
- json validation file created
- created script to check core-vpc json files against
  validation json file
- github actions job created to trigger on PR for specific json file
  changes

### DIFF
--- a/.github/workflows/validate-core-vpc-configuration.yml
+++ b/.github/workflows/validate-core-vpc-configuration.yml
@@ -1,0 +1,21 @@
+name: "Validate core-vpc configuration"
+
+on: 
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'scripts/tests/validate/fixtures/pre-terraform-core-vpc-validation.json'
+      - 'environments-networks/**.json'
+
+jobs:
+  validate-core-vpc-config-files:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/validate-core-vpc-configuration.yml
```diff
@@ -0,0 +1,21 @@
+name: "Validate core-vpc configuration"
+
+on: 
+  pull_request:
+    branches:
+      - master
```
jakemulley, line 6: This should be `main` 👍🏼

.github/workflows/validate-core-vpc-configuration.yml
```diff
@@ -0,0 +1,21 @@
+name: "Validate core-vpc configuration"
+
+on:
```
jakemulley, line 3: Is it worth adding a `workflow_dispatch` trigger ([example](https://github.com/ministryofjustice/modernisation-platform/blob/main/.github/workflows/check-links.yml#L4)) so we can run this manually via GitHub actions?

jakemulley (changes requested):

.github/workflows/validate-core-vpc-configuration.yml
donmasters: added

.github/workflows/validate-core-vpc-configuration.yml
donmasters: updated to main

jakemulley (approved):

PR #341, opened 2021-02-10
Add workflow to deploy GitHub resource changes via Terraform

### LINKED ISSUE #88: Automate terraform init, plan for GitHub resources
We should automate `terraform init` and `terraform plan` when a PR is opened that changes the GitHub resources as configured in `terraform/github`. We should add the output as a PR comment so we know what changes are pending if the PR is merged.

To run `plan`, we need to set a [PAT](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) that has the scope for organisation and repository management within the Ministry of Justice organisation on GitHub.

For this, we should use a dedicated GitHub user that is part of the organisation and has the right scope set, as the preset `GITHUB_TOKEN` in GitHub Actions only has the scope for the repository it is called in.

jakemulley:
This PR creates a GitHub workflow that does the following:

- runs `terraform init`, `plan` on PR creation when a file in `terraform/github` changes
- runs the above and `apply` when a PR that affects a `terraform/github` file is merged into `main`

Resolves #88.


vevago3535 (approved):
approved

philhorrocks (approved):
approved

PR #343, opened 2021-02-11
Create-core-network-services-deployment

donmasters:
- Created github actions yaml file
- Created script for Terraform Plan and update comments


philhorrocks (dismissed):
approved

scripts/terraform_plan_comment_update.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+MESSAGE=`terraform plan -chdir="$1" -no-color | ./scripts/redact-output.sh | grep '^Plan:'`
```
jakemulley, line 3: Is it worth using the output of `terraform-plan.sh` here, as it already has the correct flags set and redacts the output? It means we'd only need to update `terraform-plan.sh` rather than this file too for the same output?

jakemulley (changes requested):

.github/workflows/core-network-services-deployment.yml
```diff
@@ -0,0 +1,55 @@
...
+  TF_IN_AUTOMATION: true
+
+jobs:
+  core-network-services-deployment-plan:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Terraform plan
```
jakemulley, line 27: You'll need to add a `terraform init` here otherwise it'll fail because of uninstalled providers 👍🏼

jakemulley (changes requested):

.github/workflows/core-network-services-deployment.yml
```diff
@@ -0,0 +1,55 @@
...
+          chmod u+x scripts/terraform_plan_comment_update.sh
+          scripts/terraform_plan_comment_update.sh terraform/environments/core-network-services
+        env:
+          SECRET: ${GITHUB_TOKEN}
+          GITHUB_REPOSITORY: ${GITHUB_REPOSITORY}
+          PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
+
+  core-network-services-deployment-real-apply:
+    runs-on: ubuntu-latest
+    if: github.event.ref == 'refs/heads/main'
```
jakemulley, line 38: You can use `if: ${{ github.event.pull_request.merged == 'true' }}` to check if the PR has been merged into main, rather than checking the `ref`, if easier.

.github/workflows/core-network-services-deployment.yml
```diff
@@ -0,0 +1,55 @@
...
+    needs: [core-network-services-deployment-plan]
+    steps:
```
jakemulley, line 40: GitHub Action jobs spin up new containers so you'll need to redo `terraform init` (and maybe `terraform plan`?) here too.

.github/workflows/core-network-services-deployment.yml
jakemulley, line 27: You'll also want to have a `uses: hashicorp/setup-terraform@v1.3.2` to install Terraform on the container 👍🏼

philhorrocks: ok - will add it to the yaml steps

scripts/terraform_plan_comment_update.sh
philhorrocks: updated to reference scripts/terraform_plan.sh

.github/workflows/core-network-services-deployment.yml
philhorrocks: updated

.github/workflows/core-network-services-deployment.yml
philhorrocks: updated

.github/workflows/core-network-services-deployment.yml
philhorrocks: updated

.github/workflows/core-network-services-deployment.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: hashicorp/setup-terraform@v1.3.2
+      - name: Run terraform plan in terraform/environments
+        run: |
+          scripts/terraform-init.sh terraform/environments/core-network-services
+          PLAN-DETAILS=`scripts/terraform-plan.sh terraform/environments/core-network-services`
+          scripts/terraform_plan_comment_update.sh ${PLAN-DETAILS}
```
jakemulley, line 35: `./scripts/terraform_plan_comment_update.sh` might need to be `./scripts/update-pr-comments.sh` which is in this PR 👍🏼

jakemulley (changes requested):

.github/workflows/core-network-services-deployment.yml
donmasters: fixed

jakemulley (approved):

PR #346, opened 2021-02-11
bug fix core network services automation

donmasters:


philhorrocks (approved):
added a line to sitch the terraform workspace to core-network-services-production

PR #373, opened 2021-02-19
VPC Hub fixes: point NACL rules to the right resource, add missing tags

jakemulley:
This PR:

- Points `core-network-services` to the VPC Hub module, which replaces the `core-vpc` module
- Adds missing tags
- Points `aws_network_acl_rule` to the right NACL, rather than just `public`
- Fixes outputs in VPC Hub to meet the previous implementation
- Fixes an issue with NAT being configured even if `var.gateway` isn't NAT


donmasters (approved):
LGTM

PR #393, opened 2021-02-25
DNS Automation with IAM

donmasters:
Created a Route53 DNS solution that automates Public and Private Zones with IAM controls.

- Automates
  - Creation of Public and Private zones
  - Creates associated IAM roles and policies to give specific accounts the relevant control


philhorrocks (approved):
approved

PR #400, opened 2021-02-25
Breaking change: users now need to provide key, values for environments

jakemulley:
This PR changes current environments to use the new format of the `environments` list as part of the preparation for being able to provide GitHub team names for use with AWS SSO to provide access to outside teams. As this is a breaking change, it prompted the release of [modernisation-platform-terraform-environments@v2.0.0](https://github.com/ministryofjustice/modernisation-platform-terraform-environments/releases/tag/v2.0.0), so updates the module too.

It updates the guidance on creating new environments to reflect the change.


jakemulley:
You can see the result of this PR returns [no changes](https://github.com/ministryofjustice/modernisation-platform/pull/400/checks?check_run_id=1980424330#step:5:61) (which is good) due to the new module.

zuriguardiola (approved):

PR #416, opened 2021-02-26
Remove quotes around template folder

davidkelliott:
We added quotes as we thought withough it would cause issues with
filenames with spaces, but it causes it to be read literally and without
doesn't cause issues with spaces anyway.


philhorrocks (approved):
looks ggod

zuriguardiola (approved):

PR #417, opened 2021-02-26
remodeled cidr-allocation table and tidyup

donmasters:
various changes

- updated titles and some formatting on CIDR table
- changed the CIDR range allocation for Subnet-Sets
- changed the CIDR allocation for Development, Test, Preproduction & Production
- added allocations that will be used for LAB Builds and Scenario testing
- removed all exisiting test data allocation for Shared-VPCs


philhorrocks (approved):
approved

PR #430, opened 2021-03-02
Remove core-vpc-non-live-data

jakemulley:
This PR removes `core-vpc-non-live-data` from the Modernisation Platform as part of the the guide in the #429.

I expect this to return 2 to destroy:

- the AWS secret needs to be replaced, so will be destroyed (to remove the account ID)
- the random string that is set as the email address for the account will also be destroyed as it's a separate Terraform resource

### DIFF
--- a/environments/core-vpc.json
+++ b/environments/core-vpc.json
@@ -3,10 +3,6 @@
     {
       "name": "production",
       "access": []
-    },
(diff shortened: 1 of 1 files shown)


davidkelliott (approved):
👍

PR #453, opened 2021-03-04
feature/modify-json-files-to-create-core-vpc-items

donmasters:
- Remove core-vpc configuration for LAA and HMPPS, as these were used for testing only
- Add configuation for Scenario Lab
- Update json tests for above changes


philhorrocks (approved):
approve

PR #473, opened 2021-03-16
Feature/private acm configuration and member account RAM share config

philhorrocks:
This PR fixes and adds a number of core features:

- Automates ACM Private Root CA and Subordinate certificates. 
- ACM Private CA revocation S3 bucket and list
- ACM PCA subordinate RAM shares for live and non-live CA
- Fixes the core account RAM shares for subnet-sets by association the correct subnets in the resources section
- Fixes the base member account RAM share mapping so that the member accounts map to the correct RAM share hosting the    correct subnet-sets 
- fix the RAM member account tagging for subnets

### DIFF
--- a/terraform/environments/core-network-services/acm-pca.tf
+++ b/terraform/environments/core-network-services/acm-pca.tf
@@ -7,6 +7,7 @@ provider "aws" {
   region = "eu-west-2"
 }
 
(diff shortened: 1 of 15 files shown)


donmasters (approved):
LGTM

PR #477, opened 2021-03-18
added RAM share fix and member VPC templates

philhorrocks:
changes:

- ACM S3 bucket destroy set to false
- RAM share data source call fixed
- templates for RAM setup of member accounts


donmasters (approved):
LGTM

PR #489, opened 2021-03-22
bugfix/core-network-services-routing

### LINKED ISSUE #485: IDENITIFIED WORK - NAT Gateway is not working for Dev/Test/Preproduction
## User Story

Currently NAT gateway is only working for production environments
Investigate and fix issue via automation

## Value

This fix is required to enable for NAT gateway to be used for all environments

## Definition of done

- [x] NAT Gateway works for all environments, this should be tested from resource in the shared VPCs

## Reference

donmasters:
- update core-network-services routing to enable all environments to use
  centralised NAT Gateway
- correct ternary used for Transit Gateway Routing tables, previous
  ternary would only work for production

Changes made are required for Core-network-services routing to function
correctly

Closes #485


philhorrocks (approved):
LGTM

PR #535, opened 2021-03-26
Fix/member account ram share workflow

philhorrocks:
This fixes:
- Member account RAM principal association 
- Implements automated workflow


donmasters (approved):
LGTM

PR #542, opened 2021-04-06
enhancement/update-vpc_nacl-module

donmasters:
json file uses 'name' for subnet set identifier in nacl section

changed identifier to use 'subnet_set' to make it clearer


philhorrocks (approved):
LGTM

PR #551, opened 2021-04-14
Update core-vpc rego with additional validation

davidkelliott:
Including validation for missing fields and wrong types, plus updating the matcher
against the expected values.
Fixing and adding to the unit tests for the tests. Also broke out
expected.rego into a separate file as this will get pretty big as time
goes on. Added a readme explaining how to run the tests.


ezman (approved):
LGTM

PR #553, opened 2021-04-15
Various updates to Bastion Linux module

donmasters:
- various updates to Bastion module
- added a README.md to the module
- updated Sprinkler test app to use updated bastion module
- includes an update to member-vpc module, this changes Security Groups to be singular for an endpoint


davidkelliott (approved):
Looks good 👍

PR #556, opened 2021-04-19
Fix/sso developer group

philhorrocks:
Create a new developer role that maps to a permission set "modernisation-platform-viewer"

The developer role is associated with the correct AWS account via a github_slug reference in the environments/*.json file.

Add a try statement to Terraform to manage a "null" value when the actions section of the environments/*.json file is empty.

### DIFF
--- a/environments/heater.json
+++ b/environments/heater.json
@@ -7,7 +7,7 @@
     {
       "name": "development",
 	  "access": [{
-		"github_slug": "laa-aws-infrastructure",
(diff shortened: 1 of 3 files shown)


davidkelliott (approved):
Looks good, only thing I would say is that we might want to refactor this later if we end up with lots of groups (but hopefully we won't!) 👍

jakemulley (approved):

PR #560, opened 2021-04-20
Add environments repo

davidkelliott:
We want to separate out user environments from core infrastructure.  We
will put user environment code in this repo.


ezman (approved):
LGTM

PR #576, opened 2021-04-22
Remove all member account config and TF files

zuriguardiola:


davidkelliott (dismissed):
Paired, we are confident this will not remove AWS accounts as the scripts to provision accounts do not do anything if they are not present
Edit:
Turns out it will delete everything 😂

davidkelliott (approved):
https://www.youtube.com/watch?v=YVFgI_F4R7Y

PR #577, opened 2021-04-23
Revert back to pre environments experimenting

davidkelliott:
After trying different approaches to splitting the environment we have
decided to keep the main repo as an account vending machine, and we will
have a separate workflow and state files for additional resources that
users want to add in the new environments repo.  This enables us to
truely separate the mod platform controlled and user infrastructure.


philhorrocks (approved):
:+1:

PR #578, opened 2021-04-23
Add key to environment json files to determine whether an account is …

zuriguardiola:
…core or

member.


davidkelliott (approved):
Paired

PR #603, opened 2021-04-28
Rejig concepts and add placeholder topics

davidkelliott:
A bit more tidying to get the documentation in a good place before
starting to fill out the place holder topics.


philhorrocks (approved):
LGTM

PR #605, opened 2021-04-28
feature/update-provision-member-script

donmasters:
- add a condition so the provision-member-directories.sh script also runs when testing on a Mac


davidkelliott (approved):
👍

PR #646, opened 2021-05-11
feature/remove-interpolation-from-single-sign-on

donmasters:


davidkelliott (commented):
Hey, how is this working, if it's bichard we want to use the workspace rather than the workspace with the environment part removed? Thanks

philhorrocks (approved):

PR #717, opened 2021-05-15
Add file check tests for opa

### LINKED ISSUE #713: OPA networking tests don't pick up new/deleted files
## Expected Behavior

If you add or remove an environments-networking json file, the tests should fail unless you update the tests.

## Actual Behavior

Nothing happens

  1. Add a new networking file
  1. Run tests
  1. Watch for lack of failure

davidkelliott:
OPA doesn't check the presence/absence of files because to test the json
you pass in a file, if you don't pass in a file then it can't run a test
against it.  To get round this I've added a bash check, reusing some of
the original bash check jq where possible.

Also added missing performance hub files to enable to tests to be
passing.

Resolves #713


donmasters (approved):
LGTM

PR #730, opened 2021-05-17
Add accessing the AWS console user guidance

davidkelliott:
Adding guidance on how to access the AWS console for users.  Also adding
place holder pages for the other steps users will need to do.


donmasters (approved):
LGTM

PR #737, opened 2021-05-18
Fix bug where options aren't passed in correctly

davidkelliott:
Because the options variable was in quotes it was being passed into the
command as a string


donmasters (approved):
LGTM

PR #742, opened 2021-05-19
Add networking documentation

davidkelliott:
Add networking vpc and subnet documentation.


source/concepts/networking/networking-approach.html.md.erb
```diff
@@ -0,0 +1,93 @@
...
+Protected subnets are created per account and used for [VPC endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpc-endpoints.html). We provide the following VPC endpoints as standard, and can add additional ones as required:
+
+* SSM Messages
+* S3
+* EC2
+* EC2 Messages
+
+##### Transit Gateway
+
+The [transit gateway](https://docs.aws.amazon.com/vpc/latest/tgw/what-is-transit-gateway.html) subnets are created per account to allow access to other accounts and services via the transit gateway.
```
donmasters, line 70: Is it worth mentioning that NACL rules should not be applied to Transit Gateway subnets, as this would break the networking model

source/concepts/networking/networking-approach.html.md.erb
```diff
@@ -0,0 +1,93 @@
...
+
+#### NACLs
+
+Access to the subnets is controlled with [Network ACLs (NACLs)](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html).
+
+By default these NACLs are locked down, allowing only egress and ingress to the public subnets.  We can add additional rules to allow your application to talk to services in different subnet sets as needed.
```
donmasters, line 76: could mention that egress traffic is allowed by a centralised NAT in network services account

donmasters (commented):
This looks good, just left a couple of comments

Also is the Transit Gateway mentioned somewhere else, as it would be worth mentioning that it has two routing tables that split non-prod-data and prod-data

donmasters (approved):
LGTM

PR #754, opened 2021-05-25
Add endpoint for ECR repo

davidkelliott:
ECR-public actions are only supported in the us-east-1 region.


philhorrocks (approved):
Looks good

PR #756, opened 2021-05-26
Change DNS role to be assumed by mp account

### LINKED ISSUE #746: Modify dns role in core-vpc accounts for modernisation platform member ci user to be able to assume
## User Story

The member account CI/CD needs to be able to create/modify/delete dns entries in the core-vpc accounts.  We need to modify the existing dns role so that that the member ci user can assume this role, without giving it permissions to do anything else in the core-vpc accounts.

## Definition of done

- [ ] role created as part of core-vpc workflow
- [ ] readme has been updated
- [ ] tests are green

## Reference

davidkelliott:
We want to be able to assume the dns role using the new member ci user.
This user and role will be used to make member changes to dns rather
than the main mod platform ci user and the admin role.

Resolves #746

### DIFF
--- a/terraform/environments/core-vpc/locals.tf
+++ b/terraform/environments/core-vpc/locals.tf
@@ -1,7 +1,10 @@
+data "aws_caller_identity" "modernisation-platform" {
+  provider = aws.modernisation-platform
+}
+
 locals {
   application_name       = "core-vpc"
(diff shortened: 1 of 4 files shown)


philhorrocks (approved):
LGTM

PR #762, opened 2021-05-27
Update templates and DNS role policy

davidkelliott:
To secure the members terraform, we needed to change the core-vpc role
to use a business unit scoped role rather than an all access role, and
remove the role for shared services. To do this via sed/file
manipulation was becoming messy, so instead we've created a separate
member-providers.tf template file, and depending on whether the
provision member directories script or the provision environment
directories script is running the correct template will be copied. This
also allows us to further modify the repective providers.tf files in
future if we need to and not have to further modify the scripts.

Terraform plan for the member accounts was failing with the error:
`not authorized to perform: route53:GetHostedZone on resource:`
I checked the policy and it is missing this, also it was missing some
other features which would be required to be able to update this zone.
New policy taken from here:

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/access-control-managing-permissions.html#example-permissions-record-owner

### DIFF
--- a/scripts/provision-environment-directories.sh
+++ b/scripts/provision-environment-directories.sh
@@ -84,11 +84,15 @@ copy_templates() {
 
   for file in $templates; do
     filename=$(basename "$file")
-    echo "Copying $file to $1, replacing application_name with $application_name"
-    sed "s/\$application_name/${application_name}/g" "$file" > "$1/$filename"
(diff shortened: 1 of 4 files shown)


zuriguardiola (approved):

philhorrocks (approved):
LGTM

PR #775, opened 2021-06-03
Add automation to generate Github environments

davidkelliott:
Github environments will allow users to control when their deployments
happen.  This takes the github teams defined in the environments json,
and creates an environment per mp environment (aws account). For
preprodcution and production environments, along with reviewer protection,
you can only deploy from main.

Also adds a workflow template that generates a workflow file on account file creation.


ezman (approved):
LGTM

PR #776, opened 2021-06-07
Add github slug for the test environments

davidkelliott:
In order to generate github environments we need to have a team that
will review deployments.  Doing this allows us to use workflow and
github environments for the test envs.


zuriguardiola (approved):
lgtm

PR #780, opened 2021-06-10
Fix terraform scripts always exit with 0

### LINKED ISSUE #559: Github pipeline doesn't fail if adding a user fails
## Expected Behavior

.github/terraform-github.yml to fail if adding a user fails.

## Actual Behavior

When adding a user who is not yet part of the moj organisation.
Workflow completes successfully but apply output shows an error and the new user isn't added to the team:

```
Error: PUT https://api.github.com/organizations/2203574/team/3987072/memberships/christine-elliott: 422 Validation Failed [{Resource:TeamMember Field:user Code:invalid Message:}]

  on modules/team/main.tf line 45, in resource "github_team_membership" "members":
  45: resource "github_team_membership" "members" {
```

  1. Try to add user not part of the moj organisation to the mp team
  1. Merge the pr
  1. Check the apply part of the workflow.

davidkelliott:
Because of piping the terraform command through the redaction script,
the success code will always be 0. Adding pipefail to fix this and also
-e in case there are other uncaught errors in the script.

Closes #559

### DIFF
--- a/scripts/terraform-apply.sh
+++ b/scripts/terraform-apply.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -o pipefail
+set -e
+
 # This script runs terraform apply with input set to false and no color outputs, suitable for running as part of a CI/CD pipeline.
 # You need to pass through a Terraform directory as an argument, e.g.
 # sh terraform-apply.sh terraform/environments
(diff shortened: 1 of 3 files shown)


donmasters (approved):
LGTM

PR #789, opened 2021-06-11
refactor-member-iam-role

### LINKED ISSUE #779: Refactor dns role to be generic
## User Story

The role used to modify DNS entries in the core vpc accounts is currently in the dns module https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/modules/dns-zone/main.tf#L78

There are other things that we need to do such as looking up security groups in the core vpc account, we should move the dns role to the core vpc iam terraform and rename it to something like core-vpc-access so that we can add additional permissions to its policy.  

## Value

This avoids the need to create multiple roles.

## Definition of done

- [ ] readme has been updated
- [ ] tests are green

## Reference

donmasters:
closes #779 

- remove dns member role from DNS module
- create generic role in core-vpc top level
- add additional permissions for data lookups

This change is required to make the delegate-member role more generic
rather than focused purely on DNS

### DIFF
--- a/terraform/environments/core-vpc/vpc.tf
+++ b/terraform/environments/core-vpc/vpc.tf
@@ -67,6 +67,23 @@ locals {
 
   }
 
+  account_numbers  = flatten([
+    for dept, data in local.vpcs[terraform.workspace]: {
+      key = dept
+      account_nos = flatten([
+        for subnet_set in data.cidr.subnet_sets: [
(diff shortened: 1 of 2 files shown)


zuriguardiola (approved):

davidkelliott (approved):
Tested on Sprinkler

PR #790, opened 2021-06-11
Comment out data call providers so they are instead called by local accounts.

zuriguardiola:
This change comments out the shared subnet-set vpc data object and shared subnet-set private (az (a) subnet) data object so that data calls are instead done from the member account.


donmasters (approved):
LGTM

Issue #648, opened 2021-05-11
Speed up account baselines


davidkelliott:
## User Story

As a mp engineer
I expect the new account creation process to be quick
So that I can create a new account quickly for users

## Value

Currently the baseline code runs through every single account on every run, this takes nearly an hour and will only increase as more accounts are added.

## Questions / Assumptions

Run only per the account being created so account creation is fast.
Be able to run ad-hoc to check for drift/correct any changes.

## Definition of done

- [ ] code refactored
- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

philhorrocks:
I have tested the base code in the test repo and this seems to work ok.

branch: fix/speedupbaseline

philhorrocks:
This code needs to be reviewed and then tested by onboarding a new account.

PR #808, opened 2021-06-15
Fix/speedupbaseline

philhorrocks:


davidkelliott (approved):
Ran through this in my head and tested the git diff command, it looks good to me 👍

davidkelliott:
Closes #648

PR #812, opened 2021-06-16
updated baselines for new member CI/CD IAM role

philhorrocks:
Added additional baseline for new member Ci/CD role


philhorrocks:
will fix with module version

PR #815, opened 2021-06-17
change-account-type-for-several-accounts

donmasters:
Change account type for several accounts in environments JSON files

This change is required to enable these accounts to join the correct AWS Organisations OUs


davidkelliott (approved):
Let's remember to merge this is after the other PR as this will trigger the environments workflow and we don't want it to revert.

PR #817, opened 2021-06-17
Fix another syntax error DO NOT MERGE

davidkelliott:
2nd time lucky?!


zuriguardiola (approved):

davidkelliott:
This needs to be fixed in #818 otherwise it will run on the old module and attempt to destroy the new ous.

PR #830, opened 2021-06-18
Fix/cicd member user module

philhorrocks:


davidkelliott (approved):
Looks good to me 👍  we'll have to fix the git diff command before merging this though otherwise I don't think it will run.

PR #831, opened 2021-06-18
scp-configuration-added-to-modernisation-platform-OUs

donmasters:
- three separate SCP policies have been created
- SCP policies have been applied to core, member & member-unrestricted
  OUs under the Modernisation-Platform OU

A framework was required to ensure that accounts can be correctly managed
by relevant SCPs.


davidkelliott (approved):
:shipit:

PR #836, opened 2021-06-18
Give DavidElliott user state access

davidkelliott:
In order to create SCPs in the root account you need a user with more
permissions than the ModernisationPlatformOrganisationManagement which we use
to create OUs and accounts, this change gives users with the approriate
permissions access to the state file to be able to make these changes.


donmasters (approved):
LGTM

PR #855, opened 2021-06-22
Add output for ec2 private IP to bastion module

zuriguardiola:
Bastion instance private IP is used by application instance security group rules so the output was included.


davidkelliott (approved):

philhorrocks (approved):
looks good

PR #857, opened 2021-06-22
Fix/cicd member accounts logic

philhorrocks:


zuriguardiola (approved):

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -10,9 +20,11 @@ module "cross-account-access" {
 
 module "cicd-member-user" {
 
+  count = local.account_data.account-type != "core" || local.account_data.account-type != "member-unrestricted" ? 0 : 1
```
davidkelliott, line 23: Could we flip the logic so that it's `local.account_data.account-type == "member"` ? I'm just thinking it makes it a bit clearer as to which type of accounts we want this to apply to

davidkelliott (changes requested):

terraform/environments/bootstrap/delegate-access/iam.tf
philhorrocks: yeah - no probs, will update

philhorrocks: updated to count = local.account_data.account-type == "member" ? 0 : 1

davidkelliott (approved):
👍

PR #864, opened 2021-06-22
Remove GuardDuty from baselines

### LINKED ISSUE #809: Remove guardduty from secure baselines
## User Story

Guard duty for new and existing accounts can be managed centrally https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_organizations.html

If we remove it from the baselines by removing this line then we can use the organisation set up enrol new accounts.
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/secure-baselines/main.tf#L48

We will need to do a terraform state rm on the resources that have already been created.

## Value

Speed up baselines and control guardduty in one place.

## Questions / Assumptions

The accounts are already part of the guardduty org management so no further action is required to associate them afterwards.

## Definition of done

- [ ] remove guardduty from baselines
- [ ] remove guardduty resources from tf state files (will probably need a script to loop through the 4 regions for each account)
- [ ] readme has been updated
- [ ] tests are green

## Reference

davidkelliott:
We are now enabling GuardDuty from all new accounts by default, so this
is no longer needed in the baselines.

Resolves #809

### DIFF
--- a/terraform/environments/bootstrap/secure-baselines/main.tf
+++ b/terraform/environments/bootstrap/secure-baselines/main.tf
@@ -46,7 +46,7 @@ module "baselines" {
   enabled_ebs_encryption_regions = local.enabled_baseline_regions
 
   # Regions to enable GuardDuty in
-  enabled_guardduty_regions = local.enabled_baseline_regions
+  # Guard duty is now enabled for all new accounts by default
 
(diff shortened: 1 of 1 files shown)


philhorrocks (approved):
apply

PR #873, opened 2021-06-24
Skip analytical platform on baselines

davidkelliott:
This is temporary until we delete default VPCs for all other accounts
AP will fail as they have deleted their default vpc, we want to delete all default vpcs
But are waiting for v4 of the aws provider which will allow us to do this via TF


philhorrocks (approved):
LGTM

PR #884, opened 2021-06-28
Add role to create infrastucture for members

### LINKED ISSUE #810: Create more restrictive role for member accounts
## User Story

The member accounts currently use the ModernisationPlatformAccess role created here - https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/iam.tf to create resources in their environment.  This uses a full admin policy, if we want to restrict the resources that they can create we will need a new role with a stricter policy.  This can then be used to prevent things we don't want members to be able to do such as creating network resources.

## Value

Better security and control over the platform.

## Definition of done

- [ ] new role created
- [ ] policy for role created
- [ ] members provider.tf files and template updated to use the new role
- [ ] readme has been updated
- [ ] tests are green

## Reference

davidkelliott:
Resolves #810, currently the member CI/CD user uses a unrestricted admin
role to create infrastructure.  This is a new role which can be assumed
with restricted permissions, so that we can have confidence that users
don't create resources that we don't want them to to.

### DIFF
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -63,8 +63,8 @@ data "aws_iam_policy_document" "member-access" {
       "kms:*",
       "lambda:*",
       "logs:*",
-      "organizations:Describe*", 
-      "organizations:List*", 
+      "organizations:Describe*",
+      "organizations:List*",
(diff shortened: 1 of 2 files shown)


philhorrocks (approved):
LGTM

PR #885, opened 2021-06-28
Update template and member-ci user to use new member role

davidkelliott:
Now that we have a restricted member role for member infrastructure
changes we need to update the template.

Also changing the role that the member-ci user can assume to this one.


philhorrocks (approved):
LGTM

PR #891, opened 2021-06-29
Allow github slugs in environments to be updated

### LINKED ISSUE #874: Amend github environments if team slug changes
## Expected Behavior

When you update the team github slug in the environments json file, the reviews required for a deployment in github environments is also updated for that environment.

## Actual Behavior

Currently the script checks if the environment is present and only creates it if it is missing. So updates will never happen only creates.

  1. Change the github slug for a team in the environments json file
  1. Let the workflow run
  1. Old team is still in the environment for the reviewer

davidkelliott:
Resolves #874, when the environments json file is changed then the
github environment will be updated to the current json file settings.

### DIFF
--- a/scripts/git-create-environments.sh
+++ b/scripts/git-create-environments.sh
@@ -34,6 +34,16 @@ check_if_environment_exists() {
   echo "Environment $1 exists = ${environment_exists}"
 }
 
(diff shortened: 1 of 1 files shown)


philhorrocks (approved):
LGTM

PR #903, opened 2021-07-02
Enable dynamic tagging

davidkelliott:
Currently tags are all statically defined in the locals file for member
accounts. This change pulls the tags from the environments file, so if
the tags in the environment json file are changed this will be reflected
the next time the infrastructure terraform for that account is run.

I am getting the .json file directly from github over http, rather than
referencing the file in the repository, because the same template is
used in the environments repository, and this will work from both
repositories to pull the file to generate the tags. This keeps it
consistent across the repos.

Also some minor updates to the tags to add things like the source code
repository. This will help going forward to identify if a resource in an
account was generated from the members environments repo or this repo.

Summary of changes:

- Changes to templates to get application environment json file using a http data source and use that for tags
- Changes to provision-member-directories script to modify the above tags to change the source repo to the member environment repo
- Add tags to delegate access bootstrap code
- Add additional tags such as source code where appropriate
- Update tags in member account to match the new template

Todo:
On the member environments repository update existing tags to match the new template.

### DIFF
--- a/scripts/provision-member-directories.sh
+++ b/scripts/provision-member-directories.sh
@@ -104,9 +104,11 @@ copy_templates() {
         if [ `uname` = "Linux" ]
         then
           sed -i "s/environments\//environments\/members\//g" "$1/$filename"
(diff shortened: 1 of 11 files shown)


philhorrocks (approved):
LGTM

PR #906, opened 2021-07-07
Fix manual trigger not working for dev/test

davidkelliott:
The workflow dispatch trigger wasn't working against main branch for dev
and test, sometimes we may want to run main back into an environment.


jackstockley89 (approved):
approved

PR #924, opened 2021-07-15
Fix environments repo workspace creation failing

davidkelliott:
Fixes 920

This is because the script is creating the temporary files for the main repo, running init and creating the workspace, then changing the backend and running init and create again for the environments repo. Because init has been run once before it now thinks there is a change in the backend. We need to clear the temp folders and start from scratch again for the environments repo.

The reason it works on the 2nd time of running the workflow is because it doesn't create the first workspace and then switch.

This fix puts all the file creation code into a new function, and calls
that function before it is needed. If any thing already exists it is
deleted in the function before recreating the files.

### DIFF
--- a/scripts/provision-terraform-workspaces.sh
+++ b/scripts/provision-terraform-workspaces.sh
@@ -50,29 +50,10 @@ APPLICATION=`basename "${JSON_FILE}" .json`
 done
 }
 
-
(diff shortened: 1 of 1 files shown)


jackstockley89 (approved):
Approved

PR #930, opened 2021-07-19
Fix environments repo url not substituted

### LINKED ISSUE #922: Repository url doesn't get replaced in member directory creation
## Expected Behavior

After the template file - locals.tf gets copied for the environments repo the github repository url should be replaced from:
https://github.com/ministryofjustice/modernisation-platform
to
https://github.com/ministryofjustice/modernisation-platform-environments

## Actual Behavior

URL is not being replaced

## Other information

This is because the sed command is being run inside a block that only affects backend.tf file.

davidkelliott:
Closes #922


ezman (approved):
LGTM

PR #932, opened 2021-07-19
update-member-vpc-module-NACL-config

### LINKED ISSUE #917: Harden NAT outbound NACL rules for internet traffic
## User Story

We need to harden to private subnet NACL rules and restrict outbound connections to TCP port 80 and port 443. We also require an inbound rule for TCP traffic on ephemeral ports only.  

Inbound:
140 | 0.0.0.0/0 | TCP | 1024-65535 | ALLOW | Allows inbound return traffic from the NAT device in the public subnet for requests originating in the private subnet.

Outbound: 

100 | 0.0.0.0/0 | TCP | 80 | ALLOW | Allows outbound HTTP traffic from the subnet to the internet.
110 | 0.0.0.0/0 | TCP | 443 | ALLOW | Allows outbound HTTPS traffic from the subnet to the internet.

These rules need to be updated in the the Terraform NACL config.

donmasters:
- Separate Public and Private NACL rule configuration
- Configure Private NACLs to allow ingress high ports only for
  centralised NAT
- Configure Public NACLs to all ingress for 443 only as default

closes #917

### DIFF
--- a/terraform/modules/member-vpc/main.tf
+++ b/terraform/modules/member-vpc/main.tf
@@ -91,11 +91,18 @@ locals {
     if subnet.key != "transit-gateway"
   ])
 
-  # Distinct subnets by key type not including Transit Gateway or data subnets
-  distinct_subnets_by_key_type_excluding_data = distinct([
+# Distinct subnets by key type - Private
+  distinct_subnets_by_key_type_private = distinct([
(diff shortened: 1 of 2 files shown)


davidkelliott (approved):
Looks good

PR #940, opened 2021-07-21
Update missed publisher versions

davidkelliott:
The main publish workflow was updated previously but we missed these
versions in the check-links.yml and make file.


jackstockley89 (approved):
approved

PR #950, opened 2021-08-02
Athena automation

philhorrocks:
Athena Added lambda to setup:
   -  Athena database
   - Athena table
   - Gets list of AWS mod platform accounts to populate Athena schema  

Added Terraform to automate the the build and deployment of the the lambda and setup to associated IAM role for access to Athena and Secrets Manager


terraform/environments/core-logging/lambda/index.py
```diff
@@ -0,0 +1,230 @@
...
+                return secret
+            else:
+                decoded_binary_secret = base64.b64decode(get_secret_value_response['SecretBinary'])
+                #res = json.loads(json.dumps(decoded_binary_secret, default=str))
+                return decoded_binary_secret
+                
+#Create AWS Athena database
+def create_athena_database():
+    
+    print("[+] Creating Athena database" + athena_db)
```
ezman, line 79: Might need an extra space.

ezman (approved):
LGTM

davidkelliott (changes requested):
Hi Phil could you have a look at some of the issues the linter has picked up and see if any of them are valid?  Some of them around encryption etc look relevant (some will be for the existing logging infrastructure as the scan runs on the whole core-logging folder). If the checks are completely irrelevant you can add an ignore comment in for tfsec,checkov or tflint as appropriate.  Don't worry about fixing everything now we need to take a bit of time to go through all our existing code at some point but it would be good to keep anything new we create in check. Thanks
Thanks.

philhorrocks:
needs to rebase

PR #989, opened 2021-08-11
Athena automation

philhorrocks:
- rebased the code with main
- added Terraform code to deploy Athena lambda on a cloudwatch schedule
- added Python lambda to configure Athena DB and tables


philhorrocks:
need to fix conficts

Issue #947, opened 2021-07-29
Review and rectify / ignore static code analysis warnings


davidkelliott:
## User Story

Now that we have static code analysis running we should strive to address any issues it raises as we go.  Either by fixing the issue if valid, or by ignoring it using the build in ignore comments of tfsec, checkov and tflint.

If you run a full scan now there are still quite a few issues that should be resolved, this ticket is to go through them and either fix/ignore them.

## Value

We make sure our code is following best practices, or where we have a valid reason to ignore a practice we ignore it in the code.  If a test starts of red then there is little incentive for people to make it green.  If it starts off green there is a strong incentive to keep it green.

## Definition of done

- [ ] go through static analysis scan and review raised issues
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

pete-j-g:
No action needed to correct CKV_AWS_144 from the perspective of the modernisation-platform repository, as it's currently affecting the bastion_linux module which is being moved into it's own repository.

Buckets within the Core infrastructure should be replicated across regions, as it is a requirement for PCI DSS compliance.
Member account buckets requirements will need to be assessed on case by case basis

adeweetman-al:
Re-did full scan with now reduced number of SCA warnings, updated summary below.

**IAM**
Several warnings about weak policies (e.g. without constraints / permissions management / resource exposure) Example case: CKV_AWS_109: "Ensure IAM policies does not allow permissions management / resource exposure without constraints"
FAILED for resource: aws_iam_policy_document.kms-acm

**Network**
Security Groups
Quite a few CKV2_AWS_12: "Ensure the default security group of every VPC restricts all traffic"FAILED for resource: aws_vpc.vpc

**Logging**
Quite a few [aws-cloudwatch-log-group-customer-key][LOW] Resource 'module.vpc["live_data"]:aws_cloudwatch_log_group.default' is only using default encryption
/github/workspace/terraform/modules/vpc-hub/main.tf:105-109

[aws-s3-enable-bucket-logging][MEDIUM] Resource 'aws_s3_bucket.acm-pca' does not have logging enabled.
/github/workspace/terraform/environments/core-network-services/acm-pca.tf:11-56

**S3**
Cross-regional replication should be enabled, now covered under #1044 

**Secrets**
Ensuring secrets manager secrets are encrypted using KMS

[...]

PR #1032, opened 2021-08-27
Exclude tfsec warning about public github repos

adeweetman-al:
Address part of #947 

Simple addition of exclusion already added for checkov to reduce noise from tfsec.


pete-j-g (approved):
lgtm

PR #1035, opened 2021-08-31
Exclude static code analysis open nacl ports warnings

adeweetman-al:
Addresses part of #947 - See Network category of warnings

After a team chat last week, we discussed the value of this rule [https://tfsec.dev/docs/aws/vpc/no-excessive-port-access/](https://tfsec.dev/docs/aws/vpc/no-excessive-port-access/) in the context of mod platform and the costs/benefits of it.
It was deemed to be the case that the cost of this (i.e. the noise it generates within the context of the overall design and approach of the mod platform) outweighs any benefits. We agreed that tfsec shouldn't be considered the optimal way to check for quality in the context of mod platform network security. 
This exclusion would eliminate 36 warnings based on decisions made in terraform/modules/vpc-hub/main.tf

### DIFF
--- a/.github/workflows/terraform-static-analysis.yml
+++ b/.github/workflows/terraform-static-analysis.yml
@@ -25,7 +25,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         scan_type: changed
-        tfsec_exclude: AWS095,GIT001
+        tfsec_exclude: AWS095,GIT001,AWS050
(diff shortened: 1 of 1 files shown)


davidkelliott (approved):
Happy to ignore this one, I will create a ticket though as to how we verify what we want is open is open. I'm conscious that our networking is complex, and we need to have confidence in the changes we make. https://github.com/ministryofjustice/modernisation-platform/issues/1036

pete-j-g (approved):

PR #1038, opened 2021-08-31
Suppress Checkov IAM failure messages

pete-j-g:
Addresses part of #947
Suppress warnings for policies intended for admin users


davidkelliott (approved):
👍  we may lock things down further at a later date but at the moment these should be open.

PR #1040, opened 2021-08-31
Change create workspace logic to not core

### LINKED ISSUE #1012: Unrestricted member accounts workspaces aren't created for member repo files
## Expected Behavior

Terraform workspace is created

## Actual Behavior

Workspace is not created

  1. Create unrestricted member account

## Other information

This line means it only gets created for member accounts - need to change to !=core or member and non member
https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-terraform-workspaces.sh#L118

davidkelliott:
Both member and member unrestricted account types need the workspaces to
be created.  This changes the logic so an non core account type get the
workspace created.  This is in keeping with the other logic in the
script.

Fixes #1012

### DIFF
--- a/scripts/provision-terraform-workspaces.sh
+++ b/scripts/provision-terraform-workspaces.sh
@@ -115,7 +115,7 @@ do
     fi
 
     ACCOUNT_TYPE=$(jq -r '."account-type"' ${JSON_FILE})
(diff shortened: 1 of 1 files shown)


ezman (approved):
LGTM

pete-j-g (approved):
lgtm

PR #1042, opened 2021-09-01
Add tfsec exclusion for blocking public access to CRL ACM-PCA s3 bucket

adeweetman-al:
This addresses part of #947 

Propose we add an exclusion for this since we're creating a CRL associated to the private CA, with a specific, restricted access permitted from acm-pca.amazonaws.com. This is based on https://docs.aws.amazon.com/acm-pca/latest/userguide/PcaCreateCa.html#s3-policies and https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-error-crl-acm-ca/.


davidkelliott (approved):
LGTM

PR #1043, opened 2021-09-01
Secrets Manager secret encrypted using KMS

sobostion:


adeweetman-al:
Looks good.
As discussed, there's another SCA warning about the same rule:
Check: CKV_AWS_149: "Ensure that Secrets Manager secret is encrypted using KMS"
	FAILED for resource: aws_secretsmanager_secret.member_ci_iam_user_keys
Relating to terraform/modernisation_platform_account/secrets.tf for the ci_iam user keys.
Probably makes sense to fix this as well under the same PR?

adeweetman-al:
@sobostion - I've pushed changes relating to ci secrets, equivalent to your work on the environment management secrets and made sure we're getting a green TF plan.
My feeling is that we should get one of the existing engineers to review this as well as there's a few assumptions in here (certainly from my PoV).
e.g.
- permissioning on the keys is valid to enable the right principals to use the keys to decrypt the secrets
- valid decision to have a kms key per ci user

davidkelliott (commented):
Before this gets merged please can I have a look at this next week as this is quite a key area and the secret is called from several different places, I just want to double check this and won't have time this week. 🙏  thanks

davidkelliott:
After the discussion last week we decided not to do this, the gain is too little for the risk, we will add an exclusion for SCA.

PR #1046, opened 2021-09-01
Key policy requires asterisk resource

sobostion:


pete-j-g (approved):

terraform/modernisation-platform-account/dynamodb.tf
```diff
@@ -11,13 +11,15 @@ resource "aws_kms_key" "dynamo_encryption" {
...
+
+  # checkov:skip=CKV_AWS_109: "Key policy requires asterisk resource"
+  # checkov:skip=CKV_AWS_111: "Key policy requires asterisk resource"
+
   statement {
     effect  = "Allow"
     actions = ["kms:*"]
 
-    resources = [
-      "arn:aws:dynamodb:*"
```
davidkelliott, line 19: Just curious, did this not work?

sobostion: That change didn't throw an error on terraform plan, but when it came to applying it, it gave the error "The new key policy will not allow you to update the key policy in the future".

Based on this page [here](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html) it looks like only the asterisk is allowed in the resource section of a key policy.

`Resource – (Required) In a key policy, the value of the Resource element is "*", which means "this KMS key." The asterisk ("*") identifies the KMS key to which the key policy is attached.`

PR #1064, opened 2021-09-07
Exclude checkov warning regarding default SGs and full traffic restriction

adeweetman-al:
Addresses part of #947 
Follows a chat with @davidkelliott regarding the action to take regarding these warnings.
e.g. _Check: CKV2_AWS_12: "Ensure the default security group of every VPC restricts all traffic" FAILED for resource: aws_vpc.vpc_
https://docs.bridgecrew.io/docs/networking_4

Outcome of discussion was to exclude these warnings. Default security groups are unused. Better longer-term solution would be to bring these under better terraform management / be able to remove them.

### DIFF
--- a/terraform/modules/member-vpc/main.tf
+++ b/terraform/modules/member-vpc/main.tf
@@ -176,6 +176,7 @@ locals {
 
 # VPC
 resource "aws_vpc" "vpc" {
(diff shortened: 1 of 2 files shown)


davidkelliott (approved):
👍  the ability to delete default VPCs via terraform should be coming in v4 of the provider - https://github.com/hashicorp/terraform-provider-aws/milestone/119

PR #1067, opened 2021-09-08
Ignoring lack of cross-regional replication for acm-cma bucket

adeweetman-al:
Addresses part of #947 

Relates to static code analysis checkov warning:
```
Check: CKV_AWS_144: "Ensure that S3 bucket has cross-region replication enabled"
	FAILED for resource: aws_s3_bucket.acm-pca
	File: /acm-pca.tf:13-59
	Guide: https://docs.bridgecrew.io/docs/ensure-that-s3-bucket-has-cross-region-replication-enabled
```
Consistent with decisions around cross-regional replication for S3 buckets elsewhere, this PR adds an exclusion for a checkov rule suggesting that replication should be set up. Exclusion added for this particular use case to at least encourage some consideration in future cases should they be different or if design principles change.


sobostion (approved):
lgtm

davidkelliott (approved):

PR #1076, opened 2021-09-09
Fix bug where the app json file is needed

davidkelliott:
The application json file is part of the templates and needed for the
environments repository, but it isn't needed here.  Commenting rather
than removing as it's less code in the provisioning script to comment
rather than remove. Also leaving as a comment also serves as a reminder
as to how we should handle environment specific variables if it is ever
needed in this repository.

### DIFF
--- a/terraform/environments/tariff/locals.tf
+++ b/terraform/environments/tariff/locals.tf
@@ -38,5 +38,5 @@ locals {
   # environment specfic variables
   # example usage:  
   # example_data = local.application_data.accounts[local.environment].example_var
-  application_data = jsondecode(file("./application_variables.json"))
+  # application_data = jsondecode(file("./application_variables.json"))
 }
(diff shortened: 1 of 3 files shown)


terraform/environments/tariff/locals.tf
```diff
@@ -38,5 +38,5 @@ locals {
   # environment specfic variables
   # example usage:  
   # example_data = local.application_data.accounts[local.environment].example_var
-  application_data = jsondecode(file("./application_variables.json"))
+  # application_data = jsondecode(file("./application_variables.json"))
```
gfou-al, line 41: I don't mind the existing approach but I would prefer to use the terraform fileexists function to determine whether the application_variables.json exists before reading it: https://www.terraform.io/docs/language/functions/fileexists.html This way we would not need to programmatically comment it out in the provision-environment-directories.sh script. Other that that, the code changes seem fine.

gfou-al (dismissed):
I don't mind the existing approach but I would prefer to use the terraform fileexists function to determine whether the application_variables.json exists before reading it: https://www.terraform.io/docs/language/functions/fileexists.html This way we would not need to programmatically comment it out in the provision-environment-directories.sh script. Other that that, the code changes seem fine.

gfou-al (approved):
looks good

PR #1077, opened 2021-09-09
Revert "Trigger a redeploy by changing bastion_linux to false which does not …"

davidkelliott:
Reverts ministryofjustice/modernisation-platform#1075


gfou-al (approved):
lgtm

PR #1078, opened 2021-09-09
Set -e to ensure script doesn't fail silently

### LINKED ISSUE #1074: core-vpc-development-deployment ram share workflow fails silently
## Expected Behavior

Actions job fails when the script fails

## Actual Behavior

The ram share section failed silently and gave a green tick for the action.

  1.
  1.
  1.

## Other information

This fix will probably need doing for the other vpc deployment workflows

davidkelliott:
Fixes #1074

### DIFF
--- a/scripts/member-account-ram-association.sh
+++ b/scripts/member-account-ram-association.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 basedir=terraform/environments
 environment=${2}
 account=`echo ${1} | sed -e "s/-${environment}//g"`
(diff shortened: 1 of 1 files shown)


scripts/member-account-ram-association.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
gfou-al, line 3: Looks good. Just to note, this article https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/ recommends the use of `set -Eeuo pipefail`. As stated in the above: `set -e` by itself is far from enough. We can further improve upon the behavior created by set -e by combining it with set -o pipefail.

gfou-al (approved):
I approve with a comment. Refer to the code comment for details.

gfou-al (approved):
looks good

PR #1081, opened 2021-09-10
Add regional replication to S3 state bucket

davidkelliott:
Adding regional replication and switching to customer managed KMS keys for the
terraform state bucket. This follows the AWS best practice guidelines,
and if there is ever an issue where we loose access to the London region
we will still be able to see the last known state of the platform.


pete-j-g (approved):

ezman (approved):
LGTM

PR #1084, opened 2021-09-13
Fix and simplify ACM-PCA kms key policy

adeweetman-al:
Makes a fix to #1037 

Given requirement for resource element in kms key policy (see https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html), resource element has been reverted to "*" and checkov exclusions added.


davidkelliott (approved):
👍

PR #1113, opened 2021-09-17
Switch from hard coded environments reference

davidkelliott:


gfou-al (approved):
I am new to terraform, but the change looks good to me. My understanding is that this will query the resource defined in secrets.tf

```
data "aws_secretsmanager_secret" "environment_management" {
  provider = aws.modernisation-platform
  name     = "environment_management"
}
```

and will retrieve the arn value dynamically

davidkelliott:
@gfou-al that's correct, the data resource - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/secretsmanager_secret gets the secret with the name "environment_management" in the modernisation platform account (this is determined by the provider) and then we reference it and get its arn attribute.

PR #1132, opened 2021-09-28
bug core-accounts-missing-default-routes

donmasters:
- update vpc-hub module to enable the default routes to point to the
  attached Transit Gateway
- update code to supply Transit Gateway ID in
  - core-security
  - core-logging
  - core-shared-services


davidkelliott (approved):
LGTM

Issue #1061, opened 2021-09-07
Build AMI networking infrastructure in shared services


davidkelliott:
## User Story

We need a secure area in the shared services account to build ami images.

This should be something like an isolated VPC, this might need to have a public subnet if using github actions or it might be private if using AWS image builder, we should decide the tools we are using before tacking this one.

This ticket is for building the infrastructure, this should be done after the design has been done #1060 

## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

PR #1133, opened 2021-09-29
Networking for AMI building in shared services

adeweetman-al:
Adding VPC endpoints (and dependent resources) to private subnet in non_live_data vpc in shared services account as part of the capability to securely build AMIs using Image Builder.

Related to #1061 .

### DIFF
--- a/terraform/environments/core-shared-services/vpc.tf
+++ b/terraform/environments/core-shared-services/vpc.tf
@@ -41,30 +41,30 @@ module "vpc" {
 
 locals {
   # Derive non_live_data vpc and private subnets for configuration setting up the vpc endpoints
-  non_live_data_vpc_id    = one([for k in module.vpc : k.vpc_id if k.vpc_cidr_block == local.networking.non_live_data])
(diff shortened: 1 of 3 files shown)


terraform/environments/core-shared-services/vpc.tf
```diff
@@ -26,4 +37,87 @@ module "vpc" {
...
-}
\ No newline at end of file
+}
+
+locals {
+  # Derive non_live_data vpc and private subnets for configuration setting up the vpc endpoints
+  non_live_data_vpc_id = one([for k in module.vpc : k.vpc_id if k.vpc_cidr_block == local.networking.non_live_data])
+  private_subnet_ids   = flatten([for k in module.vpc : k.non_tgw_subnet_ids_map.private if k.vpc_cidr_block == local.networking.non_live_data])
+
+  # Get the private route table keys and values, then transform the values (the actual route table ids) into a list
```
davidkelliott, line 47: 👏  thanks for the comment on the dataset transformation, these are always a nightmare to look back at 😄

davidkelliott (approved):
👍

PR #1137, opened 2021-09-30
change-cidr-range-for-protected-subnets

donmasters:
- update protected cidr ranges to use addresses from 10.238.0.0/16
range rather than 10.232.0.0/16 range.

This change is due to a CIDR overlap identified as part of the work to
join Modernisation Platform TGW with PTTP production TGW. The 10.232
range is being used for Magistrates Courts


gfou-al (approved):
looks good

PR #1143, opened 2021-10-01
remove-circular-reference-from-terraform-module

donmasters:
- there is a fix in place where using a separate terraform
  module to create the tgw routes is needed


gfou-al (approved):
Looks good

PR #1160, opened 2021-10-06
Fix Error: Inconsistent conditional result types

gfou-al:
on locals.tf line 41, in locals:
  41:   application_data = fileexists("./application_variables.json") ? jsondecode(file("./application_variables.json")) : []

The true and false result expressions must have consistent types. The given expressions are object and tuple, respectively.
Error: Process completed with exit code 1.

### DIFF
--- a/terraform/environments/xhibit-portal/locals.tf
+++ b/terraform/environments/xhibit-portal/locals.tf
@@ -38,5 +38,5 @@ locals {
   # environment specfic variables
   # example usage:  
   # example_data = local.application_data.accounts[local.environment].example_var
-  application_data = fileexists("./application_variables.json") ? jsondecode(file("./application_variables.json")) : []
+  application_data = fileexists("./application_variables.json") ? jsondecode(file("./application_variables.json")) : {}
 }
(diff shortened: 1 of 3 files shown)


terraform/environments/xhibit-portal/locals.tf
```diff
@@ -38,5 +38,5 @@ locals {
   # environment specfic variables
   # example usage:  
   # example_data = local.application_data.accounts[local.environment].example_var
-  application_data = fileexists("./application_variables.json") ? jsondecode(file("./application_variables.json")) : []
+  application_data = fileexists("./application_variables.json") ? jsondecode(file("./application_variables.json")) : {}
```
gfou-al, line 41: I hope this will not re-start xhibit onboarding. If yes, then I guess reverting would not make any difference.

davidkelliott: No this shouldn't have any impact

davidkelliott (approved):
👍  this should be fixed in the environments repo too wherever it is used in there

Issue #1147, opened 2021-10-01
Spike - can developers run a Terraform plan locally


davidkelliott:
## User Story

As a developer or ops engineer
I want to run a terraform plan locally against my account
So that I can quickly iterate and check my code whilst building infrastructure

## Value

Easier workflow for developers or ops engineers using the platform who already understand and are used to working with Terraform
## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
Update blockers which prevent this:

|---|---|
|Blocker|Possible solution? | Tested? |
|Member code assumes an admin role to plan and build on the pipeline | Comment out this assume when running locally | Yes - working |
|Need to run with AWS credentials | Can get developer credentials from SSO | Yes - working|
| Need to access terraform state bucket in mp account| SSO users call the API from the member account, so giving all accounts in the modernisation platform OU read only access to the bucket will resolve this | Yes - working |
| Need to access terraform state bucket encryption key in mp account| SSO users call the API from the member account, so giving all accounts in the modernisation platform OU access to use the key for decryption will resolve this | Yes - working |
| Developer role needs permissions to decrypt using KMS keys | Give relevant permissions in SSO permission sets in root account | Yes working
| Developer role needs access to the environments secret for terraform plan in the mp account | Add policy to environments secret to allow across account access | In progress - the plan wants to destroy OUs, need to work out why and if it would have an impact

PR #1176, opened 2021-10-14
Enable SSO developer role to run terraform plan

davidkelliott:
This commit does the following things to allow modernisation platform developer SSO users to run a terraform plan locally:

- Parameter in each member account containing the environment management secret ARN so it can be referenced locally
- Create a readonly roles for the core-network-services and core-vpc accounts to enable reading DNS information (these can only be assumed by accounts in the modernisation platform OU)
- Create KMS key to enable sharing of the environment management secret to modernisation platform member accounts
- Policy alterations to share the state bucket KMS key and the bucket to mod platform member accounts.
- Changes to templates to switch to getting the environment management
secret by ARN rather than name

See this PR for changes to the modernisation-platform-developer SSO role - https://github.com/ministryofjustice/aws-root-account/pull/275


davidkelliott:
Closes #1147

Issue #1191, opened 2021-10-19
Bring default security groups under TF and remove ingress/egress rules


davidkelliott:
## User Story

Following on from https://github.com/ministryofjustice/modernisation-platform/issues/1090
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_security_group

## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

adeweetman-al:
Workflows run to deploy changes into shared services, network services, security and logging core accounts.

PR #1220, opened 2021-10-28
Secure default security groups in core, non-default VPCs

adeweetman-al:
Relates to https://github.com/ministryofjustice/modernisation-platform/issues/1191

Addition of the aws_default_security_group to the vpc-hub module to bring management of the default security group for the non-default VPC under Terraform. Removing the default inbound and outbound rules as per recommended best practice.


adeweetman-al:
An update regarding the state of this PR.
Code changes approved. Merging (and TF applies) dependent on confirmation of replacement actions that Terraform will do on TGW route table routes and subnet route table routes across logging, security and shared-services accounts.

PR #1231, opened 2021-11-01
Added scheduled runs of SCA workflow

pete-j-g:


.github/workflows/terraform-static-analysis.yml
```diff
@@ -1,6 +1,9 @@
 name: Terraform Static Code Analysis
 
 on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '* 7 * * 1-5'
```
adeweetman-al, line 6: I read this as every minute past 7am, Mon-Fri. Should this not be: 0 7 * * 1-5

pete-j-g: 🤦 you are 100% correct, I'll get it updated in a few

pete-j-g: cron expression updated

adeweetman-al (approved):
Looks good.

PR #1241, opened 2021-11-03
Delete Ops Engineering env

pete-j-g:


environments/ops-engineering.json
```diff
@@ -1,14 +1,6 @@
 {
```
davidkelliott, line 1: This whole file can be deleted as the development environment is the last one

davidkelliott (changes requested):

environments/ops-engineering.json
pete-j-g: File deleted

davidkelliott (dismissed):
👍  thanks!

davidkelliott (approved):

PR #1244, opened 2021-11-05
Fix removal of last account from a business unit

davidkelliott:
When trying to remove operations engineering account (the last
platforms-development account) there is an error on the delegate access
and single sign on plans as it is pulling in an "account name" of
platforms-development.

```
Run accounts=(platforms-development)
[+] Running single sign on baseline for account platforms-development

Initializing the backend...

Successfully configured the backend "s3"! Terraform will automatically
use this backend unless the backend configuration changes.

Initializing provider plugins...
- Finding hashicorp/aws versions matching ">= 3.47.0"...
- Installing hashicorp/aws v3.63.0...
- Installed hashicorp/aws v3.63.0 (signed by HashiCorp)

Terraform has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that Terraform can guarantee to make the same selections by default when
you run "terraform init" in the future.

Terraform has been successfully initialized!
Running terraform for workspace in terraform/environments/bootstrap/single-sign-on for account: platforms-development
[+] Terraform plan =====>
Error: Process completed with exit code 1.
```

There are a couple of issues which add to this around the existing diff
command:

```
git diff --no-commit-id --name-only --diff-filter=AM -r @^ | awk
'{print $1}' | grep ".json" | grep -a "environments//*"  | uniq | cut
-f2-4 -d"/" | sed 's/.\{5\}$//')
```

Firstly this looks for changes including deletes, we don't need to worry
about deletes for single sign on, secure baselines and delegate access,
as this infrastructure is removed during account removal, and it is
manually deleted from Terraform state as part of the delete process.

So adding `--diff-filter=AM` filters for only Additions and
Modifications.

Secondly, the `grep -a "environments/*"` part of the command, is
running as "environments*" so it is picking up the changes to the
`environment-networks/platforms-development.json` file, which shouldn't
be picked up. Changing this part to `grep -a "environments//*"` resolves
this.

Finally making similar changes with the //* to other scripts which use
it to ensure similar issues don't occur.

### DIFF
--- a/.github/workflows/core-vpc-development-deployment.yml
+++ b/.github/workflows/core-vpc-development-deployment.yml
@@ -105,7 +105,7 @@ jobs:
       - name: get changed files
         id: new_account
         run: |
-          echo "::set-output name=files::$(git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks/*" | grep "${TF_ENV}" | uniq | xargs jq  ".cidr.subnet_sets[].accounts[]" | xargs)"
+          echo "::set-output name=files::$(git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks//*" | grep "${TF_ENV}" | uniq | xargs jq  ".cidr.subnet_sets[].accounts[]" | xargs)"
       - name: Set RAM assocation for member account
         run: |
           accounts=(${{ steps.new_account.outputs.files }})
--- a/.github/workflows/core-vpc-preproduction-deployment.yml
+++ b/.github/workflows/core-vpc-preproduction-deployment.yml
@@ -106,7 +106,7 @@ jobs:
       - name: get changed files
         id: new_account
         run: |
-          echo "::set-output name=files::$(git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks/*" | grep "${TF_ENV}" | uniq | xargs jq  ".cidr.subnet_sets[].accounts[]" | xargs)"
+          echo "::set-output name=files::$(git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks//*" | grep "${TF_ENV}" | uniq | xargs jq  ".cidr.subnet_sets[].accounts[]" | xargs)"
       - name: Set RAM assocation for member account
         run: |
           accounts=(${{ steps.new_account.outputs.files }})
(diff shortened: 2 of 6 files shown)


gfou-al (changes requested):
After deleting environments-networks/house-production.json I run the old command without passing `--diff-filter=AM`
```
git diff --no-commit-id --name-only -r @^
environments-networks/house-production.json
environments-networks/platforms-development.json
terraform/environments/core-logging/athena.tf
terraform/modules/app-ecr-repo/main.tf
```
On the other hand, when passing `--diff-filter=AM`
```
git diff --no-commit-id --name-only --diff-filter=AM -r @^
environments-networks/platforms-development.json
terraform/environments/core-logging/athena.tf
terraform/modules/app-ecr-repo/main.tf
```
which means the code change works as expected.

After modifying environments-networks/platforms-development.json I run the old command:

```
git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks/*"
environments-networks/platforms-development.json
```
and then changed the last part to `grep "environments-networks//*"`
```
git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments-networks//*"
environments-networks/platforms-development.json
```
Unless I am missing something, the above indicates the code change is not working as expected. According to the PR description, I would expect the grep not to match `environments-networks/platforms-development.json`.

davidkelliott:
@gfou-al 
>Unless I am missing something, the above indicates the code change is not working as expected. According to the PR description, I would expect the grep not to match environments-networks/platforms-development.json.

So in your example you are grepping for `grep "environments-networks/*"` and the string contains `environments-networks/platforms-development.json` which is a match.

The current issue is that when you grep for `grep "environments/*"` , so just the files in the environments/ folder, that also matches with the string `environments-networks/platforms-development.json`.

I think this is because the / is read as an escape character, but a double  // resolves this.  So if you try the following lines of your example you should see a difference 🤞 

```
git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments/*"
environments-networks/platforms-development.json
```

and then change the last part to grep `"environments//*"`

```
git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments//*"
Hopefully nothing here
```

gfou-al (approved):
Looks good

```
git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments/*"
environments-networks/platforms-development.json
terraform/environments/core-logging/athena.tf


git diff --no-commit-id --name-only -r @^ | awk '{print $1}' | grep "environments//*"
terraform/environments/core-logging/athena.tf
```

PR #1246, opened 2021-11-05
Ignore new SCA warnings

davidkelliott:
Some new SCA warnings have appeared, looking at them:

```
[aws-lambda-enable-tracing][LOW] Resource 'aws_lambda_function.athena_table_update' uses default value for tracing_config.mode
```
We have already decided that X-Ray tracing is not required and excluded
it from other checks.

```
[aws-ecr-repository-customer-key][LOW] Resource 'module.performance_hub_ecr_repo:aws_ecr_repository.ecr_repo' configures encryption without using CMK
```
We are using the Amazon managed keys, we do not need the flexibility of
customer managed keys for this repo.


gfou-al (approved):
looks good

PR #1249, opened 2021-11-09
Change NOMIS test account to developer role

davidkelliott:
The Nomis team no longer need administrative privillages so returning
them to the developer role.


gfou-al (approved):
Looks good

PR #1256, opened 2021-11-12
Update IAM user for memeber accounts

jackstockley89:
Updating the Policy for the memeber accounts to stop them from being
able to interact with the core vpc setting and creating there own
networking, as they should using what is supplied on account creation


davidkelliott (approved):
👍

PR #1269, opened 2021-11-16
Add security group permissions to member delegation role

zuriguardiola:
This allows creation and deletion of security groups


davidkelliott (approved):
👍

PR #1274, opened 2021-11-16
Add security group permissions

zuriguardiola:
Used this as an example for allowing deletion and creation of security groups

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-policies-ec2-console.html#ex-security-groups


ezman (approved):
LGTM

PR #1280, opened 2021-11-17
core-vpc-sandbox-cidr-allocation-update

donmasters:
- update cidr-allocation.md with new CIDR range for core-vpc-sandbox
- move sprinkler and cooker applications to use new CIDR space above


gfou-al (approved):
Looks good

PR #1292, opened 2021-11-19
Restrict ec2 permissions

davidkelliott:
EC2 permissions cover a large range of networking options as well, see
here - https://docs.aws.amazon.com/AWSEC2/latest/APIReference/OperationList-query.html

We tried denying the VPC actions but the policy was too big, so trying
the other way round here - not allowing all EC2 actions only the ones
needed for building and maintaining EC2s.


gfou-al (approved):
looks good

PR #1295, opened 2021-11-22
Add github actions job for core vpc sandbox deployment

gfou-al:


gfou-al:
won't merge to main branch yet because it relies on changes inside the network-cidr-changes-work

PR #1304, opened 2021-11-24
Moved core-security services to the new IP ranges

gfou-al:


SteveLinden (approved):
New IPs are as supplied by Vodafone

SteveLinden (approved):
OK

PR #1308, opened 2021-11-26
Add new module repository for member VPC accounts

gfou-al:


SteveLinden (approved):
As discussed in the call

Issue #1316, opened 2021-12-02
New AWS account for MI Platform development


rossjones:
### Application Name

mi-platform

### GitHub team slug

analytical-platform

### Environments

- [x] Development
- [ ] Test
- [ ] Preproduction
- [ ] Production

### Tags

tag | value
--- | ---
application |  mi-platform
business-unit |  Platforms
owner |  "Analytical Platform: analytical_platform@digital.justice.gov.uk"

### Connectivity

Just Internet connectivity required

## Additional features

Account-type: member-unrestricted

## Other information

This account will be used to for:

* development of the MI platform (Redshift+Quicksight+some apps)
* cross-account IAM with analytical-platform 
* cross-account S3 bucket access with data-engineering

PR #1319, opened 2021-12-03
Set up new mi-platform unrestricted account

davidkelliott:
New unrestricted AWS development account for the mi-platform, as
detailed in #1316.


gfou-al (approved):
looks good

PR #1322, opened 2021-12-06
Add PagerDuty to the platform

davidkelliott:
This PR adds:

- set up new pagerduty users and add existing ones
- create a modernisation platform team
- pagerduty core alerts service
- a module for integrating required SNS topics with a pagerduty service
- adds the module to the modernisation platform core account
- sets up pager duty policies and schedules (based on the Cloud Platform setup)


gfou-al (approved):
looks good

PR #1325, opened 2021-12-07
Add missing sandbox role to member ci

davidkelliott:
The sprinker workflow was failing with:

```
Error: error configuring Terraform AWS Provider: IAM Role (arn:aws:iam::xxxxxxx:role/member-delegation-garden-sandbox) cannot be assumed.
```

This adds the missing resource to the allow statement


gfou-al (approved):
looks good

PR #1326, opened 2021-12-07
change-routing-configuration-for-tgw-peering

donmasters:
- remove VPC and subnets used specifically for routing workaround
- remove routing tables and entries for routing subnets
- remove FlowLogs for routing VPC
- accept TGW peering request from PTTP Transit Gateway
- update TGW routing tables to use Peering attachment


davidkelliott (approved):
👍  looks good to me

PR #1330, opened 2021-12-08
Add pagerduty workflow

davidkelliott:
This adds a workflow to automate the deployment of pagerduty terraform.

 - workflow
 - secret for the token
 - update main github repo to add the token as a secret


davidkelliott:
Note - the pagerduty plan will fail currently as it needs the token which won't be available in the repo until the github changes in this PR have been approved.

zuriguardiola (approved):

PR #1340, opened 2021-12-09
alter-HMPPS-ingress-nacls-for-test-env

donmasters:
- update HMPPS NACLs in test environment to allow traffic from
Global Protect VPN through PTTP Transit Gateway


davidkelliott (approved):
👍

PR #1341, opened 2021-12-09
Remove unneeded variable

davidkelliott:
The attachment has now been moved outside of the module, this removes
an unneeded variable and we tidied up the terraform state.


donmasters (approved):
LGTM

PR #1344, opened 2021-12-10
Integrate existing core alarms with pagerduty

### LINKED ISSUE #1334: Integrate existing core platform baseline alarms with PagerDuty and Slack
## User Story

We want the GuardDuty, SecurityHub and Config Cloudwatch alarms that are already configured to go to a service on pagerduty and then onto our #modernisation-platform-alarms channel.

## Value

We are aware of these security alarms and can deal with them or refine them as needed.

## Definition of done

- [ ] alarms integrated for all core accounts
- [ ] tests are green

## Reference

davidkelliott:
See https://github.com/ministryofjustice/modernisation-platform/issues/1334
We want the GuardDuty, SecurityHub and Config Cloudwatch alarms that are already configured to go to a service on pagerduty and then onto our #modernisation-platform-alarms channel.

Module changes:
Having the data lookups in the module breaks when you move add the
module to an account where the integration key secret data lookups don't
exist.
The change means you pass in the integration key rather than the name of
the integration and makes the module is more flexible.

Closes #1334

### DIFF
--- a/.github/workflows/core-vpc-development-deployment.yml
+++ b/.github/workflows/core-vpc-development-deployment.yml
@@ -55,7 +55,7 @@ jobs:
           terraform -chdir="terraform/environments/core-vpc" workspace select "core-vpc-${TF_ENV}"
 
           #RUN TERRAFORM PLAN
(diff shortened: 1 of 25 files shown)


SteveLinden (approved):
Looks OK to me

PR #1345, opened 2021-12-13
Cidr cleanup

gfou-al:
- Remove CIDRs from environment-networks JSON files for Protected and TGW subnets
- Remove variables for terraform/environments/core-vpc/vpc.tf for Protected and TGW subnets


donmasters (approved):
LGTM

PR #1354, opened 2021-12-14
Relax pager duty escalation policy

davidkelliott:
We don't currently have an oncall rota so no one is actively responding
to pagerduty alerts, to make the channel less noisy doing the following:

Only loop to escalate to someone once
No secondary schedule
No auto acknowledge (only auto resolve after 4 hours)

This should reduce the number of slack messages we receive.


SteveLinden (approved):
√

PR #1357, opened 2021-12-16
Suppress sign-in-without-mfa alarms

davidkelliott:
This warning whilst required for security hub, it triggers everytime we
log in using SSO (which does actually use MFA via GitHub). So
suppressing this.

Note, when using the path it needs to be a CEF field, see here - https://support.pagerduty.com/docs/pd-cef
This are case sensitive and should be lower case.

Suppressed alarms can still be seen in pagerduty they just don't create
incidents or get pushed to the slack channel.


ezman (approved):
LGTM

PR #1359, opened 2021-12-20
feature/aws-firewall

donmasters:
- Add AWS Firewall with policy and rule group for Global Protect VPN
- Configure inspection vpc and subnets to pass all external traffic to
  the AWS Firewall
- Update 'internal-inpspection-out' tgw route table with all routes
  (non-prod and prod)

This Firewall solution is required to ensure all traffic external to
Modernisation Platform is passed by stateless firewall rules before
entry.


ezman (approved):
LGTM

Issue #1239, opened 2021-11-02
Scheduled baselines doesn't run on schedule


davidkelliott:
## Expected Behavior

This job https://github.com/ministryofjustice/modernisation-platform/actions/workflows/scheduled-baseline.yml

Should run every Saturday

## Actual Behavior

It's not running on Saturday, the job itself runs fine when manually triggered, but for some reason the scheduling isn't working.

## Steps to Reproduce the Problem

  1.
  1.
  1.

dms1981:
Is it possible that this is getting caught by Github scheduler delays as described here: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events

Specifically, ```Note: The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.```

I can put up a PR to change the time to run a little earlier at, say, 22:30?

dms1981:
Following the merge of #1370 I can see that scheduled runs are taking place once more: https://github.com/ministryofjustice/modernisation-platform/actions/runs/1672383419

PR #1370, opened 2022-01-04
Fix/tf baseline schedule

dms1981:
In an attempt to mitigate #1239 , this PR moves the scheduled time back for 30 minutes. According to the github documentation, scheduling runs to begin on the hour often leads to contention which can delay the start of a run.


davidkelliott (approved):
Looks good to me

PR #1374, opened 2022-01-06
Feature/add dms to pagerduty

dms1981:
Adds `david_sibley` to programmatically generated pager duty users

Does some refactoring in how locals are supplied to `for_each` loops, replacing lists with maps and simplifying some expressions.


davidkelliott (approved):
Looks good to me, thanks for fixing this 👍

Issue #1066, opened 2021-09-08
Create team documentation on how to log into accounts as superadmin


davidkelliott:
## User Story

As a modernisation platform engineer
I need to know which role / account to use when logging on and accessing accounts as the superadmin
So that I can do my job

## Value

Some accounts such as the core-sandbox-dev account are only accessible via the superadmin role.  Whilst we want to cut down on using this role as much as possible, we should still have documentation on how to use it when needed.

## Definition of done

- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
This actually looks like it's been done - https://user-guide.modernisation-platform.service.justice.gov.uk/team-guide/adding-a-new-team-member.html#adding-them-as-a-superadmin-in-our-aws-landing-zone
I'll go through the guidance there and see if it needs any adjustments for clarity then potentially close.

dms1981:
NB - I think I acted before I thought here. I've submitted a very simple PR around the onboarding document with some minor fixes but the goal of this issue is for specific documentation talking about using AWS roles and accounts which I'm not yet familiar enough to answer.

dms1981:
Completed through #1377

PR #1377, opened 2022-01-10
Docs/aws access

dms1981:
Adds documentation to solve #1066 . Discusses how to access and assume role for privileged accounts. Adds some context into team tools used, and how these accounts are created.

### DIFF
--- a/source/team-guide/accessing-aws-accounts.html.md.erb
+++ b/source/team-guide/accessing-aws-accounts.html.md.erb
@@ -1,59 +1,52 @@
 ---
 title: Accessing AWS accounts
-last_reviewed_on: 2021-05-18
-review_in: 2 months
+last_reviewed_on: 2022-01-10
+review_in: 6 months
 ---
 
 # <%= current_page.data.title %>
 
-## What we manage in Terraform
+## How we access AWS accounts
 
-We currently manage the following in Terraform:
+We have two common methods for accessing AWS accounts:
 
-- GitHub resources, including teams, repositories, actions secrets, etc
-- AWS resources, such as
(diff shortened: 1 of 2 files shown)


source/team-guide/accessing-aws-accounts.html.md.erb
```diff
@@ -0,0 +1,53 @@
...
+
+We have two common methods for accessing AWS accounts:
+
+- Through a [common portal](https://moj.awsapps.com/start#/).
+- Through a [superuser account](https://user-guide.modernisation-platform.service.justice.gov.uk/team-guide/adding-a-new-team-member.html#adding-them-as-a-superadmin-in-our-aws-landing-zone).
+
+### Common portal access
+
+Using a web browser, an authenticated user can navigate to the [common portal](https://moj.awsapps.com/start#/) and select
+an AWS account. The portal will allow them to log in to the AWS console with read-only privileges, or to retrieve an
```
davidkelliott, line 19:
```suggestion
an AWS account. The portal will allow them to log in to the AWS console with the relevant privileges, or to retrieve an
```
We have different levels, admin/developer/readonly/etc depending on the persons role and the account

davidkelliott (changes requested):
Looks good thanks for doing this! One minor change to make please :)

davidkelliott (changes requested):
Aha also could you add a link from the home page for the new page here please, cheers

https://github.com/ministryofjustice/modernisation-platform/blob/main/source/index.html.md.erb#L64

dms1981:
> Aha also could you add a link from the home page for the new page here please, cheers
> 
> https://github.com/ministryofjustice/modernisation-platform/blob/main/source/index.html.md.erb#L64

Added in the most recent commit

davidkelliott (approved):

PR #1378, opened 2022-01-10
Add temporary read only users to core accounts

### LINKED ISSUE #1318: IT Health check preparation - Set up user access
## User Story

4x testers (details to be confirmed) require READ ONLY access to the AWS accounts covered in the scope document - https://drive.google.com/file/d/1Nq7jSkl_CjRmmEk-lAj-MsepM7rQJIST/view?usp=sharing

Ideally we should find a way to add them via SSO, but they are not part of the MoJ Github account so we may need to find another mechanism to do this.  There may be a solution to this by finding an answer to this issue - https://github.com/ministryofjustice/modernisation-platform/issues/158

## Value

Testers have access to the platform to test.

## Definition of done

- [ ] testers given access and guidance on how to access
- [ ] testers verify that they can access the required accounts

## Reference

davidkelliott:
This is for an IT health check.  This adds the testers to the
modernisation platform core account and gives them read only access to
that account and the ability to assume a readonly role in the other core
accounts.

The readonly role is created in the required accounts.

This PR should be reverted and all users and roles removed after the
health check.

Closes #1318

### DIFF
--- a/terraform/modernisation-platform-account/iam.tf
+++ b/terraform/modernisation-platform-account/iam.tf
@@ -219,3 +219,9 @@ resource "aws_iam_access_key" "member-ci" {
     create_before_destroy = true
   }
 }
+
+# temporary users for healthcheck
+module "temporary_readonly_users" {
+  source        = "../modules/read-only-users"
+  account_alias = data.aws_caller_identity.current.account_id
+}
--- a/terraform/modules/read-only-users/main.tf
+++ b/terraform/modules/read-only-users/main.tf
@@ -1,0 +1,83 @@
+# You can define readonly usernames, and their keybase key if applicable, below, and this will automatically create:
+# - their account
+# - an attachment to the "readonly" group, which has the IAM policy of readonly and can assume readonly roles, and forced MFA
+# - if a keybase key is provided, it will also create their user login profile
+locals {
+  readonly_users = {
+    "mikey.dunwoody" = "",
+    "brett.thomas"   = "",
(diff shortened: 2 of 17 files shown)


dms1981 (commented):
Looks good to me; can see from a few sample plans that Terraform likes what's proposed too.

terraform/modernisation-platform-account/iam.tf
```diff
@@ -219,3 +219,9 @@ resource "aws_iam_access_key" "member-ci" {
     create_before_destroy = true
   }
 }
+
+# temporary users for healthcheck
```
ezman, line 223: If this block is commented out the whole module is disabled, right?
Might be worth putting a comment here for that. Rather than someone butchering the code to disable the temp access.

terraform/modules/read-only-users/main.tf
```diff
@@ -0,0 +1,83 @@
+# You can define readonly usernames, and their keybase key if applicable, below, and this will automatically create:
+# - their account
+# - an attachment to the "readonly" group, which has the IAM policy of readonly and can assume readonly roles, and forced MFA
+# - if a keybase key is provided, it will also create their user login profile
+locals {
```
ezman, line 5: How does the module react to the user list being empty?

davidkelliott: I expect it will fail, but I think this will be temporary code and removed after the health check (famous last words 😂 )

PR #1380, opened 2022-01-11
Fix secure-baselines not running on scheduled run

davidkelliott:
When the schedule baselines runs on a schedule, the last job,
secure-baselines isn't running, this runs when you manually trigger it.
I believe it is because of the if statement, the `refs/heads/main` part
of this statement isn't triggered it is not a merge into main as it is
with a PR.  Removing this if statement as it's not needed in any case.


dms1981 (approved):
LGTM

Issue #1379, opened 2022-01-10
Connectivity between NOMIS in Azure and NOMIS Test in Mod Platform


davidkelliott:
## User Story

Connectivity between NOMIS in Azure and NOMIS Test in Mod Platform

See thread here:
https://mojdt.slack.com/archives/C01CLTGTXT7/p1641826667005700

Following guidance here - https://docs.google.com/document/d/1Z9ddlu5XDAq6B6_uH2isKbcntx-nod5abNENc72wRv0/edit
We can enable connectivity with the following:

- [x] MP TGW egress route back to NOMIS VNETs
- [x] MP Firewall rule to allow traffic from NOMIS VNETs
- [x] MP NACLs rules updated to allow traffic from NOMIS VNETs
- [ ] Security group changes on the Instances in MP that are being connected to
- [ ] PTTP TGW egress route back to NOMIS VNETs

Ingress routes to the hmpps-test VPC are already present.
NOMIS Azure CIDR range - 10.101.0.0/16

Ports required - 22, 1521, 3872, 4983, 5556, 7001 and 7777

## Value

Enable connectivity between NOMIS Azure and Test NOMIS MP
## Definition of done

- [ ] above changes made
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
Connectivity has now been established and tested for port 1521, any additional ports needed we can easily add to the firewall as part of bau

PR #1396, opened 2022-01-12
Add networking for Azure to NOMIS test

davidkelliott:
This adds the following:
 MP TGW egress route back to NOMIS VNETs
 MP Firewall rule to allow traffic from NOMIS VNETs
 MP NACLs rules updated to allow traffic from NOMIS VNETs

This initial PR is for database connectivity, we will add further
connectivity in later PRs.

Part of issue #1379

### DIFF
--- a/terraform/environments/core-network-services/firewall.tf
+++ b/terraform/environments/core-network-services/firewall.tf
@@ -169,15 +169,15 @@ resource "aws_networkfirewall_firewall_policy" "external_inspection" {
     stateless_fragment_default_actions = ["aws:drop"]
     stateless_rule_group_reference {
       priority     = 101
-      resource_arn = aws_networkfirewall_rule_group.global_protect.arn
+      resource_arn = aws_networkfirewall_rule_group.stateless_rules.arn
     }
   }
 }
 
-resource "aws_networkfirewall_rule_group" "global_protect" {
-  description = "Global_Protect"
+resource "aws_networkfirewall_rule_group" "stateless_rules" {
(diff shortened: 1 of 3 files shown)


dms1981 (approved):
LGTM, but there will come a point where the capacity on the rule group needs bumping, and that will force a destroy/recreate

terraform/environments/core-network-services/firewall.tf
```diff
@@ -228,11 +228,35 @@ resource "aws_networkfirewall_rule_group" "global_protect" {
...
+        stateless_rule { # Azure NOMIS test to MP Nomis database
+          priority = 200
+          rule_definition {
+            actions = ["aws:pass"]
+            match_attributes {
+              source {
+                address_definition = "10.101.0.0/16" # Azure NOMIS Test
+              }
+              source_port {
+                from_port = 1024
```
gfou-al, line 240: If I understand correctly, this configuration allows any unprivileged source port in the range 1024-65535. Just wondering if this might be too open or if there's any reference on why we need it it be so open. Is this to allow NAT-ports?

gfou-al (approved):
1. Refer to comment in code
2. Looking at https://docs.google.com/document/d/1Z9ddlu5XDAq6B6_uH2isKbcntx-nod5abNENc72wRv0/edit# The "Security Groups" configuration. The db server in Nomis test already has a security group rule allowing inbound connection on port 1521. I guess if we were to open up more ports, we would need to add new rules.

PR #1398, opened 2022-01-13
Fix invalid firewall group name

davidkelliott:
It seems that _ is not allowed in firewall group names, changing to -


gfou-al (approved):
looks good

Issue #1383, opened 2022-01-11
Connectivity between Cloud Platform and NOMIS Test in Mod Platform


davidkelliott:
## User Story

Connectivity between NOMIS in Azure and NOMIS Test in Mod Platform

See thread here:
https://mojdt.slack.com/archives/C01A7QK5VM1/p1641834843001700

Following guidance here - https://docs.google.com/document/d/1Z9ddlu5XDAq6B6_uH2isKbcntx-nod5abNENc72wRv0/edit
We can enable connectivity with the following:

MP TGW egress route back to CP
MP Firewall rule to allow traffic from CP to MP NOMIS
MP hmpps-test subnet NACLs rules updated to allow traffic from CP
Security group changes on the Instances in MP that are being connected
PTTP TGW egress route back to CP

Ports required - 1521

Here is an example of the JDBC connection for NOMIS from prison api to azure DB:
jdbc:oracle:thin:@(DESCRIPTION = (ADDRESS = (HOST = 10.40.37.132) (protocol = tcp) (port = 1521))(FAILOVER = YES)(CONNECT_DATA = (SERVICE_NAME = NOMIS_APIRO_TAF)(FAILOVER_MODE = (TYPE = SELECT) (METHOD = BASIC))))

## Definition of done

- [x] CP service able to talk to MP NOMIS DB
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
Complete and tested

PR #1401, opened 2022-01-14
Add networking for Cloud Platform to NOMIS test

gfou-al:
This adds the following:
- MP TGW egress route back to CP
- MP Firewall rule to allow traffic from CP to MP NOMIS
- MP hmpps-test subnet NACLs rules updated to allow traffic from CP

This initial PR is for database connectivity, we will add further
connectivity in later PRs.

Part of Issue #1383


davidkelliott (approved):
LGTM 👍

Issue #1355, opened 2021-12-14
Access denied to the modernisation-platform-logs-cloudtrail-logging bucket for something


davidkelliott:
We are getting unauthorised API alerts in the core logging account.

S3 is trying to access the logging buckets and getting access denied, this is probably an issue with our permissions, unfortunately I can’t see where (other than S3) the request is coming from - https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/cloudtr[…]ccessDenied$2522$26start$3D-10800000 (core-logging account)

We need to work out what is trying to put objects in this bucket and why, if it is supposed to be able to we need to give it the correct permissions.

https://moj-digital-tools.pagerduty.com/alerts/Q08ZQ5IE14B7XP

dms1981:
From a fairly deep investigation detailed in this slack message - https://mojdt.slack.com/archives/C013RM6MFFW/p1642171710172900 - it looks like these alerts are the result of the S3 Server Access Logs for `modernisation-platform-logs-cloudtrail` writing to a KMS-encrypted destination. Unfortunately there doesn't appear to be any way to allow the account that writes access logs to safely have access to the encryption key, so we'll need to switch the logging buckets to use AWS:SSE encryption.

PR #1402, opened 2022-01-17
added logging.s3.amazonaws.com to kms principals

dms1981:
This PR solves #1355 - from what I could tell the AWS S3 logging service could not access the KMS key used to encrypt the bucket. I previously investigated switching the bucket encryption to use AWS:SSE instead of AWS:KMS which solved the issue, but in discussion with AWS support found that the principal `logging.s3.amazonaws.com` would also work.
While the documented solution is to switch the bucket to AWS:SSE, I think that allowing the principal access to the logging bucket key is more in line with how we have delivered encryption to logging buckets.


davidkelliott (approved):
Looks like a good approach 👍

PR #1418, opened 2022-01-26
replace / with : to fix invalid principal identifier for KMS policy

dms1981:
This PR resolves an issue with the creation of KMS keys; namely an invalid principal which was down to a `/` where a`:` was appropriate.


davidkelliott (approved):
👍

PR #1419, opened 2022-01-27
Feature/ppud production

dms1981:
PR to create ppud production environment as detailed here: #1376 

Not to be merged until a corresponding PR has been raised and completed for adding the team slug to terraform

PR #1420, opened 2022-01-27
added ppud-replacement-devs to application_teams

dms1981:
Add local value for new github slug, allowing merge of #1419


jackstockley89 (approved):
LGTM

PR #1423, opened 2022-01-27
Decrease scope of permissions for state bucket

davidkelliott:
The permissions to read the state buckets were introduced to allow
developers to run plans locally, however, they only need access to the
top level state file and the member environment state files for this.


dms1981:
LGTM. We may want to revisit with conditions if we want more fine grained access, but this will work.

dms1981 (approved):

Issue #1376, opened 2022-01-06
New PPUD environment for production data migration


davidkelliott:
## Environment details

There is a new PPUD service in Cloud Platform that requires data from a service via PSN.  Traffic cannot be routed to Cloud Platform directly due to CIDR range clashes.  This environment is to enable data to be transferred to an EC2 which can then be uploaded from MP to CP.

### Application Name

ppud

### Environments

- [ ] Development
- [ ] Test
- [ ] Preproduction
- [x] Production

### Tags

tag | value
--- | ---
application | ppud
business-unit |  hmpps
owner | 

#### Subnet sets

- [x] General - this application will use the same subnet as other applications in your business unit
- [ ] Isolated - this application has highly sensitive data which must have its own isolated subnet

#### How do users connect to the application?

- [ ] Over the public internet
- [ ] With a purple cabled device (please give details)
- [ ] With a MoJ Official device

#### Connectivity to other applications or external parties

PSN connectivity required

## Additional features

- [ ] Additional VPC Endpoints
- [ ] Extended DNS Zones
- [ ] Other - please specify

dms1981:
PPUD production environment now created and able to access HMPPS VPC resources through RAM principal association

PR #1431, opened 2022-01-28
Fix/vpc deployment workflows

dms1981:


dms1981:
NB. This PR relates to #1376 - it ensures that directories for accounts are properly scanned

davidkelliott (approved):
👍

PR #1432, opened 2022-01-28
Add automation to create collaborator IAM users

### LINKED ISSUE #158: New feature to add external collaborators as IAM users
At MOJ we use a mix of external collaborators and GitHub organisation members to define access to repositories. With the current AWS SSO implementation, we import all organisation members and team affiliations. GitHub external collaborators aren't part of a team, but are part of repositories; so we need to explore options for importing external collaborators and a "team membership" to provide access to AWS accounts.

davidkelliott:
Closes #158

This adds functionality to add external collaborators who are not part
of the ministryofjustice GitHub organisation to the Modernisation
Platform. See the user guidance added in this PR for the full process to
set up users.

Overview:

- collaborators.json - json file where all collaborators and their
account associations can be added
- OPA tests to ensure the above json is in the correct format
- new roles added to all core and member accounts via delegate access -
read-only and developer, collaborators will only be able to assume these
roles for their listed accounts
- new collaborators module to set up users
- new IAM group and policy for collaborator users
- call the new module from the modernisation-platform core account
- minor change to github create environments script (we are not amending
this script to add collaborators but a small fix was noticed)


dms1981 (approved):
I don't have much value to offer here beyond praise; code looks good, tests cover appropriate scope, and documentation is clear and comprehensible. IAM policies look fine at first glance too.

Issue #1382, opened 2022-01-11
Connectivity between PSN and PPUD production environment


davidkelliott:
## User Story

The MP ppud-production account, private subnets, need to be able to access:

www.ppud.gsi.gov.uk
51.247.2.115 port 443 

## Value

Enable PPUD replacement team to get data from PPUD

## Definition of done

- [ ] routing and firewall changes in place
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
Reopened as corresponding change required in https://github.com/ministryofjustice/deployment-tgw to route traffic back to Modernisation Platform TGW

dms1981:
https://github.com/ministryofjustice/deployment-tgw/pull/27 <- corresponding PR in the deployment-tgw repository

dms1981:
All changes made

dms1981:
Some progress made here: we discovered that although routing was in place to deliver traffic to the PTTP transit gateway it was not being correctly routed from there. We raised a PR - which was completed - to add the necessary routes to the relevant route table.
We then discovered that our traffic was reaching a set of security appliances where it was being accepted but not correctly routed back. We supplied the relevant administrator with our needs and are waiting to hear back from them.

dms1981:
Still waiting on routing change. We have chased, but do not have a date for the change yet.

dms1981:
As a follow-up @dazoakley and I have been in contact via email with the technical services operations manager for networks, and are awaiting a response.

dms1981:
According to @dazoakley the relevant change was scheduled to take place on the evening of the 22nd so I'll check in with him to chase this today.

dms1981:
PSN Palo change was successful; had to work around issues with inspection vpc temporarily pending a support case with AWS / further internal discussion in the team.
#1638 refers to the workaround

PR #1433, opened 2022-02-01
added route to ppud psn endpoint via pttp tgw peering

dms1981:
Solves #1382 by adding route via PTTP TGW and firewall rule for returning traffic
Check of existing routes and acl entries in HMPPS VPC and TGW shows that this should meet the requirements for this story.


davidkelliott (approved):
Looks good, are you adding the route to the PTTP gateway as well like - https://github.com/ministryofjustice/deployment-tgw/pull/26/files
?

dms1981:
While this PR has issues with static code analysis, they're related to an upgrade to our static analysis action not the PR itself.

PR #1442, opened 2022-02-03
Feature/xp preprod

dms1981:
Creates a preoproduction environment for Xhibit Portal, allocates range for HMCTS preproduction


ezman (approved):
LGTM

environments/xhibit-portal.json
```diff
@@ -10,6 +10,15 @@
         }
       ]
     },
+    {
```
davidkelliott, line 13: This file should be in a separate PR and merged in first to ensure the AWS account is created before any networking takes place.

policies/networking/expected.rego
```diff
@@ -13,7 +13,8 @@ expected =
     "cica-development": "10.233.2.64/26",
     "hmpps-production": "10.233.2.128/26",
     "hmcts-development": "10.233.2.192/26",
-    "hmcts-production": "10.233.3.0/26"
+    "hmcts-production": "10.233.3.0/26",
```
davidkelliott, line 16: Actually these 2 blocks and associated tests should have been removed, we now allocate transit gateway and protected subnets from the general subnet set, we used to have to do this manually hence the need to check them, but this is no longer required.

davidkelliott (changes requested):

policies/networking/expected.rego
dms1981: Those parts have been removed from the latest commit - thanks for the comment!

environments/xhibit-portal.json
dms1981: That's also been cleaned up as it's not relevant for this stage - thanks for the comment!

davidkelliott (approved):
👍

PR #1443, opened 2022-02-03
Delete unneeded OPA tests

davidkelliott:
After refactoring the networking code to remove the need to manually
assign protected and transit gateway subnets, adjustments to these tests
were missed.  Also fixing the call as the jq was looking for no existent
keys now.


dms1981 (approved):
LGTM; should prevent some issues in future.

PR #1449, opened 2022-02-07
Fix/iam user module bump

dms1981:
Accounts for some checkov checks that refer to open target resources, and accounts for the bump in iam-user module


dms1981:
TFSEC will need some investigation, but is unrelated to this specific PR.

davidkelliott (approved):
👍  thanks David

PR #1451, opened 2022-02-07
Add permissions needed to run plans as collab dev

davidkelliott:
These permissions are required to enable collaborators with developer
access to run terraform plans locally.


dms1981 (approved):
LGTM

PR #1454, opened 2022-02-07
Amend templates

SteveLinden:
Add data.tf (similar to main from modwenisation-platform-environments) and amended locals.tf


davidkelliott (approved):
Looks good thanks Steve!

PR #1457, opened 2022-02-09
Restrict permissions of CICD user

dms1981:
This PR exists to tighten what the CICD user can do. We had a recent case where a team was able to create and add a role policy to the CICD user but not remove it. We probably don't want to allow the unexpected expansion of what the CICD user can do, so this PR adds a statement preventing the CICD user from making changes to itself (eg, attaching extra role policies).

### DIFF
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -1,11 +1,10 @@
 locals {
-
-
   account_name = try(regex("^bichard*.|^remote-supervisio*.", terraform.workspace), replace(terraform.workspace, regex("-[^-]*$", terraform.workspace), ""))
-
-
   account_data = jsondecode(file("../../../../environments/${local.account_name}.json"))
+}
 
+data "aws_iam_user" "cicd_member_user" {
+  user_name = "cicd-member-user"
 }
 
 module "cross-account-access" {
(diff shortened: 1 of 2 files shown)


terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -130,6 +129,28 @@ data "aws_iam_policy_document" "member-access" {
...
+      "iam:DeleteRolePermissionsBoundary",
+      "iam:DeleteRolePolicy",
+      "iam:DetachRolePolicy",
+      "iam:PutRolePermissionsBoundary",
+      "iam:PutRolePolicy",
+      "iam:UpdateAssumeRolePolicy",
+      "iam:UpdateRole",
+      "iam:UpdateRoleDescription"
+    ]
+    condition {
```
davidkelliott, line 147: I don't think this condition is needed is it?

dms1981: Pulled the condition as discussed

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -1,11 +1,10 @@
 locals {
-
-
   account_name = try(regex("^bichard*.|^remote-supervisio*.", terraform.workspace), replace(terraform.workspace, regex("-[^-]*$", terraform.workspace), ""))
-
-
   account_data = jsondecode(file("../../../../environments/${local.account_name}.json"))
+}
 
+data "aws_iam_user" "cicd_member_user" {
```
davidkelliott, line 6: Just thinking...this data look up will fail if we ever rebuild this from scratch as the user won't exist yet.  Would it be possible to get the arn from the module where it is created on line 20 please?  You'll have to add an output for the arn to the module.

davidkelliott (changes requested):

terraform/environments/bootstrap/delegate-access/iam.tf
dms1981: @davidkelliott - I had considered that as an alternative, yes. I will explore that now. Should have a PR up that covers that change shortly.

davidkelliott (approved):
Awesome! Cheers

PR #1458, opened 2022-02-10
Fix/cicd member user

dms1981:
Removed module output and utilises partial wildcard match for IAM Deny policy


davidkelliott (approved):
:shipit:

PR #1465, opened 2022-02-14
Resolve tfsec errors

dms1981:
* Adds a handful of exclusions for wildcards
* Refactors some exclusions
* Adds a public access block to control access to ACM:PCA bucket based on [AWS Guidance](https://aws.amazon.com/about-aws/whats-new/2021/05/aws-certificate-manager-private-certificate-authority-supports-storing-crls-in-private-s3-buckets/)
* Refactors configuration for ACM bucket to be compliant with v4.0 of Terraform AWS Provider
* Sets AWS provider to use only releases in the v4.x train


davidkelliott (dismissed):

davidkelliott (approved):
🥇

PR #1468, opened 2022-02-15
Prevent auto upgrade to AWS provider 4.0.0

davidkelliott:
Currently our versioning is as follows:

  version = ">= 3.47.0"
This means that the pipelines will always get the latest version (this doesn’t happen locally as you have to run an init upgrade if you’ve run init previously), this is causing issues now when we move to version 4 where there are breaking changes.

If we change the way we version to only allow patch and minor changes, but not major version changes:

  version = ">= 3.47.0, < 4.0.0"

This will still allow minor upgrades to happen automatically, but any breaking changes won’t.
Then we can go through each tf and upgrade and make the changes required manually.

### DIFF
--- a/terraform/environments/analytical-platform-data/versions.tf
+++ b/terraform/environments/analytical-platform-data/versions.tf
@@ -1,7 +1,7 @@
 terraform {
   required_providers {
     aws = {
-      version = ">= 3.47.0"
+      version = ">= 3.47.0, < 4.0.0"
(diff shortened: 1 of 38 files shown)


gfou-al (approved):
There are static code analysis errors. I thought we had fixed all of them for the modernisation-platform repository. From a quick look, they don't seem to be related.

PR #1471, opened 2022-02-16
Widen route to PSN on TGW

dms1981:
This PR widens the range of addresses being routed from the ModPlatform TGW to the PTTP TGW for PSN endpoints


davidkelliott (approved):
Looks good to me

PR #1472, opened 2022-02-16
Add guidance for adding connectivity to mp

davidkelliott:


SteveLinden (approved):
Some lines are a little long but other than that it's OK

PR #1473, opened 2022-02-16
Add Modernisation Platform Engineer role to SSO

davidkelliott:
This role allows read only access and enables the user to log support
tickets with AWS.

This is designed to be used as a daily troubleshooting role, rather than
full admin access.  We want to avoid using admin access where possible
as it make accidental or deliberate click ops possible.

Over time we can add more permissions to this role if we feel we need
them. This is part of a longer term strategy to only use admin role as
breakglass emergency roles.


dms1981 (approved):
LGTM; observed successful outcome of test application

PR #1474, opened 2022-02-16
Add role for shared services account

davidkelliott:
This adds a new role to give readonly permissions, EC2 image builder
imaage start permissions and support read only permissions to any users
in accounts in the modernisation platform organisational unit.

This is required to enable members to start and troubleshoot their image
builder pipelines.


dms1981 (approved):
LGTM

PR #1477, opened 2022-02-18
Fix Cloudtrail logs lambda and athena

davidkelliott:
Started out just looking to fix some open permissions highlighted by
SCA.

Then noticed the lambda didn't have permissions to write to logs.

Then once there were logs saw that the lambda was failing as didn't have permissions.

The method of getting permissions by assuming a role was broken, and now unnecessary as the secret is shared across accounts so changed that.

Then fixed the lambda only to see the Athena queries it was running didn't work. 

So fixed them as well.

That will teach me to try to fix an SCA warning.


davidkelliott:
Deployed already as needed to test the lambda

dms1981 (approved):
Looks good to me; discussed with @davidkelliott

PR #1485, opened 2022-02-22
Add NACL rules to allow inbound traffic

zuriguardiola:
These NACL changes were implemented in preprod and prod and now need to 
be done in development as the majority of testing is being done there.


davidkelliott (approved):
👍

PR #1486, opened 2022-02-22
Remove MP OUs SCP prevent root access policy

### LINKED ISSUE #1324: Remove deny all on mp SCPs and update delete account guidance
## User Story

We are now blocking root account access at the P&A OU level, this means that the blocking root access on our SCPs is no longer needed.  Also it means we no longer need to detach and attach policies, we can just move the account to the closed account ou before deletion.  So the new process will look like this:

Remove account from terraform state
Move account to closed account ou
Delete account
Run terraform

## Definition of done

- [ ] readme has been updated
- [ ] tests are green

## Reference

davidkelliott:
This is now implemented at the P&A level so inherited from therefore no
longer required at this level. Updating the userguidance as well to
reflect this.

(also fixed a typo)

Closes #1324

### DIFF
--- a/source/team-guide/deleting-an-environment.html.md.erb
+++ b/source/team-guide/deleting-an-environment.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: Deleting an environment (AWS account)
-last_reviewed_on: 2021-10-27
-review_in: 3 month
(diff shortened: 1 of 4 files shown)


dms1981 (approved):
LGTM

PR #1513, opened 2022-03-04
added stub github usernames where not supplied

dms1981:
This temporarily fixes an issue with the OPA tests for collaborators.json
When a field is picked up as empty, the field fails a test condition. However, we don't always know if we will be supplied with a github username. As a temporary measure I've added comments into these fields

### DIFF
--- a/collaborators.json
+++ b/collaborators.json
@@ -20,7 +20,7 @@
     },
     {
       "username": "lewis.smith",
-      "github-username": "",
+      "github-username": "#no-username-supplied",
       "accounts": [
         {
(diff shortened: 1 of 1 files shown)


davidkelliott (approved):
I think this is a good work around, we need to ensure we record these usernames when we are adding them as collaborators to repos so explicitly confirm that we don't have a username is fine

dms1981:
I think from a good-test-practice perspective I'd want to warn when fields are empty rather than fail (as there are legitimate cases where we don't have github usernames supplied at first), but yeah, this solves the test fails in the short term.

PR #1520, opened 2022-03-09
Skip CKV_GIT_4 where check not appropriate

dms1981:
Adds a Checkov skip after discussion and analysis


dms1981:
From my understanding, the failure is triggered by hard-coded placeholder values. While plaintext secrets are supplied during CI/CD runs, these are encrypted at reset mitigating the impact of this check.

SteveLinden (approved):

PR #1536, opened 2022-03-14
NACL tightening

gfou-al:
Do not allow private and data subnets in core-network-services account communicating with public internet addresses.


davidkelliott (approved):
Looks good to me, we have identified route table recreations in the core-security, core-logging and core-shared-services accounts.  This will lead to temporary connectivity issues in these accounts, but we don't believe there will be any significant impact.  There are no route table changes in the core-network-services account.

PR #1553, opened 2022-03-16
Allow outgoing traffic on 443

davidkelliott:
The restriction of all traffic to and from 0.0.0.0/0 caused the
core-shared-services calls to S3 Gateway endpoints to fail.  This is
because S3 Gateway endpoints use public IPs (even though the traffic is
routed privately
https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html)
This change adds in rules to allow outgoing traffic to port 443, and
incoming traffic from S3 (on ephemeral ports
https://aws.amazon.com/premiumsupport/knowledge-center/connect-s3-vpc-endpoint/#Network_ACL_rules)

### DIFF
--- a/terraform/modules/vpc-hub/main.tf
+++ b/terraform/modules/vpc-hub/main.tf
@@ -70,14 +70,16 @@ locals {
     for rule in local.nacl_rules : join("-", values(rule)) => rule
   }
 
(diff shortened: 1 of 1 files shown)


dms1981 (dismissed):
LGTM

Issue #1559, opened 2022-03-17
Bug: users cannot delete Route 53 resolver endpoints


dms1981:
## Expected Behavior

Given a resource configured as follows:
```
resource "aws_route53_resolver_endpoint" "main" {
  provider = aws.core-vpc

  name      = "domain-name"
  direction = "OUTBOUND"

  security_group_ids = [
    aws_security_group.name.id
  ]

  ip_address {
    subnet_id = data.aws_subnet.name.id
  }
}
```
Terraform ought to be able to destroy this resource in the event of a change being made requiring recreation.

## Actual Behavior

As seen here: https://github.com/ministryofjustice/modernisation-platform-environments/runs/5585846379?check_suite_focus=true an attempt to destroy a resolver rule leads to an AWS API error as follows:
```
Error: error deleting Route53 Resolver endpoint (rslvr-out-11fa98e330ca4c829): AccessDeniedException: User: arn:aws:sts::453136268281:assumed-role/member-delegation-hmcts-preproduction/1647520544149370034 is not authorized to perform: route53resolver:DeleteResolverEndpoint on resource: arn:aws:route53resolver:eu-west-2:453136268281:resolver-endpoint/rslvr-out-11fa98e330ca4c829 because no identity-based policy allows the route53resolver:DeleteResolverEndpoint action
```

## Steps to Reproduce the Problem

  1. Create resolver endpoint through terraform in member account (modernisation-platform-environments/terraform/environments/*)
  1. Delete resolver endpoint from terraform code
  1. Run terraform

Issue #1558, opened 2022-03-17
Bug: users cannot delete Route 53 resolver rules


dms1981:
## Expected Behavior

Given a resource configured as follows:
```
resource "aws_route53_resolver_rule" "main" {
  provider = aws.core-vpc

  domain_name          = "$domain-name"
  name                 = "$domain-name"
  rule_type            = "FORWARD"
  resolver_endpoint_id = aws_route53_resolver_endpoint.main.id

  target_ip {
    ip = aws_instance.name.private_ip
  }
}
```

Terraform ought to be able to destroy this resource in the event of a change being made requiring recreation.

## Actual Behavior

As seen here: https://github.com/ministryofjustice/modernisation-platform-environments/runs/5585027156?check_suite_focus=true#step:4 an attempt to destroy a resolver rule leads to an AWS API error as follows:

```
Error: error deleting Route 53 Resolver rule association: AccessDeniedException: User: arn:aws:sts::*:assumed-role/*/* is not authorized to perform: route53resolver:DisassociateResolverRule on resource: arn:aws:route53resolver:eu-west-2:*:resolver-rule/* because no identity-based policy allows the route53resolver:DisassociateResolverRule action
```

## Steps to Reproduce the Problem

  1. Create resolver rule through terraform in member account (modernisation-platform-environments/terraform/environments/*)
  1. Delete resolver rule from terraform code
  1. Run terraform

PR #1617, opened 2022-03-23
Add permissions to IAM role for member-delegation

dms1981:
Adds new `route53resolver` permissions to solve #1558 and #1559 
These will need to be tested - most likely through creation and deletion of resolver endpoints in sprinkler account.
This will allow customers to make changes to resolver endpoints and rules as required.


davidkelliott (approved):
Looks good to me

PR #1618, opened 2022-03-24
Document the process of requesting public certificates

gfou-al:
- Document the process of requesting public certificates for domains that are not managed by the Modernisation Platform team
- Linux instructions are included in the existing documentation for which links are provided
- Windows instructions are included in a separate how to article
- Moved the Windows page to the user-guide folder and added a link to it on the index page under ### How to guides


zuriguardiola (dismissed):

source/user-guide/certificate-import-windows.html.md.erb
```diff
@@ -0,0 +1,109 @@
...
+
+Use PowerShell to execute the `certreq -new` command.
+
+(The `performance-hub` repo has a PowerShell script which wraps the `certreq -new` command to generate a CSR and save it to a text file: `.\lib\generateCSR.ps1`)
+
+## 2. Request certificate
+
+Email certificates@digital.justice.gov.uk with the CSR(s) and domain(s), and they will send the certificates.
+
+Note these will be just text files with the certificate data - no private key, and no certification chain (this is important later).
```
dms1981, line 40: What do you mean by "text files with the certificate data" - is this just the public key for the cert being requested?

davidkelliott (changes requested):
Looks good, I think though that users will not necessarily check the concepts pages if they want to do a Linux import.  I think it would be better to have the user-guidance for both, and have a linux section at the top with the link to how to do the import.  So something like - 
user-guide/certificate-import.html
```
## Intro
...
## Linux
...
## Windows
...

```
I just think that as a user I'm not going to check the concepts section if I want to know how to do something I would check the user guidance and how to guides.  I think that and addressing David's comment then it's good to go 👍

source/user-guide/certificate-import-windows.html.md.erb
gfou-al: This is a PEM-formatted identity certificate that has been digitally signed using the private key of the certificate authority.

To clarify on the above, I have now replaced the comment with the following note:

Note these will be plain text files with the body of the PEM-formatted server certificate from your certificate authority - no private key, and no certification chain (this is important later). The body of the PEM-formatted server certificate is an identity certificate that has been digitally signed using the private key of the certificate authority.

I hope this makes it somehow clearer.

davidkelliott (approved):
Lgtm

PR #1626, opened 2022-03-25
added data calls for data subnets to template data.tf

dms1981:


davidkelliott (approved):

davidkelliott (commented):
Thanks for changing the names to be consistent!⭐️

PR #1638, opened 2022-03-30
Temporary bypass for inspection VPC

dms1981:


davidkelliott (changes requested):
Could you put a brief description of the issue we’re trying to work through with this in the commit please?  Just in case this code ends up staying

dms1981:
> Could you put a brief description of the issue we’re trying to work through with this in the commit please? Just in case this code ends up staying

Added an appropriate description. Thanks for the reminder!

davidkelliott (approved):

Issue #1632, opened 2022-03-29
Add AmazonEC2ReadOnlyAccess to cicid-member-policy


gfou-al:
## User Story

We need to add the AmazonEC2ReadOnlyAccess IAM policy to the cicid-member-policy to allow cicd-member-user accessing information about EC2 instances running under the member account.

In particular, the Nomis team is setting up a Prometheus server pod in the Cloud Platform cluster. The pod needs to scrape the mod-platform EC2 instances using a service account (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY) with the AmazonEC2ReadOnlyAccess policy. Example configuration:

```
      - job_name: 'ec2_discovery'
        scrape_interval: 1m
        scrape_timeout: 1m
        ec2_sd_config:
          - region: eu-west-2
            profile: prometheus-CP
            access_key: $PROMETHEUS_AWS_ACCESS_KEY # AWS_ACCESS_KEY_ID for prometheus-CP
            secret_key: $PROMETHEUS_AWS_SECRET_KEY # AWS_SECRET_ACCESS_KEY for prometheus-CP
            port: 9100
```

To allow automated access to your AWS account the Modernisation Platform provides a "CI/CD (Continuous integration / continuous delivery)" user - cicd-member-user, as documented here: https://user-guide.modernisation-platform.service.justice.gov.uk/concepts/sdlc/user-workflow.html#deploying-applications

We need to enhance the cicd-member-user with the additional capability, by adding the AmazonEC2ReadOnlyAccess IAM policy to the cicid-member-policy.

If you need further information about the use case, please reach out to David Teeling.

## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

PR #1662, opened 2022-04-06
Add EC2 readonly policy to member cicd user

davidkelliott:
This is to allow cicd-member-user accessing information about EC2
instances running under the member account


davidkelliott:
Closes #1632

PR #1676, opened 2022-04-12
Allow member-ci user to iam:ListAccountAliases

gfou-al:
Allow member-ci user to iam:ListAccountAliases. Related error:
Switched to workspace "sprinkler-development".
Error: failed get account alias: AccessDenied: User: arn:aws:iam::946070829339:user/member-ci is not authorized to perform: iam:ListAccountAliases on resource: *
	status code: 403, request id: 57ff0700-601c-4471-bcc2-002773ad5e73
Error: Process completed with exit code 255.

### DIFF
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -65,6 +65,7 @@ data "aws_iam_policy_document" "member-access" {
       "glue:*",
       "guardduty:get*",
       "iam:*",
+      "iam:ListAccountAliases",
       "kms:*",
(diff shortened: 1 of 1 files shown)


davidkelliott (changes requested):
Those permissions are already covered in iam:*
I think this is the wrong policy let me check...

davidkelliott:
Actually this is the right policy but the issue that you are having is that you are trying to run nuke as the member-ci user, you will need to assume the `MemberInfrastructureAccess` role for the account you are running nuke in

gfou-al:
Closing this PR as per above comment by Dave.

PR #1696, opened 2022-04-20
Added nomis-production to the hmpps-production.json

SteveLinden:
Include the new nomis environment in the hmpps-production.json file.


SteveLinden:
Change ported in a later pull

PR #1698, opened 2022-04-21
Nomis production firewall changes

SteveLinden:
Put in production firewall changes for Nomis.
Also added nomis-production to hmpps-production.json where it had been missed.

### DIFF
--- a/terraform/environments/core-network-services/firewall.tf
+++ b/terraform/environments/core-network-services/firewall.tf
@@ -299,6 +299,122 @@ resource "aws_networkfirewall_rule_group" "stateless_rules" {
             }
           }
         }
+        stateless_rule {
+          priority = 100
+          rule_definition {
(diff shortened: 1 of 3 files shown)


terraform/environments/core-network-services/firewall.tf
```diff
@@ -299,6 +299,122 @@ resource "aws_networkfirewall_rule_group" "stateless_rules" {
             }
           }
         }
+        stateless_rule {
+          priority = 100
```
dms1981, line 303: This probably needs changing as it will conflict with an earlier rule that's got the same priority

dms1981 (changes requested):
I think you'll want to check your rule priorities here to align them with the priorities that have already been set

zuriguardiola (approved):

dms1981 (approved):

PR #1721, opened 2022-04-29
Allow the member-ci user to access the nuke_account_ids secret 

gfou-al:
to resolve the error:

An error occurred (AccessDeniedException) when calling the GetSecretValue operation: User: arn:aws:iam::<account id>:user/member-ci is not authorized to perform: secretsmanager:GetSecretValue on resource: nuke_account_ids because no identity-based policy allows the secretsmanager:GetSecretValue action


davidkelliott (approved):
👍

PR #1730, opened 2022-04-29
Shared services workflow wasnt triggering on apply

davidkelliott:
Fix to allow this to run the apply job as well after the new environment
workflow runs.


zuriguardiola (dismissed):

dms1981 (approved):
LGTM

PR #1733, opened 2022-05-05
New example account to use for example code

SteveLinden:
Setting up account that will be used for example code

### DIFF
--- a/environments/example.json
+++ b/environments/example.json
@@ -12,7 +12,7 @@
     }
   ],
   "tags": {
-    "application": "modernisation-platform",
+    "application": "example",
     "business-unit": "Platforms",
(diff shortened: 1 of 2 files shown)


environments/example.json
```diff
@@ -0,0 +1,20 @@
...
+      "access": [
+        {
+          "github_slug": "modernisation-platform",
+          "level": "developer"
+        }
+      ]
+    }
+  ],
+  "tags": {
+    "application": "modernisation-platform",
```
davidkelliott, line 15:
```suggestion
    "application": "example",
```

davidkelliott (changes requested):
Small change to the name for the tags please just so we are clear everything is for the example account

davidkelliott (approved):

PR #1735, opened 2022-05-05
Adding the CIDR for example-development account

SteveLinden:
Adding the CIDR for the new example-account


davidkelliott (changes requested):
This won't need a new CIDR allocation as it can be put under platforms development - general

SteveLinden:
> 

I'll close the pull request and amend the code. Still need the expected rego change though?

davidkelliott (approved):
LGTM

PR #1737, opened 2022-05-05
updated script to include gomod dependency checks for tests

dms1981:
* Add some content to `generate-dependabot-file.sh` to find and populate entries for `go` dependencies such as those used in tests


davidkelliott (approved):
👍

PR #1789, opened 2022-05-18
updated review date for subnet allocation

dms1981:


davidkelliott (changes requested):
I think this one needs some updating as we changed the way we do the subnets when we redid the cidr ranges, new images are here - https://app.diagrams.net/#G1w8dZs5vZOoy0dvCnRr5P6GHc3VQRdwLq on the shared-VPC tab

davidkelliott (approved):

PR #1809, opened 2022-05-24
Fixing code to remove "name" from example.json - duplicate tages when…

SteveLinden:
… building bastion
Very simple change - replacing "name" with "instance" to try to cure the issue


davidkelliott (changes requested):
I would just delete this name tag altogether

PR #1923, opened 2022-06-21
Feature/hmpps ses endpoints

dms1981:
* Create SES endpoints in HMPPS dev and HMPPS test
* Update reference to member VPC module

### DIFF
--- a/environments-networks/hmpps-development.json
+++ b/environments-networks/hmpps-development.json
@@ -13,7 +13,10 @@
   },
   "options": {
     "bastion_linux": true,
-    "additional_endpoints": [],
+    "additional_endpoints": [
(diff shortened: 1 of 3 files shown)


davidkelliott (dismissed):
Has the new module been tested with sprinkler or another account? changes to this module will affect all VPCs including production ones.

dms1981:
> You're right, it's not been tested in Sprinkler yet - I'll put it through there now and see the behaviour first-hand. Expect a new PR!

dms1981:
Testing for this was successful, and was discussed in Slack. Will re-run the GH actions prior to requesting a review again.

zuriguardiola (approved):

PR #1961, opened 2022-07-04
Allow access to pagerduty secret for other accounts

dms1981:
* Add new KMS key an alias for pagerduty secret
* Create cross-account access policy for pagerduty secret
* Add cross-account access policy to pagerduty secret
* Create IAM policy for new KMS key
* Attach policy to new KMS key

### DIFF
--- a/terraform/pagerduty/kms.tf
+++ b/terraform/pagerduty/kms.tf
@@ -4,6 +4,6 @@ resource "aws_kms_key" "pagerduty" {
 }
 
 resource "aws_kms_alias" "pagerduty" {
-  name          = "pagerduty-secret"
+  name          = "alias/pagerduty-secret"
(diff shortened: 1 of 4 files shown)


terraform/pagerduty/kms.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_kms_key" "pagerduty" {
+  tags = local.tags
+}
+
+resource "aws_kms_alias" "pagerduty" {
+  target_key_id = aws_kms_key.pagerduty.id
```
davidkelliott, line 6: Do you want to add in a name here to be consistent with the other kms alias resources?

dms1981: I do! Good spot!

davidkelliott (approved):
Thank you 🙏

PR #1973, opened 2022-07-06
Open NACLs on hmpps-production

davidkelliott:
This is to allow traffic from the Azure noms-mgmt-live vnet to the
hmpps-production environment.


dms1981 (approved):
LGTM!

PR #1974, opened 2022-07-07
Fix equip SES user policy

davidkelliott:
The workflow was failing with:

```
Error: Missing required argument
  on iam.tf line 12, in resource "aws_iam_user_policy" "email_policy":
  12: resource "aws_iam_user_policy" "email_policy" {
The argument "policy" is required, but no definition was found.
```

Added the policy to match the exising inline policy and imported it into
terraform, ran apply and no changes are required in production.

### DIFF
--- a/terraform/environments/equip/iam.tf
+++ b/terraform/environments/equip/iam.tf
@@ -10,6 +10,18 @@ resource "aws_iam_access_key" "email" {
 }
 
 resource "aws_iam_user_policy" "email_policy" {
-  name = format("%s-%s-email_policy", local.application_name, local.environment)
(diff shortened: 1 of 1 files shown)


SteveLinden (approved):
LGTM

dms1981 (commented):
This policy is also covered in a data call so it could be referenced like so:
`policy = data.aws_iam_policy_document.email.json`

davidkelliott:
> This policy is also covered in a data call so it could be referenced like so: `policy = data.aws_iam_policy_document.email.json`

Didn't see that! Updated to use that instead

dms1981 (approved):
LGTM

PR #2000, opened 2022-07-13
Refactor TGW attachment tagging

dms1981:
* Use a `for_each` loop to take multiple tags in for TGW VPC attachments
* Refactor where the attachment module is called to supply both the name, and the local tag values


davidkelliott (approved):
Great work, this is really good thanks 👏

Issue #1988, opened 2022-07-12
Add Cloudwatch alarms to Transit Gateway


dms1981:
## User Story

As a modernisation platform engineer 
I want timely alerts issued when a transit gateway VPC attachment has no traffic passing over it
So that I am informed of potential outages for platform customers

## User Type(s)

Modernisation Platform Engineer - will be alerted via PagerDuty
Modernisation Platform Customer - will received support for outages faster

## Value

We don't want to rely on customers reporting issues, so in the unlikely event that a VPC is associated from our Transit Gateway then we want our own monitoring in place.
Having this monitoring will improve the service we offer to our customers and ensure their confidence in our abilities.

## Questions / Assumptions / Hypothesis

* Cloudwatch metric alarms for `TransitGatewayAttachment` dimension against `PacketsIn` metric will be sufficient
* 0 bytes over 5 minutes will be an appropriate alarm
* All VPC attachments to the `Modernisation Platform Transit Gateway` will be monitored

### Hypothesis 
If we use CloudWatch to report low traffic alarms to PagerDuty
Then we will be aware of potential outages such as a VPC attachment being disassociated from a VPC

## Definition of done

- [x] cloudwatch alarms created in `core-network-services`
- [x] cloudwatch alarms linked to SNS topic
- [x] SNS topic reports to pagerduty integration
- [x] another team member has reviewed
- [x] SCA tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)
[CloudWatch metrics for your transit gateways](https://docs.aws.amazon.com/vpc/latest/tgw/transit-gateway-cloudwatch-metrics.html)

PR #2002, opened 2022-07-13
Add Cloudwatch Alarm & PagerDuty integration for TGW Attachments

dms1981:
* Create a rich list of Transit Gateway attachments matching `is-production:true`
* Use the list of production Transit Gateway attachments to create CloudWatch metric alarms
* CloudWatch alarms will report when `PacketsIn` on an attachment is below `1` for five minutes.
* SNS topic will receive these alarms and pass through PagerDuty integration to `high priority alarms`
* Solves #1988


davidkelliott (changes requested):
Actually sorry one sca warning to address please!

davidkelliott (approved):

PR #2007, opened 2022-07-14
Create Out of Hours Support

squadMoj:
Adding a page on our OOH support model


source/user-guide/Out of Hours Support
```diff
@@ -0,0 +1,70 @@
+---
+owner_slack: "#modernisation-platform"
+title: Our offer to you
```
squadMoj, line 3: Need to change title to: Out of hours support

PR #2013, opened 2022-07-15
Switch monitored TGW attachments to use curated list

dms1981:
* Use curated list to only include high priority attachments seeing active use


davidkelliott (approved):
Great work 👍

dms1981:
> Great work 👍

Certainly a lot more work than I'd intended at the outset!

Issue #1397, opened 2022-01-12
Investigate use of route table propagation for Transit Gateway


dms1981:
## Value

At present we make use of a combination of local values and for expressions to put routes into our transit gateway route tables. This has the unfortunate side effect of requiring hard-coded attachment values and hard-coded routes.
Terraform offers a way to use AWS's features to achieve the same end-goal with less code, removing our reliance on hard-coded values and reducing the complexity of our code.

## Questions / Assumptions

What would a successful test look like?
What would we define as an improvement to the quality of our codebase?
What documentation (if any) would need to be updated as a result?

## Definition of done

- [ ] code change has been tested
- [ ] code change has been reviewed by another team member
- [ ] all attached VPCs make use of propagation
- [ ] any duplication from static routes is removed

## Reference

* [How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)
* (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ec2_transit_gateway_route_table_propagation)
* (https://aws.amazon.com/blogs/architecture/field-notes-working-with-route-tables-in-aws-transit-gateway/)

dms1981:
solved by https://github.com/ministryofjustice/modernisation-platform/pull/2045

PR #2043, opened 2022-07-26
Add TGW propagation statements

dms1981:
* Add two statically typed lists with `live_data` and `non_live_data` VPCs
* Add data call to create a list of all TGW attachment IDs
* Add data call to create map of all TGW attachment attributes 
* Adds propagation statements to populate route tables with relevant attachments

### DIFF
--- a/terraform/environments/core-network-services/locals.tf
+++ b/terraform/environments/core-network-services/locals.tf
@@ -29,4 +29,32 @@ locals {
     "hmpps-production-attachment",
     "core-network-services-live_data-attachment",
   ]
+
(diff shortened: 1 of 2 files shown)


dms1981:
This PR covers part of #1397 - a follow-up PR will be raised to remove the terraform code that statically adds routes into the two route tables as these will not be required once the same routes are supplied via propagation.
In the interim, the static routes will be preferred over the routes supplied via propagation

davidkelliott (approved):
Looks like a sensible safe approach 👍

PR #2051, opened 2022-07-27
remove creation of static TGW routes from vpc-tgw-routing module

dms1981:
This removes the creation of static routes in the relevant `live_data` / `non_live_data` route table as well as the `firewall` route table. These routes are now supplied through dynamic propagation, so the static routes are no longer required.


davidkelliott (approved):
Looks good and tested in Dev 👍

PR #2071, opened 2022-08-03
Fix/static code analysis

dms1981:
* address some deprecated interpolations
* address tfsec alerts; generally set to ignore as we were getting flags against intentional uses


zuriguardiola (approved):

ep-93 (approved):
LGTM

PR #2091, opened 2022-08-08
OIDC provider permissions

dms1981:
* Added ability for github-actions role to assume necessary roles
* Partially completes #2306

### DIFF
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -188,6 +188,7 @@ data "aws_iam_policy_document" "assume_role_policy" {
     principals {
       type = "AWS"
       identifiers = ["arn:aws:iam::${local.modernisation_platform_account.id}:root",
+        "arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:role/github-actions",
         "arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:user/testing-ci"
       ]
     }
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -32,13 +32,14 @@ module "cicd-member-user" {
 
 module "member-access" {
   count  = local.account_data.account-type == "member" && terraform.workspace != "testing-test" ? 1 : 0
-  source = "github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access?ref=v2.0.0"
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access?ref=v2.1.0"
   providers = {
     aws = aws.workspace
   }
-  account_id = local.modernisation_platform_account.id
-  policy_arn = aws_iam_policy.member-access[0].id
-  role_name  = "MemberInfrastructureAccess"
+  account_id             = local.modernisation_platform_account.id
+  additional_trust_roles = "arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:role/github-actions"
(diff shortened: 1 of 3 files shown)


terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -188,6 +188,7 @@ data "aws_iam_policy_document" "assume_role_policy" {
     principals {
       type = "AWS"
       identifiers = ["arn:aws:iam::${local.modernisation_platform_account.id}:root",
+        "arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:role/github-actions",
```
davidkelliott, line 191: So this is the right code but in the wrong place...this is going to be a little more complex than we initially thought.

This code is for the MemberInfrastructureAccess role created only in the testing-test account.

We actually need to add this line in the role created in all the other member accounts which is here - 

https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/iam.tf#L33

But that is created by an external module which assigns the assume role here:
https://github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access/blob/main/main.tf#L2

And that is used in multiple places...so we would need to either pass in the policy or do some sort of conditional logic.  I seem to remember this is why I didn't use this module for the role in the testing-test account because there was no way I could allow the testing-ci user to assume it.

terraform/environments/core-network-services/iam.tf
```diff
@@ -16,6 +16,17 @@ resource "aws_iam_role" "dns" {
...
           "Condition" : {}
+          }, {
+          "Effect" : "Allow",
+          "Principal" : {
+            "AWS" : "arn:aws:iam::*:role/github-actions"
+          },
+          "Action" : "sts:AssumeRole",
+          "Condition" : {
+            "ForAnyValue:StringLike" : {
+              "aws:PrincipalOrgPaths" : ["${data.aws_organizations_organization.root_account.id}/*/${local.environment_management.modernisation_platform_organisation_unit_id}/*"]
```
davidkelliott, line 27: Looks good, I think we can lock this down further to the member ou, but we don't have that stored as a secret, so I think it's best to go with this for now and then lock it down further as part of that other issue.

terraform/environments/core-vpc/vpc.tf
```diff
@@ -263,6 +263,7 @@ resource "aws_iam_role" "member-delegation" {
           "Principal" : {
             "AWS" : concat(
               local.expanded_account_numbers_with_keys[each.key],
+              format("%s:role/github-actions", local.expanded_account_numbers_with_keys[each.key]),
```
davidkelliott, line 266: Actually thinking about this, if we've already allowed any of the accounts to assume the role with the statement on line 265, I don't think we need this do we?

davidkelliott (changes requested):

terraform/environments/core-vpc/vpc.tf
dms1981: It's probably superfluous. 265 will expand out to something like `arn:aws:iam::111111111111:root` which ought to allow any user or role to assume the `member-delegation-$business_unit-$env` role.

terraform/environments/bootstrap/delegate-access/iam.tf
dms1981: OK! I've address this through this PR: https://github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access/pull/7
This allows me to supply `additional_trust_roles` through which I can pass the relevant `github-actions` role.

davidkelliott (approved):
Plan for delegate access run locally against sprinker, looks good. 👍

PR #2114, opened 2022-08-10
Initial commit for the new env for data-insight

ep-93:


dms1981 (approved):
Approved - we know the OPA test will fail, but this will let us see what happens

PR #2149, opened 2022-08-16
OIDC for all member accounts

dms1981:
* Adds github-oidc module to `delegate-access` step, ensuring that required resources are provisioned when accounts are created.


davidkelliott (approved):
Looks good 👍

PR #2150, opened 2022-08-16
Remove S3 access bucket logging for state bucket

davidkelliott:
The access logs bucket currently doesn't work, this is due to AES256 not
being available in the module for the target bucket (fixed) and a deny
policy preventing any aws:SecureTransport false traffic.  We don't want
to remove this policy, so access log buckets will need to be created
manually or we add the full code to the S3 bucket module with a flag, eg
enable_access_logs which will create everthing.

Having compared the differences in the logs verses what we get with
cloudtrail here - https://docs.aws.amazon.com/AmazonS3/latest/userguide/logging-with-S3.html
We don't need the additional logs for the state bucket.  We will
re-address this if a user needs it.

### DIFF
--- a/terraform/modernisation-platform-account/s3.tf
+++ b/terraform/modernisation-platform-account/s3.tf
@@ -94,7 +94,7 @@ module "state-bucket-s3-replication-role" {
 }
 
 module "state-bucket" {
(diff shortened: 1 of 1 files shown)


dms1981 (approved):
LGTM

PR #2151, opened 2022-08-16
Features/instance scheduler role

gfou-al:
For the most part, it's a copy of the "member-access" role and the Instance Scheduler Lambda resources under the modernisation-platform-account.

### DIFF
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -180,7 +180,7 @@ resource "aws_iam_policy" "member-access" {
 }
 
 module "instance-scheduler-access" {
-  count  = local.account_data.account-type == "member" && terraform.workspace != "testing-test" ? 1 : 0
+  count  = local.account_data.account-type == "member" 1 : 0
   source = "github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access?ref=v2.2.0"
   providers = {
     aws = aws.workspace
--- a/terraform/environments/bootstrap/delegate-access/iam.tf
+++ b/terraform/environments/bootstrap/delegate-access/iam.tf
@@ -38,7 +38,8 @@ module "member-access" {
     aws = aws.workspace
   }
   account_id             = local.modernisation_platform_account.id
-  additional_trust_roles = [format("arn:aws:iam::%s:role/github-actions", local.environment_management.account_ids[terraform.workspace])]
+  additional_trust_roles = [format("arn:aws:iam::%s:role/github-actions", local.environment_management.account_ids[terraform.workspace]),
+    "arn:aws:lambda:eu-west-2:${local.environment_management.account_ids["modernisation_platform_account_id"]}:function:golang-instance-scheduler"]
   policy_arn             = aws_iam_policy.member-access[0].id
   role_name              = "MemberInfrastructureAccess"
 }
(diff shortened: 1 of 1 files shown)


terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
   policy      = data.aws_iam_policy_document.member-access.json
 }
 
+module "instance-scheduler-access" {
+  count  = local.account_data.account-type == "member" && terraform.workspace != "testing-test" ? 1 : 0
```
davidkelliott, line 183:
```suggestion
  count  = local.account_data.account-type == "member" 1 : 0
```
You can create this role in the testing-test account as well so that the scheduler can also be tested in there if needed

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access?ref=v2.2.0"
+  providers = {
+    aws = aws.workspace
+  }
+  account_id             = local.modernisation_platform_account.id
+  additional_trust_roles = [format("arn:aws:iam::%s:role/github-actions", local.environment_management.account_ids[terraform.workspace])]
```
davidkelliott, line 189: I think the additional_trust_roles would need to give permissions to the lambda running the scheduler in the shared services account, is that right @dms1981 ?

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+  policy_arn             = aws_iam_policy.instance-scheduler-access[0].id
+  role_name              = "InstanceSchedulerAccess"
+}
+
+#tfsec:ignore:aws-iam-no-policy-wildcards
+data "aws_iam_policy_document" "instance-scheduler-access" {
```
davidkelliott, line 195: This seems like quite a lot of permissions, would it not only need describe and start stop EC2 permissions?

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+      "ec2:SendDiagnosticInterrupt",
+      "ec2:*LaunchTemplate*",
+      "ec2:*PlacementGroup*",
+      "ec2:*IdFormat*",
+      "ec2:*Spot*"
+    ]
+    resources = ["*"] #tfsec:ignore:AWS099 tfsec:ignore:AWS097
+  }
+
+  statement {
```
davidkelliott, line 228: I don't thing this is needed as the role will just be doing the starting and stopping, it won't be assuming any other roles

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+    actions = [
+      "sts:AssumeRole"
+    ]
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+
+  statement {
```
davidkelliott, line 239: As above I don't think this is needed

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+    actions = [
+      "sts:AssumeRole"
+    ]
+    resources = [
+      "arn:aws:iam::${local.modernisation_platform_account.id}:role/developer",
+      "arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:role/developer"
+    ]
+  }
+
+  statement {
```
davidkelliott, line 250: This role is being called by the lambda so I don't think it needs any permissions to read/write to logs

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,166 @@ resource "aws_iam_policy" "member-access" {
...
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+    resources = [
+      format("arn:aws:logs:eu-west-2:%s:*", local.modernisation_platform_account.id),
+      format("arn:aws:logs:eu-west-2:%s:*", local.environment_management.account_ids[terraform.workspace])
+    ]
+  }
+
+  statement {
```
davidkelliott, line 275: This and the following statement look to be a copy and past of the original role and are not needed here

davidkelliott (changes requested):
A good start, it probably needs some input from @dms1981 as well :)

terraform/environments/bootstrap/delegate-access/iam.tf
dms1981: It also needs permissions to read tags if we want to pick instances for starting / shutting down based on tags. I'd thought the following permissions would be appropriate:
```
ec2:DescribeInstances
ec2:DescribeInstanceStatus
ec2:DescribeTags
ec2:StartInstances
ec2:StopInstances
```

terraform/environments/bootstrap/delegate-access/iam.tf
dms1981: The goal for this ticket should be that the lambda function has a role that allows it to assume a role in a separate account (or, I suppose, in the same account).
That assumable role should be the one that has both the permissions to start / stop instances, and a trust relationship statement allowing the lambda function to assume the role.
So... @davidkelliott , yes. and @gfou-al , also yes. But the plan is, in future, to run the lambda from `core-shared-services`.

terraform/environments/bootstrap/delegate-access/iam.tf
gfou-al: Added "ec2:DescribeInstanceStatus",
            "ec2:DescribeTags"

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -179,6 +179,48 @@ resource "aws_iam_policy" "member-access" {
...
 }
 
+module "instance-scheduler-access" {
+  count  = local.account_data.account-type == "member" 1 : 0
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-cross-account-access?ref=v2.2.0"
+  providers = {
+    aws = aws.workspace
+  }
+  account_id             = local.modernisation_platform_account.id
+  additional_trust_roles = [format("arn:aws:iam::%s:role/github-actions", local.environment_management.account_ids[terraform.workspace]),
```
davidkelliott:
```suggestion
  additional_trust_roles =
```
Github actions role doesn't need permissions to this role

davidkelliott (approved):
Thanks looks good to me!

dms1981 (approved):
LGTM

PR #2173, opened 2022-08-24
Initial add to create referrals, however I may need to break up the p…

ep-93:
…rocess for the pipeline to do the correct stages,  anything under environments-networks might need to be removed and done later


dms1981 (approved):
LGTM!

PR #2182, opened 2022-08-26
Removing the entire env to complete the removal process.

ep-93:


SteveLinden (commented):
Are you sure you want to remove all those .tf files? backend.tf, base_variables.tf, locals.tf, platform_secrets.tf, providers.tf, subnet_share.tf, versions.tf and the json networking.auto.tfvars.json

ep-93:
> 

I'm not 100 percent on anything as this is my first time doing this, however I think if I remove ALL of them, when re adding whats needed, the pipeline should generate a new PR to add the terraform files back in. The ones that are currently there are not actually applied or used.

dms1981 (approved):

PR #2205, opened 2022-08-31
amended how modernisation platform account ID is referenced

dms1981:
* `modernisation_platform_account_id` reference updated to reflect where it sits in the json


ep-93 (dismissed):
LGTM

ep-93 (approved):
LGTM

PR #2211, opened 2022-08-31
Change how we check for names, the command now only outputs matches i…

ep-93:
…f they appear in enviroment-networks also


davidkelliott (dismissed):
Tested locally, working as expected 👍

davidkelliott (approved):
👍

SteveLinden (approved):

PR #2215, opened 2022-08-31
Feature/update docs action

dms1981:
* Remove legacy docs actions
* Create new publish action using modern documents action


davidkelliott (dismissed):
Looks good, could we add in the Slack failure notification to the build and deploy steps so we are notified if there are any issues please?

```
    - uses: 8398a7/action-slack@v3
      name: Slack failure notification
      with:
        status: ${{ job.status }}
        fields: workflow,job,repo,commit,message
      env:
        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
      if: ${{ failure() }}
```

Also could you update the make file so that we can easily run this locally?

Happy for these to be in separate PR(s) though if you want to focus on getting this working first.  Thanks

davidkelliott (approved):

PR #2218, opened 2022-09-01
Add new OPA test to check valid Access level

davidkelliott:
A recent error when creating an environment could have been avoided with
better validation of the json file.  This test should prevent the same
thing happening again.


SteveLinden (approved):

dms1981 (approved):
LGTM

PR #2229, opened 2022-09-01
Remove duplicate refer-and-monitor account

davidkelliott:
The account was created in error and replaced by refer-monitor

Also added some clarifying notes to the delete guidance


ep-93 (approved):
LGTM

PR #2233, opened 2022-09-02
remove Christine Elliot from members list 

dms1981:
Having Christine included now that she's left the team meant that terraform apply runs were failing


ep-93 (approved):
LGTM

PR #2258, opened 2022-09-06
Initial commit for log retention docs, needs to be populated.

ep-93:


davidkelliott:
Were all of our logs compliant? I don't see any changes to the logs themselves

davidkelliott (commented):
I'm wondering actually do we need this as it is repeating the security guidance, should we just signpost to that so that we don't risk becoming out of date if they change it?

PR #2265, opened 2022-09-07
Create Athena user guide

SteveLinden:
Create initial user guide for Athena


ep-93 (approved):
LGTM, lets see what they look like pushed!

PR #2269, opened 2022-09-07
Create assumable role githubreadonly with managed policies

gfou-al:
This is for the github-actions role within the member SSO accounts to be able to assume this githubreadonly role to make the initial OIDC request through the AWS provider.

This will enable us to get things from the main mp account when using
github actions role from the local account or when using the superadmin
role when running locally

### DIFF
--- a/terraform/modernisation-platform-account/iam.tf
+++ b/terraform/modernisation-platform-account/iam.tf
@@ -396,7 +396,7 @@ data "aws_iam_policy_document" "github_actions_assume_role_policy" {
     condition {
       test     = "ForAnyValue:StringLike"
       variable = "aws:PrincipalArn"
-      values   = [
+      values = [
         "arn:aws:iam::*:role/github-actions",
         "arn:aws:iam::${data.aws_caller_identity.current.id}:role/superadmin"
       ]
--- a/terraform/modernisation-platform-account/iam.tf
+++ b/terraform/modernisation-platform-account/iam.tf
@@ -373,3 +373,48 @@ resource "aws_iam_role_policy_attachment" "instance-scheduler" {
   policy_arn = aws_iam_policy.instance-scheduler.arn
   role       = aws_iam_role.instance-scheduler.name
 }
+
+# GitHub Actions readonly role
+data "aws_iam_policy_document" "github_actions_assume_role_policy" {
+  version = "2012-10-17"
+
+  statement {
+    effect  = "Allow"
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    condition {
+      test     = "ForAnyValue:StringLike"
+      variable = "aws:PrincipalOrgPaths"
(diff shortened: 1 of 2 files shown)


terraform/modernisation-platform-account/iam.tf
```diff
@@ -372,4 +372,23 @@ resource "aws_iam_policy" "instance-scheduler" {
...
\ No newline at end of file
+}
+
+# Create assumable roles with managed policies
+# This is for the github-actions role within the member SSO accounts to be able to assume this githubreadonly
+# role to make the initial OIDC request through the AWS provider
+module "iam_assumable_roles" {
+  source               = "terraform-aws-modules/iam/aws//modules/iam-assumable-roles"
+  version              = "~> 4.11"
+  max_session_duration = 43200
```
davidkelliott, line 383:
```suggestion
  max_session_duration = 3600
```
Since this role will only be used for pipelines shall we reduce the session duration to 1hour max?

terraform/modernisation-platform-account/iam.tf
```diff
@@ -372,4 +372,23 @@ resource "aws_iam_policy" "instance-scheduler" {
...
+
+  # Read-only role
+  create_readonly_role       = true
+  readonly_role_name         = "githubreadonly"
+  readonly_role_requires_mfa = true
+
+  # Allow created users to assume these roles
+  trusted_role_arns = [
```
davidkelliott, line 391: This will need to be for all member accounts, so the trust policy might have to have `*` as the principal and a condition similar to:
```
    condition {
      test     = "ForAnyValue:StringLike"
      variable = "aws:PrincipalOrgPaths"
      values   = ["${data.aws_organizations_organization.root_account.id}/*/${local.modernisation_platform_ou_id}/*"]
    }
```

terraform/modernisation-platform-account/iam.tf
```diff
@@ -372,4 +372,23 @@ resource "aws_iam_policy" "instance-scheduler" {
...
+# role to make the initial OIDC request through the AWS provider
+module "iam_assumable_roles" {
+  source               = "terraform-aws-modules/iam/aws//modules/iam-assumable-roles"
+  version              = "~> 4.11"
+  max_session_duration = 43200
+
+  # Read-only role
+  create_readonly_role       = true
+  readonly_role_name         = "githubreadonly"
+  readonly_role_requires_mfa = true
```
davidkelliott, line 388: I think this should be false as the Github role won't have MFA.

terraform/modernisation-platform-account/iam.tf
davidkelliott: I'm not sure if the module supports passing in of policies or conditions? If not it might have to be created using individual resources rather than the module.

davidkelliott (changes requested):

ep-93 (approved):
LGTM.

davidkelliott (approved):

PR #2283, opened 2022-09-10
Feature/vpc creation prod preprod

ep-93:


dms1981 (commented):
Can we set the bastion flag as false for the new VPCs? Without having any services using them, we don't have a need for bastions in them.

davidkelliott (approved):
Checked and verified, all good 👍

PR #2286, opened 2022-09-13
VPC Creation documentation

ep-93:
Page made, will need to link to it from other page, but can be revised in its current state.


SteveLinden (approved):
Not sure if you should put the real IP addresses in but other than that it looks OK

davidkelliott (approved):
Looks good, one minor one is I would add a link to the index page for it under runbooks as well - https://user-guide.modernisation-platform.service.justice.gov.uk/#runbooks

PR #2293, opened 2022-09-13
Add permission for OIDC role to decrypt KMS

davidkelliott:
This is required for decrypting the state bucket. The state bucket KMS key has not be specified as a resource as you cannot use the alias in the policy, and we don't want to hard code the key ARN.

Adding the githubReadOnly role and fixing the member delegation role.


ewastempel (approved):
LGTM

PR #2302, opened 2022-09-14
Add environment management info to Github

davidkelliott:
In order to not hard code account numbers, Github needs access to the map of account names to account numbers in the environemnt management json.

### DIFF
--- a/.github/workflows/new-environment.yml
+++ b/.github/workflows/new-environment.yml
@@ -33,6 +33,8 @@ jobs:
     env:
       AWS_ACCESS_KEY_ID:  ${{ secrets.PRIVILEGED_AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY:  ${{ secrets.PRIVILEGED_AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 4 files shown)


.github/workflows/new-environment.yml
```diff
@@ -33,6 +33,8 @@ jobs:
     env:
       AWS_ACCESS_KEY_ID:  ${{ secrets.PRIVILEGED_AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY:  ${{ secrets.PRIVILEGED_AWS_SECRET_ACCESS_KEY }}
+      TF_VAR_github_token: ${{ secrets.TERRAFORM_GITHUB_TOKEN }}
```
gfou-al, line 36: Is this being used anywhere? I can't seem to find the usage. Same for the below.

gfou-al (approved):
Other than the comment that I can't find the usage of `TF_VAR_github_token`, this PR looks good.

.github/workflows/new-environment.yml
davidkelliott: It's a built in Terraform feature, any environmental variable which starts TV_VAR_ will be picked up by Terraform as a variable - https://www.terraform.io/cli/config/environment-variables#tf_var_name

PR #2313, opened 2022-09-15
Horrible hackiness to get sprinkler and cooker to work

davidkelliott:
Sprinkler and cooker were created as development accounts, so have development as the environment and platforms as the business unit.

They have since been used to test the vpc sandbox so actually are sandbox environment and garden and house business units respectively.

All this means is that logic to restrict what role the github-actions role can assume gives the incorrect business unit and environment.

The options to get this to work are:

1. Rebuild sprinkler and cooker from scratch with new accounts with the sandbox environment, however this will also need an exclusion from the opa tests otherwise we'll have to have sandbox as a valid environment which we don't want.

2. Don't have sprinkler and cooker run using the same workflows as everything else, which would then be hard to make sure the things we are building are compatible not to mention maintaining different workflows.

3. Have 2 lines of awful hackiness.

1 is probably the nicest option, but also the longest and most amount of
  work.  So I'm thinking 3 and create a card in the backlog.


dms1981 (approved):

davidkelliott:
Issue for permanent fix here - https://github.com/ministryofjustice/modernisation-platform/issues/2314

PR #2327, opened 2022-09-21
Give Github actions role put permissions

davidkelliott:
Allow the github actions role permission to modify the state bucket objects so that terraform applies can work when updating state.


SteveLinden (changes requested):
There's a difference in the http options. One is ~>3 and one is 2 (delegate-access is 3.0, modernisation-platform-account is 2.0)
http = {
      version = "~> 3.0"
      source  = "hashicorp/http"
    }

SteveLinden (approved):

PR #2329, opened 2022-09-22
Removing interventions-reporting-discovery as not a MOJ team.

julialawrence:
`nacl` in the network definition json is not required any longer since the refactoring done last sprint.


davidkelliott (approved):
👍 good spot with the NACLs

Issue #2030, opened 2022-07-19
Test and fix if broken local Terraform plan After OIDC Change


zuriguardiola:
## User Story

Before we can roll out the OIDC changes to all accounts, we need to make sure we haven't broken the way that local plans are run.

## Definition of done

- [ ] readme has been updated
- [x] user docs have been updated
- [x] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
https://github.com/ministryofjustice/modernisation-platform/issues/1975

PR #2346, opened 2022-09-28
Updating local plan docs in line with OIDC changes #2030

julialawrence:
Updating local terraform plan instructions to remove `platform_secrets.tf` as that is now the same for GHA and local runs.


davidkelliott (approved):
Tested running locally on sprinkler (modernisation-platform-sandbox role) and on example (modernisation-platform-developer role), both working as expected 👍

PR #2359, opened 2022-09-30
Trigger scheduled baselines when iam module update

davidkelliott:
If this module is modified then the scheduled baselines will need to run again to deploy the changes.


julialawrence (approved):
👍

PR #2372, opened 2022-10-04
Remove users who have left the team

davidkelliott:
Remove users who have left the team and bump the superadmin version


gfou-al (commented):
good to go

PR #2374, opened 2022-10-05
Remove tgw-return-routes module and conduct refactoring

dms1981:
The `core-vpc-tgw-return-routes` module provided the following:
* static routes in the transit gateway route tables
* static route in the core-network-services public route tables

Since we now utilise route propagation in the transit gateway route tables, these static routes are not necessary.
Further to this, as the static routes in the core-network-services public route tables can be easily summarised, doing so will reduce the complexity of our code.

This change, therefore, removes the `core-vpc-tgw-return-routes` module and supplies summarised routes to the public route tables in the `live_data` and `non_live_data` public route tables for the `core-network-services` VPCs.


terraform/environments/core-network-services/vpc.tf
```diff
@@ -33,4 +33,25 @@ module "vpc_hub" {
...
+}
+
+resource "aws_route" "modernisation_platform_10-20-0-0--16" {
+  for_each               = local.networking
+  destination_cidr_block = "10.20.0.0/16"
+  gateway_id             = aws_ec2_transit_gateway.transit-gateway.id
+  route_table_id         = module.vpc_hub[each.key].public_route_tables.id
+}
+
+resource "aws_route" "modernisation_platform_10-25-0-0--16" {
```
davidkelliott, line 45: What is this range?  I see the core account range 10.20 above, and 10.26 is dev and test but I'm not familiar with this one should it be 10.27?  Is it worth adding a comment so we know what each of the ranges are?

dms1981: You're totally right! I've pushed a fix for this. Good catch!

davidkelliott (approved):
Excellent gardening work! 👏

PR #2377, opened 2022-10-06
removed vpc-tgw-routes module

dms1981:
As routes are now written to the core-network-services VPCs in summarised form, it is not necessary for each VPC in the `core-vpc-$environment` to add their own route through a map of subnet sets.


julialawrence (approved):
🔥

PR #2380, opened 2022-10-06
Replace security hub findings script with Go one

davidkelliott:
The existing script does not work and is not fit for purpose. This replaces that script with a new Go script that retrieves the findings across all MP accounts.  Note that this only gets findings from the security hub product which are CRITICAL or HIGH.

Added in other products such as Config and Inspector findings.

Example output can be found here - https://docs.google.com/spreadsheets/d/1b4pJXD1DkOixhQpR8UraFF_yt1ryl4HEZBApt-4YVeE/edit#gid=835871853


scripts/internal/get-security-hub-findings/main.go
```diff
@@ -0,0 +1,164 @@
...
+	// Get accounts secret
+	environments := getSecretsManagerSecret(cfg, "environment_management")
+
+	var allAccounts map[string]interface{}
+	json.Unmarshal([]byte(environments), &allAccounts)
+
+	for _, record := range allAccounts {
+		if rec, ok := record.(map[string]interface{}); ok {
+			for key, val := range rec {
+				// if strings.Contains(key, "sprinkler-development") {
```
gfou-al, line 42: I would remove the commented-out code

davidkelliott: I would to, but I think if I do I'll forget that I can filter it to one account easily here 😂

julialawrence (approved):
👍

PR #2381, opened 2022-10-07
Apply default routes via TGW in VPC module

dms1981:
* Update reference to `modernisation-platform-terraform-member-vpc`
* Remove reference in `environments/core-vpc` to local TGW routing module as this is now handled by the `modernisation-platform-terraform-member-vpc`
* Remove redundant code elements

This PR will require approximately two runs to successfully complete, allowing a five minute interval for AWS API actions to take place, based on observations in sandbox environment


julialawrence (approved):

davidkelliott (approved):
Looks good

PR #2384, opened 2022-10-10
Add OIDC provider to core-security-production account

julialawrence:
- Adding OIDC provider to `core-security-production`
- Adding a new trust policy for the `github-actions` role for core accounts and member accounts so that the role in member accounts can't assume permission roles from within mod-platform-environments repo.


davidkelliott (approved):
Looks good 👍

PR #2387, opened 2022-10-11
Consolidate code for routes back to Transit Gateway

dms1981:
* Remove `core-vpc-tgw-routes` module
* Removed `aws_route` resources from `core-network-services` so they can be migrated into the `vpc-hub` module
* Move functionality to route via Transit Gateway by default to `vpc-hub` module where `gateway = "transit"`
* Include routes back to Modernisation Platform via Transit Gateway where `gateway = "nat"`
* Resolve thrashing on `source_code_hash` for Athena table update


terraform/modules/vpc-hub/main.tf
```diff
@@ -288,6 +288,27 @@ resource "aws_route" "public-internet-gateway" {
   route_table_id         = aws_route_table.public.id
 }
 
+resource "aws_route" "public_mp_10-20-0-0--16" {
```
davidkelliott, line 291: Could we change the names of these to reflect what they are please? eg:
public_mp_core
public_mp_dev_test
public_mp_preprod_prod

terraform/modules/vpc-hub/main.tf
```diff
@@ -288,6 +288,27 @@ resource "aws_route" "public-internet-gateway" {
...
+  count                  = var.gateway == "nat" ? 1 : 0
+  destination_cidr_block = "10.20.0.0/16"
```
davidkelliott, line 293: Could we move the CIDR ranges to a local (probably outside the module) so that we can change them centrally if needed?  Eg have locals like:
```
core_cidr = 10.20.0.0/16
dev_test_cidr = 10.26.0.0/16
preprod_prod_cidr = 10.27.0.0/16
```
This way it's clear what range each resource is referring to and also if we have one central point for this range it's easy to update if we need to change them or rebuild in another region.

davidkelliott: out of scope for this

davidkelliott (approved):

PR #2390, opened 2022-10-11
Instance Scheduler resources

gfou-al:
Add IAM role and lambda function deployment for the Instance Scheduler Lambda Function. Use the latest tag.


davidkelliott (dismissed):
Looks good (noting to change role to more restrictive one after testing deployment)

terraform/environments/core-shared-services/instance-scheduler-lambda-function.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_lambda_function" "instance-scheduler-lambda-function" {
+  #checkov:skip=CKV_AWS_116
+  #checkov:skip=CKV_AWS_117
+  #checkov:skip=CKV_AWS_272 "Code signing not required"
+  function_name                  = "instance-scheduler-lambda-function"
+  handler                        = "main"
+  reserved_concurrent_executions = 0
+  runtime                        = "go1.x"
```
ewastempel, line 8: can we lock the version more? what do we use in other scripts?

terraform/environments/core-shared-services/instance-scheduler-lambda-function.tf
```diff
@@ -0,0 +1,12 @@
...
+  image_uri                      = "${module.instance_scheduler_ecr_repo.ecr_repository_name}:latest"
```
ewastempel, line 9: latest tag should be default, but ideally we want to pass this as a variable, don't we?

davidkelliott (dismissed):

ewastempel (changes requested):
the static code analysis are failing now for: https://github.com/ministryofjustice/modernisation-platform/actions/runs/3243700870/jobs/5318796513

terraform/environments/core-shared-services/instance-scheduler-lambda-function.tf
gfou-al: I don't think we can. According to https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-Runtime at the moment `go1.x` seems to be the only valid value for the runtime configuration for the Go language.

terraform/environments/core-shared-services/instance-scheduler-lambda-function.tf
gfou-al: We use terraform variables to pass different values to terraform modules. In this case, the lambda is not part of a module and only deployed as a single instance in core-shared-services. I don't see why we would be using a variable here, but let me know if you could point me to an example in the existing code.

gfou-al:
> the static code analysis are failing now for: https://github.com/ministryofjustice/modernisation-platform/actions/runs/3243700870/jobs/5318796513

Done

terraform/environments/core-shared-services/instance-scheduler-lambda-function.tf
ewastempel: cool ,thanks for the explanation

ewastempel: ok, let's tackle it in the future

ewastempel (approved):
Cool

PR #2391, opened 2022-10-11
feature/add oidc provider core shared services

julialawrence:
- Adding ECR repo and OIDC provider to core shared services.
- Adding ECR repo and OIDC provider to core shared services.


davidkelliott (approved):
👍 looks good, my only thought is because this is a random one off OIDC for the shared services account would it better sit in with the other core-shared services code, as if we had any more OIDC providers in this code the logic would start to get complex, happy to merge as is now and refactor later though once we've played around with the deployment of the lambda if it makes sense

PR #2394, opened 2022-10-12
Enable appliance mode for Inspection VPC attachment

dms1981:
* Enable `appliance_mode` support for the Inspection VPC attachment
* More information on the implications of this change can be found [here](https://docs.aws.amazon.com/vpc/latest/tgw/transit-gateway-appliance-scenario.html#:~:text=vpc%2Dattachment%20command.-,Stateful%20appliances%20and%20appliance%20mode,-When%20appliance%20mode).

In brief, `appliance_mode` ensures that traffic flows stay within an AWS availability zone ensuring that the statefulness of flows is observed by the same endpoint.


julialawrence (approved):
LGTM

PR #2417, opened 2022-10-14
Add patching ADR

davidkelliott:


architecture-decision-record/0022-patching-strategy.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Status
+
+✅ Accepted
+
+## Context
+
+The Modernisation Platform will end up hosting much of the MoJs legacy infrastructure, like any software, this will need patching and updates to ensure the software is secure. We need a strategy to ensure we stay patched and up-to-date. We held a team session to discover what we are currently patching and discuss what we could be patching. The Miro board output of the session can be found [here](https://miro.com/app/board/uXjVPPdFVPU=/). The findings are summarised below.
+
+## Patching and updating we currently do as a platform
```
ewastempel, line 13: Would it be worth adding content of the patching we are doing, meaning a brief list of what gets updated? It might be just me, but I feel like I am guessing if we update a version of software here or a version of our code and I think it is different per item.

ewastempel (changes requested):
I only had one comment, otherwise it looks great :)

architecture-decision-record/0022-patching-strategy.md
davidkelliott: Thanks for the review! Could you give an example please I'm not sure I understand correctly :)

ewastempel (approved):
LGTM

PR #2420, opened 2022-10-14
Fixing OIDC-related issues with new accounts

julialawrence:
The environments setup in the modernisation-platform repo and modernisation-platform-environments repo shared platform_secrets.tf and locals.tf template definitions.

These two files have been changed to implement the OIDC changes, which added reference to OIDC resources to the modernisation-platform/terraform/environments/<team-name> directories and
resulted in failures during ram share job.

This PR

- Consoledates all OIDC-related changes to locals.tf and providers.tf
- Creates a member-locals.tf to be used in the environments repo
- Amends the file generation scripts to place the correct versions of locals.tf files in the right repos
- Amends the existing maatdb files with the fixes to allow ram share to run

### DIFF
--- a/scripts/provision-environment-directories.sh
+++ b/scripts/provision-environment-directories.sh
@@ -85,7 +85,7 @@ copy_templates() {
   for file in $templates; do
     filename=$(basename "$file")
 
-    if [ ${filename} != "member-providers.tf" ] && [ ${filename} != "data.tf" ]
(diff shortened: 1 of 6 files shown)


dms1981 (approved):

davidkelliott (approved):
Looks like a good approach and consistent with how we've handled other differences. 👍

PR #2424, opened 2022-10-17
Using SSO credentials instead of superadmin

julialawrence:
Giving AdministratorAccess SSO role enough permissions to run `terraform apply` in modernisation-platform-environments

- Permission to assume MemberInfrastructureAccess
- Permission to assume ModernisationPlatformAccess
- Permissions to write to the state s3 bucket
- Permissions to assume modify-dns-records roles in core-vpc accounts


dms1981 (approved):
LGTM!

PR #2425, opened 2022-10-17
Create a resource share for the environment management secret

gfou-al:
This is needed so that the instance scheduler lambda function that runs within the core-shared-services-production account can access the secret.

### DIFF
--- a/terraform/environments/core-shared-services/iam.tf
+++ b/terraform/environments/core-shared-services/iam.tf
@@ -161,7 +161,7 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
       "secretsmanager:ListSecretVersionIds"
     ]
     resources = [
-      "*"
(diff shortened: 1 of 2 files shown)


terraform/environments/core-shared-services/iam.tf
```diff
@@ -138,6 +138,32 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
+    sid    = "AllowAccessEnvironmentManagementSecret"
+    effect = "Allow"
+    actions = [
+      "secretsmanager:GetResourcePolicy",
+      "secretsmanager:GetSecretValue",
+      "secretsmanager:DescribeSecret",
+      "secretsmanager:ListSecretVersionIds"
+    ]
+    resources = [
+      "*"
```
julialawrence, line 164: Sorry George, I just looked up how we can restrict this. Could you please change this * to what I posted in slack?

julialawrence (changes requested):
Details in slack

julialawrence (approved):
LGTM

PR #2426, opened 2022-10-17
Removing handler and runtime from lambda function

julialawrence:
Those are provided within the Docker image and throw an error if specified in the resource where package type is Image


gfou-al (approved):
looks good

davidkelliott (approved):

PR #2438, opened 2022-10-18
Allow ECR image retrieval for the lambda role

gfou-al:


terraform/environments/core-shared-services/iam.tf
```diff
@@ -162,7 +162,28 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
     resources = [
       format("arn:aws:secretsmanager:eu-west-2:%s:secret:environment_management*", local.environment_management.modernisation_platform_account_id)
-   ]
+    ]
+  }
+  statement {
+    sid    = "AllowECRImageRetrieval"
+    effect = "Allow"
+    actions = [
+      "ecr:SetRepositoryPolicy",
```
julialawrence, line 171: Is this permission needed for pulling images?

terraform/environments/core-shared-services/iam.tf
```diff
@@ -162,7 +162,28 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "ecr:CompleteLayerUpload",
```
julialawrence, line 174: Same here. Does lambda need upload permissions?

terraform/environments/core-shared-services/iam.tf
```diff
@@ -162,7 +162,28 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
+      "ecr:DescribeImages",
+      "ecr:DescribeRepositories",
+      "ecr:UploadLayerPart",
```
julialawrence, line 177: Same here, upload permissions. If these are needed that's fine, but are theu#

terraform/environments/core-shared-services/iam.tf
```diff
@@ -162,7 +162,28 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
+      "ecr:ListImages",
+      "ecr:InitiateLayerUpload",
```
julialawrence, line 179: Again, uploading. Is lambda likely to do this
?

terraform/environments/core-shared-services/iam.tf
```diff
@@ -162,7 +162,28 @@ data "aws_iam_policy_document" "instance-scheduler-lambda-function-policy" {
...
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetRepositoryPolicy",
+      "ecr:PutImage"
```
julialawrence, line 182: We are looking to get image not put it I don't think.

julialawrence (commented):
Just some niggles with the perms.

gfou-al:
Closing this PR because the additional permissions were not required. The original error mislead us to believe we had to add permissions, but after locally re-applying through terraform the lambda a few times, the error was gone. It could have been related to the fact that the lambda deployment caches out the image instead of pulling the latest image.

PR #2441, opened 2022-10-19
Allow lambda updates in core-shared-services

julialawrence:
This allows the core-shared-services github-actions role to update lambda code when a new version of the image is released.


gfou-al (approved):
Looks good

PR #2444, opened 2022-10-19
Fix core-logging workflow apply step

dms1981:
Fix the JQ for the apply step so that the account id is properly exported


julialawrence (approved):
Yep. Good. :)

PR #2445, opened 2022-10-19
Features/add lambda retrieval policy

gfou-al:


terraform/environments/core-shared-services/ecr_repos.tf
```diff
@@ -50,6 +50,8 @@ module "instance_scheduler_ecr_repo" {
     local.environment_management.account_ids["core-shared-services-production"]
   ]
 
+  enable_retrieval_policy_for_lambdas = ["arn:aws:lambda:eu-west-2:${local.environment_management.account_ids["core-shared-services-production"]}:function:*"]
```
julialawrence, line 53: I'd just add the arn of the specific function here rather than enabling it for all lambdas in the account.

julialawrence (approved):
Be good on the next go-around to pass the arns of the functions as a list but that is a nice-to-have.

PR #2458, opened 2022-10-21
Allow the lambda role to assume the restricted role in member accounts

gfou-al:


julialawrence (approved):
As the lambda runs in core-shared-services-production account, giving MP account permission to assume it isn't needed.

PR #2484, opened 2022-10-24
Skip the accounts from instance scheduling

gfou-al:
Skip the accounts ccms-ebs-development maatdb-development from instance scheduling until the feature is tested


ewastempel (approved):
LGTM

PR #2488, opened 2022-10-24
Firewall policy module

dms1981:
* Adds a module to create a firewall rule group and policy
* Adds module to `core-network-services`
* Creates a source json to supply the module with an open  stateful firewall rule allowing all traffic through


dms1981:
One thing I'm not sure about is how to supply firewall rule SIDs. I believe they're a [requirement](https://www.digitalocean.com/community/tutorials/understanding-suricata-signatures#:~:text=The%20sid%20and%20rev%20Keywords) - but I'm not sure how to decently structure them. I might have to do so outside of the module and pass the values in, but for the time being I'm supplying a random integer.

davidkelliott (approved):
👍 Nice!

PR #2502, opened 2022-10-25
Adding perms to GH actions so that it can run lambda in go tests

ewastempel:
Currently, the GH action role/policy in the modernisation-platform-instance-scheduler repo does not have permissions to run the lambda function (which gets called called with the GO tests). This PR is to add these permissions.

It may be removed in the future, if the tests change to not call the lambda.


ep-93 (approved):

gfou-al (changes requested):
There are failures:

```
Running tflint in terraform/environments/bootstrap/delegate-access
Excluding the following checks: terraform_unused_declarations
Failed to check ruleset; Failed to check `aws_iam_policy_sid_invalid_characters` rule: /github/workspace/terraform/environments/bootstrap/delegate-access/iam.tf:2,18-22: Error in function call; Call to function "try" failed: no expression succeeded:
- Error in function call (at /github/workspace/terraform/environments/bootstrap/delegate-access/iam.tf:2,22-28)
  Call to function "regex" failed: pattern did not match any part of the given string.
- Error in function call (at /github/workspace/terraform/environments/bootstrap/delegate-access/iam.tf:2,114-120)
  Call to function "regex" failed: pattern did not match any part of the given string.

At least one expression must produce a successful result., and 4 other diagnostic(s)
tflint_exitcode=1
```

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -484,7 +484,8 @@ data "aws_iam_policy_document" "oidc_assume_role_member" {
...
       local.application_name == "sprinkler" ? format("arn:aws:iam::%s:role/member-delegation-garden-sandbox", local.environment_management.account_ids["core-vpc-sandbox"]) : format("arn:aws:iam::%s:role/modernisation-account-limited-read-member-access", local.environment_management.modernisation_platform_account_id),
-      local.application_name == "cooker" ? format("arn:aws:iam::%s:role/member-delegation-house-sandbox", local.environment_management.account_ids["core-vpc-sandbox"]) : format("arn:aws:iam::%s:role/modernisation-account-limited-read-member-access", local.environment_management.modernisation_platform_account_id)
+      local.application_name == "cooker" ? format("arn:aws:iam::%s:role/member-delegation-house-sandbox", local.environment_management.account_ids["core-vpc-sandbox"]) : format("arn:aws:iam::%s:role/modernisation-account-limited-read-member-access", local.environment_management.modernisation_platform_account_id),
+      format("arn:aws:iam::%s:role/InstanceSchedulerLambdaFunctionPolicy", local.environment_management.account_ids["core-shared-services-production"])
```
davidkelliott, line 488: I don't think this is needed here, the OIDC provider you will be using is here - https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/core-shared-services/iam.tf#L203

davidkelliott: Ah I see you have it below, I think this change can be removed

ewastempel: I wasn't sure, since it's going to iterate through other accounts... ?

ewastempel: I shall remove it then. I can always raise another PR :D

davidkelliott (approved):

gfou-al (approved):

Issue #1859, opened 2022-06-08
Start using SSO credentials instead of superadmin for local work.


davidkelliott:
## User Story

The superadmin credentials (AWS access keys) should not be used everyday, we should switch to SSO credentials and delete all superadmin creds, creating them to use as one off only when absolutely necessary. 

## User Type(s)

MP engineer

## Value

Increase the security of the platform

### Hypothesis 
If we use SSO credentials for local development we won't need to have superadmin credentials anymore

### Proposal
Work out how to use SSO credentials with our current Terraform, then once tested and documented remove superadmin credentials.

## Definition of done

- [ ] superadmin credentials deleted
- [x] readme has been updated
- [x] user docs have been created/updated
- [x] another team member has reviewed
- [x] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

julialawrence:
Relevant PR: https://github.com/ministryofjustice/modernisation-platform/pull/2455

julialawrence:
Related PR: https://github.com/ministryofjustice/aws-root-account/pull/612

- Enable AdministratorAccess role in Modernisation Platform account

julialawrence:
More PRs:

https://github.com/ministryofjustice/modernisation-platform/pull/2509 -- merging providers for delegate-access
https://github.com/ministryofjustice/modernisation-platform/pull/2508 -- doc updates
https://github.com/ministryofjustice/modernisation-platform-ami-builds/pull/113 -- sprinkler and example fixups found while testing.

PR #2509, opened 2022-10-27
Local plans without file changes

julialawrence:
Linked story: https://github.com/ministryofjustice/modernisation-platform/issues/1859

Updating `providers.tf` and `locals.tf` in `environments/bootstrap/delegate-access` to allow local plans without file changes.

- Adding a `whoami` `aws_iam_session_context` datasource to retrieve the arn of the calling role
- Updating AWS provider declarations to assume a role based on the arn of the calling role.


davidkelliott (changes requested):
The SCA is still failing, I think the tflint ignore may need to be on the data policy resources where the local.application_name is called?

davidkelliott (approved):

PR #2511, opened 2022-10-27
Adding GH actions role principal to the lambda function role policy document

ewastempel:
The lambda role needs to be assumable by GH actions role.


gfou-al (approved):
looks awesome :)

julialawrence (approved):
👍

PR #2519, opened 2022-10-28
Add documentation for shared KMS keys

davidkelliott:
Also adds the required data sources to the data.tf, so users don't need to add this themselves.

A further PR to add this data to the existing environments in the environments repo will be required.


ewastempel (approved):
LGTM

PR #2520, opened 2022-10-28
Resolve issue with replacing FW rule group

dms1981:
Without an explicit `create_before_destroy` lifecycle statement, terraform will try to destroy the existing network firewall rule group, and fail as the rule group cannot be destroyed while in use. This should change the behaviour to create a new rule group, replace the original rule group, and then delete the original resource


gfou-al (approved):
looks good

PR #2522, opened 2022-10-28
Fix wrong code source url in tag

davidkelliott:
Also remove some code which isn't needed now that we've moved to a separate member-locals file rather than substituting.


julialawrence (approved):
Lgtm

PR #2543, opened 2022-11-03
Add "kms:CreateGrant" to instance scheduler role to try fixing the issue

gfou-al:
State transition reason: Server.InternalError
State transition message: Client.InternalError: Client error on launch


julialawrence (approved):
Yep looks good!

PR #2544, opened 2022-11-03
Ignore open KMS resource

davidkelliott:
Cannot restrict by KMS alias so leaving open resource and ignoring check


julialawrence (approved):
👍

PR #2552, opened 2022-11-04
S3 lock for core-vpc-sandbox

SteveLinden:
Initial set up for just core-vpc-sandbox for testing.


terraform/environments/core-vpc/iam.tf
```diff
@@ -1,3 +1,48 @@
...
+  path        = "/"
+  description = "S3 and dynamo policy"
+
+  policy = jsonencode({
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Effect" : "Allow",
+        "Action" : "s3:ListBucket",
+        "Resource" : "arn:aws:s3:::mybucket"
```
dms1981, line 16: Probably want to change this to refer to the correct bucket.
Have you considered maybe using the output of a data call to populate this, to keep you from needing a hard coded reference in the policy?

terraform/environments/core-vpc/iam.tf
```diff
@@ -1,3 +1,48 @@
...
+          "dynamodb:GetItem",
+          "dynamodb:PutItem",
+          "dynamodb:Query",
+          "dynamodb:UpdateItem"
+        ],
+        "Resource" : ["arn:aws:dynamodb:eu-west-2:${local.environment_management.modernisation_platform_account_id}:table/modernisation-platform-terraform-state-lock"],
+        "Condition" : {
+          "ForAllValues:StringEquals" : {
+            "dynamodb:LeadingKeys" : [
+              "myorg-terraform-states/myapp/production/tfstate",    // during a state lock the full state file is stored with this key
```
dms1981, line 39: Are these key values correct? I suspect we're not using `myorg-terraform-states/myapp` in the DynamoDB config.

terraform/environments/core-vpc/iam.tf
SteveLinden: Bucket name corrected.

terraform/environments/core-vpc/iam.tf
SteveLinden: May be corrected with "modernisation-platform-terraform-state/global-resources/terraform.tfstate" and "modernisation-platform-terraform-state/global-resources/terraform.tfstate-md5"

SteveLinden: Change made and pushed

julialawrence (approved):
LGTM. State locking is happening. Plans show no changes.

PR #2557, opened 2022-11-07
Patch Manger sprinkler module

ep-93:
Moved from environments account so that I can apply this correctly, with the correct permissions. Should create 4 changes, including a patch group and baseline


terraform/environments/sprinkler/ssm.tf
```diff
@@ -0,0 +1,163 @@
...
+        name   = "InstanceId"
+        values = aws_ssm_maintenance_window_target.ssm-targets.*.id
+      }
+    }
+  }
+}
+
+###### s3 Bucket Patch Logs #####
+
+resource "aws_s3_bucket" "patching-log-bucket" {
```
julialawrence, line 156: As part of further refinement, would be good to use the s3 module for this bucket.

dms1981 (approved):
Discussed with @ep-93 - LGTM!

PR #2559, opened 2022-11-07
Restrict CreateGrant permission in instance scheduler role

julialawrence:
Instance Scheduler needs kms:CreateGrant permissions to manage encrypted ebs volumes. The grantee in this instance is the EC2 service, so it's possible to lock the permission down using `kms:GrantIsForAWSResource` condition.


davidkelliott (approved):
👍

PR #2567, opened 2022-11-07
Refactoring  modernisation-platform-account workflow to use OIDC

ewastempel:
In addition to using OIDC in the modernisation-platform-account workflow, the env variables has been extracted from the job and set as global variables and added some spacing and consistency too.


julialawrence (approved):
👍

gfou-al (approved):
👍

PR #2571, opened 2022-11-07
Update to OIDC for SSO and secure baselines

davidkelliott:
Skipping the jobs which use the elevated root creds.


gfou-al (approved):
👍

julialawrence (approved):
👍

PR #2573, opened 2022-11-07
Update docs with instance scheduler info

davidkelliott:
Updating the offer and sustainability page now that we have the instance scheduler up and running.


gfou-al (approved):
👍

PR #2574, opened 2022-11-08
Turn on instance scheduling for nomis dev and test

gfou-al:
The only remaining environment is nomis-preproduction


julialawrence (approved):
LGTM

PR #2581, opened 2022-11-09
Migrating lambda creation code to module.

julialawrence:
Moving lambda and related iam code to the module:

* lambda function resource
* lambda iam role
* lambda iam role trust policy
* lambda allowed triggers permissions


gfou-al (approved):
👍

PR #2594, opened 2022-11-11
Secure baselines plan+apply taking ~60m, running up against token expiration.

julialawrence:
https://github.com/ministryofjustice/modernisation-platform/actions/runs/3435363726/attempts/1

The above run failed because the token expired before the job could complete. I'm seeing runtimes of between 45 and 55 minutes fairly regularly. Splitting them out is a clean way of fixing without adjusting token TTL which isn't needed for anything else.


ewastempel (approved):
LGTM

PR #2605, opened 2022-11-11
SSO removal now happens in SCIM

davidkelliott:
See the PR here - https://github.com/ministryofjustice/moj-terraform-scim-github/pull/36

So we no longer need to manually remove users.


julialawrence (approved):
LGTM

PR #2608, opened 2022-11-11
Add dynamodb table to backend.tf for locking

SteveLinden:
Added change to core-security backend.tf for locking purposes. Will test it works in here.


julialawrence (approved):
LGTM. Checks are passing. State is locking.

PR #2611, opened 2022-11-11
Decrease alert sensitivity

dms1981:
Our current alarms alert if one datapoint is missed, or polls as `0`, in a 5 minute window. The net effect of this is that in any rolling five minute window an alert can trigger if no traffic is detected. Our goal with these alerts was to warn us if a major even occurred in a VPC (such as a routing change or ACL / security change) stopped traffic flowing through the Transit Gateway. The alert is so sensitive that so far we've only had one situation where the alert has been legitimate.


davidkelliott (approved):
Looks good thanks for this!

PR #2614, opened 2022-11-11
Removing write access for ops eng team to MP module repositories

ewastempel:
Removing write access for ops eng team to MP module repositories, as they have removed themselves manually already.
Additionally, they do not have accounts in our platform anymore.


julialawrence (approved):
Terraform gives, terraform takes away. 👎

dms1981 (approved):

PR #2632, opened 2022-11-14
Pin the slackbot version

davidkelliott:
Following best practice guidance, pinning this to a commit hash.

See guidance here https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions


ewastempel (approved):
❤️

PR #2633, opened 2022-11-14
Make alarm less sensitive

davidkelliott:
The alarm is going off regularly as it is too sensitive.  Changing so that there has to be 15minutes of zero traffic before it triggers.


julialawrence (approved):
LGTM.
If this is firing because of missing polls, would it help to set missing data to ignore?

PR #2636, opened 2022-11-15
Data and Insights PRP and PRD

ep-93:
Initial commit, creating data and insights pre prod and prod environments in this repo, will have a corresponding change in the environments repo too


davidkelliott (changes requested):
Please add to the OPA tests as well :)

PR #2651, opened 2022-11-15
Adding dms:* permissions.

ep-93:
Requested in the ask mod platform channel.


ewastempel (approved):

terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -427,6 +427,7 @@ data "aws_iam_policy_document" "developer-additional" {
       "ec2:CreateSnapshot",
       "ec2:CreateSnapshots",
       "ec2:CreateTags",
+      "dms:*",
```
julialawrence, line 430: Based on the error Hari's having, I think this needs to go into MemberInfrastructureAccess role policy. I believe this is adding it to the collaborator developer role.

julialawrence: This is the policy it should be in: https://github.com/ministryofjustice/modernisation-platform/blob/2e9230ba53ae225bcb0d6b956a4c5f975a7677c6/terraform/environments/bootstrap/delegate-access/iam.tf#L45

julialawrence (changes requested):

davidkelliott (approved):

julialawrence (approved):

PR #2652, opened 2022-11-16
Adding environents management secret to AMI repo to be consumed by the OIDC github workflow.

julialawrence:
To use OIDC provider in authentication, the workflow needs to know the full arn of the role it's assuming post-authentication before terraform can run. Therefore, the environments-management secret, which contains all the member account numbers is loaded as a GHA secret instead.

Additionally, fixing a bug introduced in the previous OIDC PR: passing the ami repo name with the organisation name prefix.

ETA: It looks like one of the certs in the certificate chain for the github oidc host have been updated, as one of the thumbprints has changed. I verified the one starts with 69* is still valid.

### DIFF
--- a/terraform/environments/secrets.tf
+++ b/terraform/environments/secrets.tf
@@ -108,7 +108,7 @@ locals {
 
 # Store environment management secret in Github secrets
 resource "github_actions_secret" "environment_management" {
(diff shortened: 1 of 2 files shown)


ep-93 (approved):
Looks good to me!

PR #2653, opened 2022-11-16
Module to automate the patching of ec2 instances in each account

ep-93:
This is the creation of the repo which will be used as the ssm module, automating the patching of ec2 instances in each account.


julialawrence (dismissed):
LGTM

Issue #2635, opened 2022-11-15
Runbook for CW Networking Alarms


julialawrence:
## User Story

As a modernisation platform engineer
I would like to have a runbook explaining how to investigate, remediate and potentially tune cloud watch networking alarms
So that we have a standard practice we can refer to

## User Type(s)

Modernisation platform engineers
On-call engineers

## Value

The runbook itself and the process of writing it is a good opportunity to increase and share the knowledge of MP networking.

### Hypotheses

If the runbook is available and useful, more people will feel comfortable being on-call and more confident they will be able to handle the alarms without needing escalation.

### Proposal

An experienced engineer pairs with a less experienced engineer to walk through the process of resolving the networking alarms.

The walkthrough is used as a basis for the runbook

## Definition of done

- [ ] Runbook has been created and published on the MP docs site
- [ ] user docs have been updated
- [ ] another team member has reviewed

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/cloudwatch-networking-alarms.html#a-summary-of-current-alarms

PR #2654, opened 2022-11-16
Runbook for cloudwatch networking alarms

dms1981:
This PR adds a runbook covering our networking-related CloudWatch alarms, with some discussion on what can cause the alerts to trigger, and what actions can be taken in the event that an alert has triggered.

### DIFF
--- a/source/runbooks/cloudwatch-networking-alarms.html.md.erb
+++ b/source/runbooks/cloudwatch-networking-alarms.html.md.erb
@@ -33,6 +33,8 @@ traffic from exiting a VPC.
 It is unlikely to cause traffic to be blocked from entering a VPC as the transit gateway endpoint subnets have fixed
 permissive NACLs.
 
+Outputs to **#modernisation-platform-low-priority-alarms**.
+
 ### network-gateway-changes
 
 This alarm triggers when an API call is made to create, delete, or attach an Internet Gateway or Customer Gateway. As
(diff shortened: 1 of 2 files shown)


dms1981:
This PR resolves #2635

source/runbooks/cloudwatch-networking-alarms.html.md.erb
```diff
@@ -0,0 +1,97 @@
...
+
+This alarm triggers when no traffic has been seen traversing a VPC Transit Gateway attachment and exists on a per-VPC
+basis. The intent for this alarm was to alert us when a change or event occurred that caused a cessation of traffic from
+a VPC to the Modernisation Platform Transit Gateway.
+
+Should this alarm trigger you should confirm the VPC in question and assess the behaviour of traffic over the alarm
+period, extending past this alarm to observe what normal behaviour should look like. It is possible for this alarm to
+trigger during ordinary operation when the polling window is short; no traffic over five minutes is not uncommon.
+
+In the event that this alarm triggers, and it has been deemed legitimate, your next step should be to confirm that there
```
julialawrence, line 61: What's the best way to know if it's legitimate?

dms1981: It's something to be judged on a case by case basis, but I'll put an example of a legitimate trigger and a false positive in if that helps?

julialawrence: It would really help thank you!

julialawrence (approved):
Thank you! This will make it easier to know how to handle alarms!

Issue #2549, opened 2022-11-04
Enable Application teams to contribute to modules in our module repositories


ewastempel:
## User Story

As a modernisation platform user I want to be able to write terraform modules to keep the code dry.

## User Type(s)

Application/infrastructures users who host their infrastructure in MP.

## Value

Keeping the code dry and sharing the modules across the teams. Additionally, it is less work for the MP team, if the modules can be implemented through the community efforts.

## Questions / Assumptions / Hypothesis

Users should be able to write to all MP module repositories. 

Timebox implementation to 2 days for for/each loop - manual if it goes beyond this

### Proposal

See the code here, where the write access to repositories is done per repository basis (for all app teams in the local list of app teams): https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/github/main.tf#L351

Refactor the code so that it can iterate through the list of all module repositories and set the write access for all app teams in the list programatically.
e.g. the repo list could be a json file to be read as a local variable of the repo list

## Definition of done

- [ ] readme has been updated
- [ ] contributor guidelines added to docs - indicating generic vs specific
- [ ] linked to github workflow for contributors
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
Solved by https://github.com/ministryofjustice/modernisation-platform/pull/2670

PR #2670, opened 2022-11-17
Allow push access for modernisation platform repositories

dms1981:
As detailed in https://github.com/ministryofjustice/modernisation-platform/issues/2549 this PR does the following:
* Moves the `github_team_repository` code into a child module
* Adds a data call for `github_repositories` and a local value to create a list of repositories
* Uses the module in conjunction with the local value to supply each module with a list of teams to be given `push` access

The goal of this PR is to reduce the amount of manual intervention involved in allowing Modernisation Platform customers the ability to contribute to our code.

### DIFF
--- a/terraform/github/README.md
+++ b/terraform/github/README.md
@@ -6,7 +6,7 @@ This directory creates and maintains the following GitHub items for the Modernis
   - team membership
   - team access to repositories
 
-The state is stored in S3, as defined in [main.tf](main.tf).
+The state is stored in S3, as defined in [backend.tf](backend.tf).
 
 ## How to create a new repository for a terraform module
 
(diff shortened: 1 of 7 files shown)


davidkelliott:
Looks good!  One minor thing, just looking at the plan - modernisation-platform-hello-world is an archived repo, could you filter those out?

Also a few SCA fails (nothing major) to fix please.

dms1981:
> Looks good! One minor thing, just looking at the plan - modernisation-platform-hello-world is an archived repo, could you filter those out?
> 
> Also a few SCA errors to fix please (nothing major)

Can do on both counts, although filtering out the archived repository might be a bit ugly. I'll have a think about the best way to handle it.

davidkelliott (approved):

Issue #2593, opened 2022-11-10
Refactor OIDC Module to accept multiple repos, roles and role policies


julialawrence:
## User Story

As a modernisation platform developer, I would like to refactor the OIDC module to allow one provider to be used across multiple repos with different roles and permission sets.

## User Type(s)

Modernisation Platform Developer 

## Value

Enable rollout of OIDC across all our environments will improve our security posture.

### Proposal

Create roles separately and pass them to the OIDC module?
Pass multiple policies to the module and have it create the roles?

We'll also need to check if we're making assumptions based on the roles names in any policies, etc.

## Definition of done

- [ ] OIDC module modified
- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

julialawrence:
Modified OIDC module to accept multiple repositories: https://github.com/ministryofjustice/modernisation-platform-github-oidc-provider/pull/13

julialawrence:
This story covers changes required for two separate OIDC scenarios that don't fall into the pattern of our core modernisation platform repo and our environments repo.

1. We would like to be able to use the same mechanism we're employing to run terraform in `modernisation-platform` repo to run terraform in our `modernisation-platform-ami-builds` repo. Since the [github-actions](https://github.com/ministryofjustice/modernisation-platform-github-oidc-provider/blob/main/iam.tf#L2) role created with the OIDC provider accepts only one repository for its trust policy, the PR above modifies the module to accept a list of repositories. For details on how that change is then used to convert AMI workflows to OIDC, please see https://github.com/ministryofjustice/modernisation-platform/issues/2563 and https://github.com/ministryofjustice/modernisation-platform-ami-builds/pull/125

[...]

PR #2685, opened 2022-11-18
Adding org account id as GH secret.

julialawrence:
Relevant Issue: https://github.com/ministryofjustice/modernisation-platform/issues/2593

A small prerequisitve to converting org-related workflows to OIDC.

In order to assume a role in the Org account, GHA will need the full arn, including the account id before the terraform can run and fetch it. 
Therefore, need to add the account id as a Github secret.


julialawrence:
Not needed. It's already part of the env management secret.

PR #2695, opened 2022-11-21
Add production account for delius-iaps

dms1981:
* Add a statement to the `delius-iaps.json` to cover the creation of a production account


ewastempel (approved):
LGTM

PR #2696, opened 2022-11-21
added delius-iaps account to hmpps-production environment

dms1981:


davidkelliott (changes requested):
Could you update the expected rego as well please! Thanks

dms1981:
> Could you update the expected rego as well please! Thanks
Is it not already here? https://github.com/ministryofjustice/modernisation-platform/blob/feature/add-delius-iaps-prod-to-accounts/policies/environments/expected.rego#L19

davidkelliott (approved):

PR #2700, opened 2022-11-21
Fix ram share not running on network change

davidkelliott:
The old command was working based on if there was an addition of the networking.auto.tfvars.json, but this file doesn't change after it's initially been created.  Now we are looking for change in the environments-networks/*.json files, these change every time there is a new ram share, either by the file being created in the first instance or by and account being added to it.

The only problem is there is no simple way to know which account has been added, it's easy to do if there is only one account added to existing ones but for the first time file creation it's all new.

So this solution will return all of the accounts and the ram share will run for all of the accounts in a VPC, this shouldn't be a problem however as it should just run and show no changes if they haven't been newly added.


julialawrence (approved):
Lgtm

Issue #2200, opened 2022-08-30
Amend routing for live environments as part of NWFW work


dms1981:
## User Story

As a modernisation platform engineer
I want to route live traffic via the AWS Network Firewall
So that I can use the firewall to secure traffic flows 

## Value

As part of the AWS Network Firewall work we're undertaking, once the NWFW has been prepped for traffic we'll need to make some routing table changes to send traffic destined for the live environments via the NWFW, and traffic from the live environments destined for other MOJ endpoints via the NWFW.

Add static route to external route table:  
* `10.27.0.0/16 via inspection VPC attachment. `

Remove VPC propagated routes for production / preproduction VPCs from external route table:  
* eg `10.27.8.0/21 via hmpps-production-attachment. `
* eg `10.27.104.0/21 via platforms-preproduction-attachment. `

Add static route to firewall route table:  
* `0.0.0.0/0 via MP-PTTP peering attachment. `

Add static routes to live_data route table:  
* `10.0.0.0/8 via inspection VPC attachment. `
* `172.16.0.0/12 via inspection VPC attachment. `
* `192.168.0.0/16 via inspection VPC attachment. `

Remove static route from live_data route table:  
* `10.101.0.0/16 via MP-PTTP peering attachment.` 
* `10.40.0.0/16 via MP-PTTP peering attachment.`
* `10.244.0.0/20 via MP-PTTP peering attachment.`
* `10.47.0.64/26 via MP-PTTP peering attachment.`
* `10.47.0.128/26 via MP-PTTP peering attachment.`
* `10.184.0.0/16 via MP-PTTP peering attachment.`
* `172.20.0.0/16 via MP-PTTP peering attachment.`

Test connectivity between AWS & Azure:

Traffic from AWS to Azure should flow as follows:
`member-vpc > MP TGW > firewall > MP TGW > PTTP TGW > Azure`

Traffic from Azure to AWS should flow as follows:
`Azure > PTTP TGW > MP TGW > firewall > MP TGW > member-vpc`

## Definition of done

- [x] routes changed
- [x] tests successful

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

julialawrence:
PR to amend the routes: https://github.com/ministryofjustice/modernisation-platform/pull/2702

PR #2702, opened 2022-11-22
Amending routing for live environments as part of NWFW work

julialawrence:
Relevant Issue: https://github.com/ministryofjustice/modernisation-platform/issues/2200

This PR does the following:

* Adds a route to `10.27.0.0/16` via the inspection VPC to the `external` route table
* Removes propagations for production and preproduction environments from the `external` route table
* Adds routes to RFC1918 private address ranges via the inspection VPC to the `live_data` route table
* Removes non-RFC1918 routes from the `live_data` route table

The net effect of this PR is that traffic coming over the PTTP TGW peering attachment to a preproduction/production environments will route through the inspection VPC (and therefore through the AWS Network Firewall), and traffic from a preproduction/production environment going to an internal IP address will also route through the inspection VPC.


dms1981 (approved):
LGTM; lines up with both the changes to the lower environments, and the LLD showing the implementation of network firewall in our traffic path.
Probably worth warning our customers in the #modernisation-platform-notifications channel before implementing.

PR #2733, opened 2022-11-30
Adding pre-prod for iaps

SteveLinden:
Added the iaps pre-prod section to the json.


dms1981 (commented):
Don't forget you'll want to add this to `policies/environments/expected.rego` too.

dms1981 (approved):
LGTM - can see it's in the `expected.rego` too

PR #2735, opened 2022-11-30
added missing option to tunnel log options

dms1981:
Without the provision of a `log_output_format` attribute, the AWS API will select `json` by default. However, on seeing this Terraform will attempt to delete this value leading to an unneccesary loop.

This fix PR adds the missing option to prevent this behaviour


davidkelliott (approved):
Thanks for spotting this!

Issue #2655, opened 2022-11-16
Request for mlra-test, mlra-preprod and mlra-prod


mikereiddigital:
### Application Name

MLRA

### Description of application

This is for the remaining environments of the LAA's "MLRA" application that we will be migrating from other AWS accounts to MP.

### GitHub team slug

laa-aws-infrastructure

### Environments

| Environment 		| Access level |
| --- 				| --- |
|<ul><li>[ ] Development</li></ul>	| <ul><li>[ ] view-only</li><li>[ ] developer</li><li>[ ] sandbox</li></ul> |
|<ul><li>[x] Test</li></ul>			| <ul><li>[ ] view-only</li><li>[x] developer</li></ul> |
|<ul><li>[x] Preproduction</li></ul>| <ul><li>[ ] view-only</li><li>[x] developer</li></ul> |
|<ul><li>[x] Production</li></ul>	| <ul><li>[ ] view-only</li><li>[x] developer</li></ul> |

### Tags

tag | value
--- | ---
application | MLRA
business-unit | LAA
owner | LAA

### Subnet sets

- [x] General - this application will use the same subnet as other applications in your business unit
- [ ] Isolated - this application has highly sensitive data which must have its own isolated subnet with no network connectivity

### How do users connect to the application?

- [ ] Over the public internet
- [ ] With a purple cabled device (please give details)
- [ ] With a MoJ Official device
- [x] Via other existing AWS account

### Connectivity to other applications or external parties

These vpcs used by accounts will need to be hooked up to the MoJ's Transit Gateway as per laa-development.

## Additional features

- [ ] Additional VPC Endpoints
- [ ] Extended DNS Zones
- [ ] Other - please specify

## Definition of done

- [ x] account created
- [ x] customer informed

SteveLinden:
Code to create the new environments is created #2734

SteveLinden:
Added the access requirements via #2734 (pull request 2736)

SteveLinden:
Sent a slack message to @mikereiddigital to let him know the accounts have been created. Said to raise anything on the ask channel if there are issues.

PR #2736, opened 2022-11-30
Feature/new mlra systems #2655

SteveLinden:
Including the access required which was not in the previous release.


davidkelliott (changes requested):
In the expected.rego the blocks are already there, you don't need to add them again just the new values to the blocks

PR #2749, opened 2022-12-02
Ram association fix - one line

davidkelliott:
Passing in multiple lines to the github action output doesn't work, it only takes the first line.  This change merges the lines in to one, and then itterates through strings on a single line separated by a space.


ewastempel (approved):
nice!

PR #2751, opened 2022-12-02
Replace comma with space

davidkelliott:
Forgot to do this before the last push, to iterate through it needs a space.


ewastempel (approved):
LGTM

Issue #2400, opened 2022-10-13
Refactor Nuke and Redeploy After Nuke for OIDC


julialawrence:
## User Story

As a platform developer, I would like to refactor our nuke code to make it compatible with OIDC, retire the use of hard-coded credentials, and make the process of adding accounts to the nuke list more automated.

## Value

- Using OIDC instead of static credentials improves the security posture of the platform
- Nuking sandbox environments saves money by destroying unused infrastructure and drives good practices such as less use of "advanced devops" in favour of IaC.

## Questions / Assumptions / Hypothesis

We currently maintain the nuke list in a AWS secret in MP which is manually updated when a new sandbox account is created.
Due to how OIDC works, accessing a secret is difficult.

Refactor the code to use github secrets for account information.

## Definition of done

- [ ] readme has been updated
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

julialawrence:
https://github.com/ministryofjustice/modernisation-platform-terraform-environments/pull/23

Creating three dynamically generated outputs which will then be passed to the autonuke script via GitHub Actions Secrets

- `environment_nuke_accounts`: all sandbox accounts which are not explicitly annotated with `"nuke" : "exclude"` in the environment json file
- `environment_rebuild_after_nuke_accounts`: list of accounts to be rebuilt after autonuke completes. Annotated with `"nuke": "rebuild"` in the environments json files
- `blocklist_nuke_accounts`: all preproduction, production and core accounts.

julialawrence:
https://github.com/ministryofjustice/modernisation-platform/pull/2772

This PR creates GitHub secrets in the modernisation-platform-environments repo which contain the list of accounts to nuke, rebuild after nuke and exclude from nuke.

By default, all sandbox accounts will now be automatically added to the nuke list. This PR uses a newly-introduced optional "nuke" field to explicitly exclude digital-prison-reporting-development from nuke
Our nuke process runs in two parts: the nuke which with some exceptions removes existing resources from an account and rebuild-after-nuke which then runs the corresponding terraform to rebuild the account to the state defined in code. By default, the accounts are NOT rebuilt. Setting "nuke" field to "rebuild" triggers a rebuild after nuke for the account. The json
[...]

PR #2772, opened 2022-12-06
Environment file and environments terraform modifications for awsnuke

julialawrence:
Relevant Issue: https://github.com/ministryofjustice/modernisation-platform/issues/2400

This PR creates GitHub secrets in the `modernisation-platform-environments` repo which contain the list of accounts to nuke, rebuild after nuke and exclude from nuke. 

- By default, all sandbox accounts will now be automatically added to the nuke list. This PR uses a newly-introduced optional "nuke" field to explicitly exclude `digital-prison-reporting-development` from nuke
- Our nuke process runs in two parts: the nuke which with some exceptions removes existing resources from an account and rebuild-after-nuke which then runs the corresponding terraform to rebuild the account to the state defined in code. By default, the accounts are NOT rebuilt. Setting `"nuke"` field to `"rebuild"` triggers a rebuild after nuke for the account. The json 
files for the accounts currently being rebuilt have been modified with the appropriate value for nuke. 

The code that generates the values for the new GH secrets is here: https://github.com/ministryofjustice/modernisation-platform-terraform-environments/pull/23


davidkelliott (approved):
👏

PR #2777, opened 2022-12-07
Add pagerduty service for Nomis

davidkelliott:
Adding a Pagerduty service and integration for Nomis, once created we will manually associate the service with the slack channel (this cannot be coded unfortunately).

This also creates a new dead end escalation policy which goes through to our dead end user.  This saves having to create the required escalation policy, schedule and user for teams which only want to use this for slack integration.


ewastempel (approved):
LGTM

PR #2778, opened 2022-12-07
Add a sandbox role for collaborators

davidkelliott:
This has the same permissions as the internal sandbox role.

This role will only be created in member developer accounts.

Moved all collaborator roles into a new separate file as it was getting confusing.

Further work is required to enable collaborators to assume this role, but that will be in a separate PR.


julialawrence (approved):
LGTM. It's nice splitting them out into a separate tf file!

PR #2780, opened 2022-12-07
Move RAM share to script and fix

davidkelliott:
The code needed to work out which accounts to run the RAM share on was getting to complex to account for all the scenarios.

The moves the logic for identifing which accounts to run the RAM share on into a separate script.

If this works then it would probably make sense to combine both the scripts, keeping them separate for now though as we won't be able to verify that it works until the next new account.

High level logic is as follows:

```
Get list of environments-network changes
If there are changes
  For all accounts in a changed file:
    Check if the required networking.auto.tfvars.json is present
      If so run the RAM share terraform
      If not skip

Get list of new networking.auto.tfvars.json files
If there are new files
  For all the files
    Run the RAM share terraform
```

This should fix the issues we've been having, including RAM share failing with no networking.auto.tfvars.json, not running if networking.auto.tfvars.json changes, not running if environment-networks changes.


ep-93 (approved):
LGTM, happy to run locally if you want another person to test though.

PR #2789, opened 2022-12-08
Adjustments to VPN configuration

dms1981:
* Hard coded inside tunnel addresses for VPN to Azure
* Programatic advertisement of MP network ranges over VPN connections (by advertising the `external` route table)


davidkelliott (approved):
👏

PR #2790, opened 2022-12-08
Enable security scanning on repos

davidkelliott:
This was manually enabled on some repos, adding to code for all.


julialawrence (approved):
Lgtm

PR #2796, opened 2022-12-09
Add MGN user for PPUD

davidkelliott:
This user is required for the local MGN agents to transfer data to the MGN service.


julialawrence (approved):
LGTM

PR #2818, opened 2022-12-13
Update team.html.md.erb

seanprivett:
All correct, just changed reviewed date. Although we should revisit early in the new year.


julialawrence (approved):
👍

PR #2819, opened 2022-12-13
Change name from northgate to six degrees

davidkelliott:
Naming confusion, these are actually two different data centres, we are connecting to six degrees not northgate.


SteveLinden (approved):
OK

PR #2823, opened 2022-12-13
Add Directory Services permissions to sandbox

davidkelliott:
Enabling sandbox role collaborators to make changes to directory services.  These permissions match what are available on the pipeline.


ewastempel (approved):
LGTM

Issue #1959, opened 2022-07-04
Switch to SSO user for local tests


davidkelliott:
## User Story

To run tests locally you need to get the testing-ci user credentials from secrets manager.

A better way to do this would be to use the SSO admin user for the testing-test account, that way the credentials are only temporary and we don't need a separate helper script.  Changes needed for this to work:

Update MemberInfrastructureAccess role permissions for the testing-test account only, to allow the SSO admin user to assume the role.
Update the state bucket and kms key permissions to allow the SSO admin user to read and write to the state file.

## Definition of done

- [x] Update MemberInfrastructureAccess role permissions
- [ ] Update the state bucket and kms key permissions to allow the SSO admin user to read and write to the state file
- [ ] readme has been updated
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
Update to state bucket already done in other SSO work
https://github.com/ministryofjustice/modernisation-platform/pull/2824

PR #2824, opened 2022-12-14
Enable local tests to be run with testing-test SSO

davidkelliott:
- Gives the testing-test SSO admin user permission to assume the MemberInfrastructureAccess role used in testing.
 - Removes the script to get the testing-ci user credentials as this is no longer needed.
 - Update the testing documentation.


davidkelliott:
Closes #1959

ewastempel (approved):
LGTM

PR #2825, opened 2022-12-14
Add additional GUI perms for sandbox collaborators

davidkelliott:
These permissions are available to the developer role but missing from the collaborator sandbox role.


julialawrence (approved):
LGTM

PR #2830, opened 2022-12-14
Change from static to dynamic github app teams

davidkelliott:
- dynamically generate a list of the application teams to be added to our repo from the json files, rather than using a static list that needs to be updated on new account creation.
- modify the workflow to run on json file changes
- update documentation


julialawrence:
Well this is pretty awesome!

julialawrence (approved):
LGTM

Issue #2576, opened 2022-11-08
Automate the Rotation of testing-test ci user AWS credentials


julialawrence:
## User Story

As a modernisation platform developer, I would like to implement automated rotation for `testing-test` ci user to reduce the security impact of using static credentials in situations where converting OIDC is not practicable.

## User Type(s)

Modernisation Platform engineers

## Value

We are unable to convert GHAs that run against the `testing-test` account to OIDC because that would prevent us from testing code that creates OIDC providers. The usage of a Modernisation Platform OIDC provider is also not ideal as that would negate the value of running tests in an isolated account. Therefore, the best approach would be to continue using `testing-test` ci credentials. Introducing regular automated rotation would reduce the security and operational issues around the usage of static creds.

### Hypothesis 
If we introduce automated rotation of testing-ci user AWS credentials that will mitigate/reduce the risk of using static credentials

### Proposal

* Move the creation of the testing-ci `aws_iam_access_key` resource and from `terraform/environments/testing` to `terraform/github`
* Link it via a lifecycle rule to a `null_resource` that gets recreated once a week. (potentially setting the trigger to year, month, week extracted from a timestamp)
* Line up order of operation so that recreation of the keys pushes the value to the secrets manager secret and then triggers the update of all the github repos that use the credentials with the new secret values.
* Schedule a rerun of the github workflow once a week.

## Definition of done

- [x] testing-test credentials updated automatically every 7 days.
- [x] readme has been updated
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

PR #2837, opened 2022-12-15
feature/rotate testing ci creds

julialawrence:
Relevant Issue: https://github.com/ministryofjustice/modernisation-platform/issues/2576

The resources relating to the `testing-ci` user, policies, secrets and keys have been moved to the GitHub folder because that is where they're chiefly consumed. Managing the aws access keys of the testing-ci user in the same folder where the github secrets are managed minimises the chance of drift between the actual key values and the contents of the github secrets.

The rotation works as follows:

- A `time_rotating` resource is created which (right now) is set to recreate every 30 minutes but once tested will recreate every 7 days.
- Due to [this issue](https://github.com/hashicorp/terraform-provider-time/issues/118), the `time_rotating` resource can't trigger a recreation of another resource directly. Therefore an intermediate `time_static` resource is required which is *changed* rather than *recreated* every rotation period. 
- The `time_static` resource is linked to the `aws_iam_access_key` resource via a `replace_triggered_by` lifecycle rule, triggering a replace operation every rotation period.
- The github secrets which consume the values of the access key resource via the `secrets_manager_secret_version` are recreated at the same time.
- The github workflow will be modified to run every 7 days to make sure that the code is run at least once during each rotation period and triggers the key rotation
  

Changes:
- Adding the code for the management and rotation of testing-ci user credentials. The rotation period is controlled via time_rotating resource but the replace_triggered_by lifecycle condition is triggered by the time_static resource due to the current implementation of teh time_rotating resource. Relevant resources have been manually imported into the github state.
- Removing testing-ci user code and related resources from the testing folder. The resources have been manually removed from the state.

To do: 
- Fix tagging
- Amend README
- Set the rotation period to 7 days
- Add schedule to github workflow


davidkelliott (approved):
Awesome!  I'm going to close of sooo many risk register issues now :D

PR #2888, opened 2022-12-22
Add migration role

davidkelliott:
This will be a role temporarily granted to teams to allow them to use services such as application migration service and database migration service in all accounts.


source/user-guide/creating-environments.html.md.erb
```diff
@@ -66,6 +66,16 @@ The idea of sandbox access is that it gives you freedom to experiment and modify
...
 
+#### migration
+
+The migration role is a temporary role, given to assist with AWS managed migration services
+
+- Log into the console
+- Developer permissions
+- Permissions to use [AWS Migration services](https://docs.aws.amazon.com/whitepapers/latest/aws-overview/migration-services.html)
+
+Once the migration of an application is complete this role should be removed.  No in production applications should have this role.
```
jakemulley, line 77: Should this guidance slightly change to something akin to "this role should be removed once used for the migration"? It currently reads as `production` _environments_ should not have this role - but if you're migrating a production database, you probably wouldn't want to copy it to a lower environment first?

jakemulley (changes requested):

terraform/environments/bootstrap/delegate-access/policies.tf
```diff
@@ -321,3 +321,155 @@ data "aws_iam_policy_document" "sandbox_additional" {
...
+      "iam:UpdateGroup",
+      "iam:UpdateLoginProfile",
+      "iam:UpdateOpenIDConnectProviderThumbprint",
+      "iam:UpdateSAMLProvider",
+      "iam:UpdateUser"
+    ]
+    resources = ["*"]
+  }
+
+  statement {
```
jakemulley, line 446: For a future PR, is it worthwhile refactoring the `Deny` statements which are the same across `view-only, developer, sandbox, administrator, migration`, to use [source_policy_documents](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#example-using-a-source-document)?

jakemulley (approved):

PR #2917, opened 2023-01-04
Runbook for emergency patching added, with details on each step needed.

ep-93:
This is a runbook of the stages needed to follow to run the patching cross-account. If it needs more detail, or screenshots, I am happy to do that, just comment away with some feedback.

* It may seem sparse, but a lot of ties into AWS supplied tools / scripts which I have linked in the documentation.


SteveLinden (approved):
May need an example yaml file but OK otherwise

PR #2923, opened 2023-01-05
Remove traffic selector for NOMS VPNs

dms1981:
In order to accommodate the requirements for the route-based VPN on the Azure side, we need to remove the specific traffic selector from the AWS side.  By having AWS initiate the VPN and propose a specific CIDR while Azure is expecting to see `0.0.0.0/0`, there is an IKE phase 2 mismatch that prevents a tunnel from being established.


ep-93 (approved):
Very open, however thanks for the explanation, I understand why. Approving as long as Dave is happy with it.

dms1981:
> Very open, however thanks for the explanation, I understand why. Approving as long as Dave is happy with it.

For context, though, traffic is still subject to network ACLs and security groups on the AWS side, as well as the AWS Network Firewall, and then whatever security tools are in place on the Azure side of the VPN.

PR #2925, opened 2023-01-05
DSO - update accounts and revert permission

jnq:
give our -development accounts permissions for sandboxing, to allow for easier testing


ewastempel (approved):
LGTM

PR #2935, opened 2023-01-06
Initial work to split out member boot strapping

### LINKED ISSUE #2460: Bug: new account creation process fails due to order of operations
## Expected Behavior

During the new account creation process, the `delegate-access` step should run successfully.

## Actual Behavior

During the new account creation process, the `delegate-access` step fails with the following error:
```
Error: Null value found in list

  with module.member-access[0].data.aws_iam_policy_document.assume-role-policy,
  on .terraform/modules/member-access/main.tf line 2, in data "aws_iam_policy_document" "assume-role-policy":
   2: data "aws_iam_policy_document" "assume-role-policy" {

Null values are not allowed for this attribute value.
```
On re-running the workflow, the `delegate-access` step works. This appears to be because the missing resource is created in parallel during the `single-sign-on` step.

* Example of failing run: https://github.com/ministryofjustice/modernisation-platform/actions/runs/3296446510/jobs/5436243462#step:5:2123

davidkelliott:
Fixes #2460

The member boot strapping is included in delegate access. This causes issues with order of resource creation as the main delegate access role created (ModernisationPlatformAccess) has to be created before anything else, but there are things in delegate access currently that need to be create after other workflows.

**This PR does the following:**

- Adds Terraform for a member-bootstrap folder
- Removes MemberInfrastructureAccess role from delegate-access
- Removes SSO references from the ModernisationPlatformAccess role, these are not needed as the MP account is already allowed.
- Removes SSO locals from delegate access which where causing the ordering issue
- Adds MemberInfrastructureAccess role to member-bootstrap
- Modifies sheduled baselines to include the member-bootstrap terraform
- Modifies new-environment to include the member-bootstrap terraform
- Amended provision-terraform-workspaces.sh to include new member-bootstrap terraform
- Update the bootstrap plan for sprinkler

**Testing completed:**
Ran provision-terraform-workspaces.sh locally to create all new member-bootstrap workspaces and to verify script changes worked.

New workflow for baselines tested here - https://github.com/ministryofjustice/modernisation-platform/actions/runs/3883478358 (also tested on sprinkler)

Local SSO Plans tested for Sprinkler and Example
environments repo workflow tested with sprinkler - https://github.com/ministryofjustice/modernisation-platform-environments/actions/runs/3883057425/jobs/6623929622


terraform/environments/bootstrap/delegate-access/iam.tf
```diff
@@ -17,8 +17,7 @@ module "cross-account-access" {
   account_id             = local.modernisation_platform_account.id
   policy_arn             = "arn:aws:iam::aws:policy/AdministratorAccess"
   role_name              = "ModernisationPlatformAccess"
-  additional_trust_roles = concat(tolist(data.aws_iam_roles.mp-sso-admin-access.arns), terraform.workspace == "testing-test" ? ["arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:user/testing-ci"] : [])
-
+  additional_trust_roles = terraform.workspace == "testing-test" ? ["arn:aws:iam::${local.environment_management.account_ids[terraform.workspace]}:user/testing-ci"] : []
```
ep-93, line 20: Formatting looks strange, but I think that's just in my head.

davidkelliott: Yes it does! I had to do a double take then it's the conditional test which makes it look odd.

PR #2950, opened 2023-01-10
Initial code for TGW Flow logs

markgov:
code still needs to be tested do not merge

### DIFF
--- a/terraform/environments/core-network-services/locals.tf
+++ b/terraform/environments/core-network-services/locals.tf
@@ -29,6 +29,50 @@ locals {
     "hmpps-production-attachment",
     "core-network-services-live_data-attachment",
   ]
+  
+  active_tgw_all_attachments = [
+    "cjse-production-attachment",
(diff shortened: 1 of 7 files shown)


terraform/environments/core-network-services/locals.tf
```diff
@@ -30,6 +30,50 @@ locals {
     "core-network-services-live_data-attachment",
   ]
 
+  active_tgw_all_attachments = [
```
davidkelliott, line 33: Could we get this list dynamically so we don't need to manually update it?  It might need to be a concat of a few different things

davidkelliott (changes requested):

dms1981 (approved):

davidkelliott (approved):

davidkelliott (changes requested):
Could you resolve the SCA warnings please?

davidkelliott (approved):
Looks good to me! 👏 First PR on MP!

PR #2962, opened 2023-01-13
updated getting-aws-credentials to reflect relevance

dms1981:
Our user guide doesn't make clear that the AWS SSO landing page is for Ministry of Justice employees, and that external collaborators should review different guidance. Given that we've had a few customers query us about this, it's worth clarifying for them.


ewastempel (approved):
LGTM

PR #2964, opened 2023-01-13
change for flow group creation

markgov:
Change made to the kms_key_id  as the documentation for the resource is actually looking for the arn of the kms key not the key id

[kms_key_id](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group#kms_key_id) - (Optional) The ARN of the KMS Key to use when encrypting log data. Please note, after the AWS KMS CMK is disassociated from the log group, AWS CloudWatch Logs stops encrypting newly ingested data for the log group. All previously ingested data remains encrypted, and AWS CloudWatch Logs requires permissions for the CMK whenever the encrypted data is requested.


dms1981 (approved):
Approved, but those `user-guide` files aren't relevant to the commit.

PR #2975, opened 2023-01-17
suggested code changes

markgov:
removed the need for encryption for now on the flow logs as we can deal with this on a separate story. made suggested changes to the code to make resource name shorter and changed to dashes instead of underscores i have also added tagging to the resource blocks as well.


ewastempel (dismissed):
LGTM

PR #2977, opened 2023-01-17
Feature/s3 ami bucket iam

ep-93:
Worked out why it hated me adding the policy, human error.

Didn't like       "s3:ListBucket"

Also didn't like how I called the module I think.

Also didn't like me missing a /*


ewastempel (approved):
LGTM

PR #2979, opened 2023-01-17
changes to the depenency and also adding destination type

markgov:
added dependancy and log type to flow log


ewastempel (approved):
LGTM - re-approving after formatting

PR #2984, opened 2023-01-18
Set max aggregation interval

markgov:
I have set the max aggregation interval as the documentation states that for transit gateway id it has to be set to 60


SteveLinden (dismissed):
Look OK to me

davidkelliott (approved):

PR #2985, opened 2023-01-18
Prevent VPN resource thrashing

dms1981:
Our VPN code relies on terraform default values for attribute such as `tunnel1_ike_versions`. This means that AWS then fills in the blanks by writing in the default values. However, on re-running terraform the `aws_vpn_connection` resource sees those defaults but does not expect them and attempts to remove them, leading to the AWS API to recreate them.

In order to prevent this thrashing, I've added a lifecycle `ignore_changes` statement to cover the relevant attributes.


davidkelliott (approved):
👏 good idea, thanks for this

PR #2987, opened 2023-01-18
new ccms ebs account

markgov:
CCMS is a system that manages certified civil legal aid applications, amendments and bills, it is our general ledger, chart of accounts and payment system and manages accounts receivable.

we have been asked to create new account for test preproduction and production

This pr set up these new accounts


ewastempel (approved):
LGTM

PR #2994, opened 2023-01-20
Add view-only and administrator access for management account.

julialawrence:
This PR adds SSO Admin and View-Only to the AP management account.


davidkelliott (approved):
👍

Issue #2980, opened 2023-01-18
New networking files created for OAS when they shouldn't be


davidkelliott:
## Expected Behavior

No new files created for OAS on new environment files run 

## Actual Behavior

These PRs are created on new environment files run:

https://github.com/ministryofjustice/modernisation-platform/pull/2976
https://github.com/ministryofjustice/modernisation-platform-environments/pull/1281

For some reason it thinks that oas should be under the HMPPS business unit.  Potentially this is an issue due to the fact there is an oasys application in hmpps.

Could be a partial string match being used to identify

I believe the offending lines might be the contains() in here:

https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-member-directories.sh#L83
https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-member-directories.sh#L79
https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-environment-directories.sh#L73
https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-member-directories.sh#L79

## Steps to Reproduce the Problem

  1. Run new environment files workflow
  1.
  1.

dms1981:
I had a conversation with ChatGPT about this problem, based on the lines that cause issues leveraging `contains()`. Here are two possible solutions:

```
RAW_OUTPUT=`jq --arg APPLICATION_NAME "$application_name" 'limit(1;.[].subnet_sets[] | select(.accounts[] | . == $APPLICATION_NAME) | { "business-unit": ."business-unit", "set": .set, "application": $APPLICATION_NAME } )' <<< "$networking_definitions"`
```
This version replaces `contains($APPLICATION_NAME)` with `. == $APPLICATION_NAME` to check for an exact match of the application name.

```
RAW_OUTPUT=`jq --arg APPLICATION_NAME "$application_name" 'limit(1;.[].subnet_sets[] | select(.accounts[] | . as $a | $a |= ($a | split("\n")) | . as $accounts | $accounts |= map(select(. == $APPLICATION_NAME)) | .[0]) | { "business-unit": ."business-unit", "set": .set, "application": $APPLICATION_NAME } )' <<< "$networking_definitions"`
```
This version uses grep command inside jq to filter the accounts array based on exact match.

Naturally these need to be considered, but given your analysis they're promising approaches!

davidkelliott:
https://github.com/ministryofjustice/modernisation-platform/pull/3023

PR #3023, opened 2023-01-26
Fix network file creation bug

davidkelliott:
Fixes new networking files created for OAS when they shouldn't be

Previously contains was picking up oasys for oas as it contains oas

By switching to test() we can add the start of line ^ and the - dash before the environment line.

So this will work for:

oas-development
oasys-development

We might stil have an issue if something came along with the name

oas-sys-development

But let's hope that doesn't happen.

### DIFF
--- a/scripts/provision-environment-directories.sh
+++ b/scripts/provision-environment-directories.sh
@@ -66,11 +66,11 @@ provision_environment_directories() {
     # ]
 
     # check if /environments-networks files exists for this application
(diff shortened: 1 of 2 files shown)


davidkelliott:
Fixes [#2980](https://github.com/ministryofjustice/modernisation-platform/issues/2980)

dms1981 (approved):

PR #3029, opened 2023-01-27
Set VPN connections to 6DG to use static addresses only

dms1981:
* Amends `vpn_connection` resource to support `static_routes_only` argument
* Set 6DG VPNs to use `static_routes_only` in the configuration for their `vpn_connections`. The CGWs will still contain BGP configuration as this is a required field


markgov (approved):
LGTM

PR #3038, opened 2023-01-31
Add secret rotation docs and move secrets to SM

davidkelliott:
Adding documentation on how to rotate our secrets.

Also creating secrets manager secrets for the MP GitHub CI user secrets which we'll be moving out of password management into GitHub secrets.


ewastempel (approved):
LGTM

PR #3039, opened 2023-01-31
Update github to pull secrets from secrets manager

davidkelliott:
Now that the PAT is stored in secrets manager it can be updated in code rather than manually in the GitHub gui.


ep-93:
As long as those 2 tokens being replaced are okay, then yes! Approved.

PR #3048, opened 2023-02-01
Change code-formatter task to use PAT

dms1981:
* Change to use CI User PAT instead of github token
* This is because the GitHub token can't be used to kick off a workflow following a push
* [Triggering a workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#:~:text=For%20example%2C%20if%20a%20workflow%20run%20pushes%20code%20using%20the%20repository%27s%20GITHUB_TOKEN%2C%20a%20new%20workflow%20will%20not%20run%20even%20when%20the%20repository%20contains%20a%20workflow%20configured%20to%20run%20when%20push%20events%20occur.)

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -14,6 +14,7 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
         with:
             persist-credentials: false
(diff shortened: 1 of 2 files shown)


dms1981:
After some discussion with @davidkelliott I'm going to close this PR & branch, and write something that runs the code formatter on a schedule, and creates a PR for review.

PR #3049, opened 2023-02-01
route53_module_#2972

SteveLinden:
This routine is to create the logging for the route53 DNS that we have in place. To build the logging mechanism we need to do this per VPC. The module to pass through VPC details to set up logging. This routine will be called from the core-vpc section of this repo and will be an additional PR

The log build is in the monitoring section under core-logging


terraform/modules/route53-logging/variables.tf
```diff
@@ -0,0 +1,11 @@
+tags = {
```
dms1981, line 1: Tags don't need to be supplied in a module, and won't work in this format. Consider having a variable that tags a map here instead.

terraform/modules/route53-logging/main.tf
```diff
@@ -0,0 +1,15 @@
+
+
+resource "aws_route53_resolver_query_log_config" "dns_logs" {
+  name            = "dns_logs"
+  destination_arn = aws_cloudwatch_log_group.aws_route53_logs_com.arn
+
+  tags = {
```
dms1981, line 7: `tags` map is empty and will need to be supplied with something. Perhaps a variable?

terraform/environments/core-sandbox/playground.tf
```diff
@@ -19,16 +19,16 @@ module "ram-principal-association" {
...
-  #   values = ["oracle-linux-5.11*"]
-  # }
-
-}
-
-output "amis" {
-  value = data.aws_ami_ids.example.ids
-}
\ No newline at end of file
+# data "aws_ami_ids" "example" {
```
dms1981, line 22: Comments here should either be removed, or the regex should be fixed as this isn't relevant to the purpose of this PR

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,44 @@ module "core_monitoring" {
...
+resource "aws_cloudwatch_log_resource_policy" "route53-query-logging-policy" {
+  provider = aws.us-east-1
+
+  policy_document = data.aws_iam_policy_document.route53-query-logging-policy.json
+  policy_name     = "route53-query-logging-policy"
+}
+
+# Example Route53 zone with query logging
+
+resource "aws_route53_zone" "logs_com" {
```
dms1981, line 43: There's no need to create a new route53 zone is there?

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,44 @@ module "core_monitoring" {
...
+    resources = ["arn:aws:logs:*:*:log-group:/aws/route53/*"]
+
+    principals {
+      identifiers = ["route53.amazonaws.com"]
+      type        = "Service"
+    }
+  }
+}
+
+resource "aws_cloudwatch_log_resource_policy" "route53-query-logging-policy" {
```
dms1981, line 34: The provider here won't work

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,44 @@ module "core_monitoring" {
   source                     = "../../modules/core-monitoring"
   pagerduty_integration_keys = local.pagerduty_integration_keys
 }
+
+provider "aws" {
+  alias  = "eu-west-2"
+  region = "eu-west-2"
+}
+
+resource "aws_cloudwatch_log_group" "aws_route53_logs_com" {
```
dms1981, line 11: Does a provider need to be hardcoded? Isn't the default provider sufficient?

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,44 @@ module "core_monitoring" {
   source                     = "../../modules/core-monitoring"
   pagerduty_integration_keys = local.pagerduty_integration_keys
 }
+
+provider "aws" {
```
dms1981, line 6: What purpose does this provider fill? If it's needed it ought to be created in `providers.tf` for consistency / ease of comprehension

terraform/modules/route53-logging/main.tf
SteveLinden: I was going to add the ones I put in the variable list. I'll have a think

terraform/modules/route53-logging/main.tf
```diff
@@ -0,0 +1,15 @@
+
+
+resource "aws_route53_resolver_query_log_config" "dns_logs" {
+  name            = "dns_logs"
+  destination_arn = aws_cloudwatch_log_group.aws_route53_logs_com.arn
```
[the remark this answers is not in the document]
SteveLinden: My plan is to build that first. Once it is in place I could then find the actual arn and use that. Not entirely sure that will work how I want but that was my thought process

terraform/environments/core-sandbox/playground.tf
SteveLinden: That's the section I commented out to get the plan to run. We need to look at it to see if the code can be amended to work correctly.

terraform/environments/core-logging/monitoring.tf
SteveLinden: Forgot to remove this code

terraform/environments/core-logging/monitoring.tf
SteveLinden: Copy and pasted the resource from terraform but forgot to remove the provider

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,29 @@ module "core_monitoring" {
   source                     = "../../modules/core-monitoring"
   pagerduty_integration_keys = local.pagerduty_integration_keys
 }
+resource "aws_cloudwatch_log_group" "aws_route53_logs_com" {
```
dms1981, line 5: Can you update the resource name here to something like `"route53-resolver-logs"`?

terraform/environments/core-sandbox/playground.tf
```diff
@@ -19,16 +19,16 @@ module "ram-principal-association" {
 
 }
```
dms1981, line 21: As previously noted, this file doesn't need to be part of the commit. Can you remove the comments from the start of the lines here and raise an issue to fix this data call / take it on as a separate task?

terraform/modules/route53-logging/variables.tf
SteveLinden: Put in variables

terraform/modules/route53-logging/main.tf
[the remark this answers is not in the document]
SteveLinden: Amended to sort that

terraform/environments/core-logging/monitoring.tf
SteveLinden: Amended to aws_route53_logs I think

terraform/environments/core-logging/providers.tf
```diff
@@ -27,4 +27,8 @@ provider "aws" {
   assume_role {
     role_arn = "arn:aws:iam::${local.environment_management.account_ids["core-network-services-production"]}:role/ModernisationPlatformAccess"
   }
+}
+provider "aws" {
```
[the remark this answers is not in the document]
SteveLinden: Left for the time being. May be removable.

terraform/environments/core-sandbox/playground.tf
SteveLinden: Will do

terraform/modules/route53-logging/variables.tf
```diff
@@ -0,0 +1,11 @@
+
```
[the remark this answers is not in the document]
SteveLinden: Amended

davidkelliott (changes requested):
Hey, I've added a few additional comments, it would also be good in the PR description if you could explain in more detail what the PR is doing, what additional work is still needed etc.

terraform/environments/core-logging/monitoring.tf
```diff
@@ -2,3 +2,29 @@ module "core_monitoring" {
...
+  name              = "aws_route53_zone.logs_com.name"
```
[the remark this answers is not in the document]
SteveLinden: Name changed but the plan was to move this out of this PR. I've created a new one which will contain, hopefully, ny other changes made to that.

terraform/environments/core-logging/providers.tf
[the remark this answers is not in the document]
SteveLinden: Change made in another PR so it won't show in this one

terraform/modules/route53-logging/variables.tf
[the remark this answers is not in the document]
SteveLinden: Commented out code removed

terraform/environments/core-sandbox/playground.tf
SteveLinden: Comments removed. Will raise an issue to fix this. Uses a very old version of linux so may no longer be available

terraform/environments/core-sandbox/playground.tf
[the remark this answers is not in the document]
SteveLinden: Removed the above from the PR so the original will stay in place.

terraform/environments/core-sandbox/playground.tf
SteveLinden: Removed the above from the PR so the original will stay in place.

terraform/environments/core-sandbox/playground.tf
SteveLinden: Removed from the PR

terraform/modules/route53-logging/locals.tf
```diff
@@ -0,0 +1,8 @@
+locals {
```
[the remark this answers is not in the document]
SteveLinden: The above changes have been made so this is resolved now

terraform/environments/core-logging/monitoring.tf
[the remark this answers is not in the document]
SteveLinden: New routine (in the other PR) has been added and this have been moved

terraform/modules/route53-logging/main.tf
```diff
@@ -0,0 +1,15 @@
+
+
+resource "aws_route53_resolver_query_log_config" "dns_logs" {
+  name            = "dns_logs"
+  destination_arn = "arn:aws:logs:eu-west-2:${local.environment_management.account_ids[terraform.workspace]}:log-group:aws_route53_logs:*"
```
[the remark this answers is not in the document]
SteveLinden: Changes made to this to pass in a variable which will be made available from the call in core-vpc (in another PR)

SteveLinden:
The description was amended to add some more detail. I believe that this now explains the purpose in more detail.

PR #3051, opened 2023-02-01
More SCA checks

markgov:
More Errors found in SCA added to call details and added ignores to the alert

### DIFF
--- a/terraform/modules/iam_baseline/main.tf
+++ b/terraform/modules/iam_baseline/main.tf
@@ -8,9 +8,12 @@ resource "aws_iam_user" "cicd_member_user" {
 resource "aws_iam_group" "cicd_member_group" {
   name = "cicd-member-group"
 }
-
+#Adding checkov skips into bellow but this will probably need to be looked at in a later story#
 #tfsec:ignore:aws-iam-no-policy-wildcards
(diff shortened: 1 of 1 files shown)


davidkelliott (commented):
You can remove the comment about resolving later, this is fine as we can't restrict access to resources any further because this possibly is applied to all accounts and we don't know what the resources will be

terraform/modules/iam_baseline/main.tf
```diff
@@ -8,9 +8,12 @@ resource "aws_iam_user" "cicd_member_user" {
 resource "aws_iam_group" "cicd_member_group" {
   name = "cicd-member-group"
 }
-
+#Adding checkov skips into bellow but this will probably need to be looked at in a later story#
```
ewastempel, line 11: have you already created a follow on story to look at it then?
I am worried we will skip them checks and then never look at it; so I would say, if you don’t think it’s resolved, don’t skip them

markgov: removed line 11  as we can't restrict access to resources

ewastempel (commented):
LGTM

davidkelliott (approved):

PR #3052, opened 2023-02-01
Disable auto-nuke on delius-jitbit dev environment

adeweetman-al:
Disable the auto-nuke feature in the jitbit dev environment.
This is likely to be temporary until we have the app pipeline built which we're currently working on. At which point, we should remove this from being a sandbox account.


ewastempel (approved):
LGTM

PR #3055, opened 2023-02-02
Update s3.tf in core-shared-services for ami s3 bucket.

ep-93:
Update s3.tf in core-shared-services for ami s3 bucket.

Added 2 permissions.


markgov (approved):
LGTM

PR #3062, opened 2023-02-02
Revert "Update s3.tf in core-shared-services for ami s3 bucket."

ep-93:
Reverts ministryofjustice/modernisation-platform#3055


markgov (approved):
LGTM

Issue #2521, opened 2022-10-28
Change some terraform template files to "platform_" files


davidkelliott:
## User Story

As an MP engineer
I need to make changes to template files and role them out easily across the platform
So that we can make updates and improvements consistently and reliably

We have some files, eg data.tf and locals.tf that we populate with a lot of platform specific things ready for users to use.  However users are adding their own data and locals to these files, which makes updating them much harder.

This ticket is to split out user data and locals into separate files, by having a platform_data.tf and platform_locals.tf in the same way we have a platform_secrets.tf

We should state at the top of these files - DO NOT AMEND PLATFORM files, or something similar, and add them to the codeowners so that only MP team can review changes to them.

This will also allow the potential in future for automating the update of these files when the template files are updated rather than having to do this manually across all the applications.

There are some files such as locals.tf and data.tf that we generate from a template. Users have added to these files, which means if we want to update them we can't just copy and past over the whole file in a bulk operation, we need to go in individually to each file and update the changed parts, increasing the time and complexity of the activity.

This task is to split these files out into their own file and rename them so it's clear this is for platform use only, eg:

platform_data.tf
platform_locals.tf

(note there may be more files the above 2 are just and example)

Then protect these files from changes with codeowners

Potentially we could dry these out with terragrunt but that would be another card

## Definition of done

- [ ] template files amended
- [ ] files split out and amended in the mp-env-repo
- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

ewastempel:
This ticket will require making changes in the `modernisation-platform-environments` for all template files within each environment.

Additionally, if it is not already implemented, it will need changes in `modernisation-platform` repo to update the file names for when a new account is created.

davidkelliott:
- [ ] apex
- [ ] ccms-ebs
- [ ] cooker
- [ ] data-and-insights-wepi
- [ ] delius-iaps
- [ ] delius-jitbit
- [ ] digital-prison-reporting
- [ ] equip
- [ ] example
- [ ] laa-oem
- [ ] long-term-storage
- [ ] maatdb
- [ ] mlra
- [ ] nomis
- [ ] nomis-combined-reporting
- [ ] oas
- [ ] oasys
- [ ] performance-hub
- [ ] ppud
- [ ] refer-monitor
- [ ] sprinkler
[...]

PR #3066, opened 2023-02-03
Moving resources around within template files

ewastempel:
Some resources in the template files do not belong to the files that they are in, hence shifting them around. This is needed as part of the https://github.com/ministryofjustice/modernisation-platform/issues/2521 ticket.

Renaming files and generating files in the environments changes will be raised in a separate PR.


davidkelliott (approved):
👍

PR #3067, opened 2023-02-03
Changes to allow one providers file

davidkelliott:
Adding the ssm parameter allows the data source to function regardless of the user source, SSO or IAM calling it.

Allowing collaborators to assume the read only role in the modernisation platform, again this gives consistency to the providers file.


ewastempel (approved):
LGTM, although it could do with rearranging resources, probably in a separate PR

Issue #2972, opened 2023-01-16
Enable DNS logs


davidkelliott:
## User Story

As a security team
I want to view DNS logs
So that I can identify abnormal or malicious activity

## User Type(s)

Security team

## Value

Be able to identify abnormal or malicious activity in DNS and therefore improve our security posture and become aware of security incidents earlier.

### Hypothesis 
If we enable DNS logs
Then a security team will be able to analyse and detect issues

Talk with security team/Steve to agree what exactly we need to implement

Check MOJ guidelines for DNS logging

### Proposal

In order to implement Route53 Resolver Query Logging we will need to do the following:
* Create one of more CloudWatch Log Groups to hold the logs in `core-logging` (potentially one for `live_data` and one for `non_live_data`)
* Create a reusable module that can take in a variable for the log group ARN and the ID of a VPC
* Call this reusable module with a for_each expression in each terraform directory that creates per-business-unit VPCs and supply it with the relevant VPC ID and log group ARN

## Definition of done

- [x] CloudWatch log group(s) created
- [x] Reusable module created
- [x] Module implemented to add logging wherever we create a VPC
- [x] DNS logging enabled
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

SteveLinden:
The DNS definitions appear to all be under modernisation platform account. To add we need to create a profile (see below) and create a log group. Note, some of the route 53 environments do not, currently, have the option of adding query logging as an option so these may require additional changes/settings.

Required profile is here ...
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "Route53LogsToCloudWatchLogs",
      "Effect": "Allow",
      "Principal": {
        "Service": [
          "route53.amazonaws.com"
        ]
      },
      "Action": [
        "logs:CreateLogStream",
        "logs:PutLogEvents"
      ],
      "Resource": "arn:aws:logs:us-east-1:278663825216:log-group:*"
    }
  ]
}

SteveLinden:
If the comment I have seen is true this might not be possible on eu-west-2

"There are restrictions on the configuration of query logging. Notably, the CloudWatch log group must be in the us-east-1 region"

dms1981:
Speaking with @SteveLinden this story is a bit unclear so I'm going to update the description to cover what's required.

[...]

PR #3077, opened 2023-02-06
core-logging-route53_logging-#2972

SteveLinden:
Create (and give access to) the log files in route53_logging.tf. This log will be used in the module route53-logging but this needs to be completed prior to that.


davidkelliott (changes requested):
This still contains the module code which was going to be split into a separate PR

davidkelliott (changes requested):
Yeah it's still deleted I'm afraid

PR #3087, opened 2023-02-07
Database user / policy.

ep-93:


davidkelliott:
@ep-93 here is a previous PR adding a migration role in case this helps cover all the areas where changes are needed https://github.com/ministryofjustice/modernisation-platform/pull/2888/files

terraform/environments/bootstrap/single-sign-on/main.tf
```diff
@@ -53,6 +53,13 @@ data "aws_ssoadmin_permission_set" "migration" {
   name         = "modernisation-platform-migration"
 }
 
+data "aws_ssoadmin_permission_set" "database-mgmt" {
+  provider = aws.sso-management
+
+  instance_arn = local.sso_instance_arn
+  name         = "modernisation-platform-database-mgmt"
```
davidkelliott, line 60: The SSO plan is failing as the name is too long, maybe rename to mp-instance-management ?

terraform/environments/bootstrap/delegate-access/policies.tf
```diff
@@ -365,3 +365,129 @@ data "aws_iam_policy_document" "migration_additional" {
...
+  statement {
+    sid    = "lambdaAllow"
+    effect = "Allow"
+    actions = [
+      "lambda:InvokeFunction"
+    ]
+    resources = ["arn:aws:lambda:*:*:function:Automation*"]
+  }
+
+  statement {
```
davidkelliott, line 466: I think this statement should be removed as it's probably too excessive for this role

davidkelliott (changes requested):

SteveLinden (commented):
There's a double "/" on one line. Probably not an issue but it might be worth checking. I suspect it's because of copy and paste
source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role"

terraform/environments/bootstrap/delegate-access/policies.tf
```diff
@@ -111,7 +111,7 @@ data "aws_iam_policy_document" "developer_additional" {
       "codebuild:PersistOAuthToken",
       "ds:*Tags*",
       "ds:*Snapshot*",
-      "ec2:StartInstances",
+      "ec2:Start-s",
```
davidkelliott, line 114: Is this change intentional?

davidkelliott (approved):
Looks good, just double check the sprinkler plan runs fine before merging once the AWS root account changes have been deployed.

PR #3090, opened 2023-02-07
Adding platform_ prefix to templates when creating accounts

ewastempel:
This is a bit hefty, so take some time reviewing it please.

This PR is implementing https://github.com/ministryofjustice/modernisation-platform/issues/2521 and  is:

- renaming the following in templates directory:

```
terraform/templates/backend.tf -> terraform/templates/platform_backend.tf
terraform/templates/base_variables.tf -> terraform/templates/platform_base_variables.tf
terraform/templates/data.tf -> terraform/templates/platform_data.tf
terraform/templates/versions.tf -> terraform/templates/platform_versions.tf
terraform/templates/member-locals.tf -> terraform/templates/platform_locals.tf
terraform/templates/member-providers.tf -> terraform/templates/platform_providers.tf
```

- replacing references of the above files in scripts (including codeowners):

```
scripts/provision-environment-directories.sh
scripts/provision-member-directories.sh
scripts/provision-terraform-workspaces.sh
```

- introducing new templates for member accounts:

```
terraform/templates/member_data.tf
terraform/templates/member_locals.tf
terraform/templates/member_secrets.tf
```

- which then get renamed to:

```
data.tf
locals.tf
secrets.tf
```

- renaming oidc-session and oidc_session to original-session and original_session in platform_data and platform_providers


scripts/provision-member-directories.sh
```diff
@@ -155,7 +156,8 @@ EOL
 **/backend.tf @ministryofjustice/modernisation-platform
 **/subnet_share.tf @ministryofjustice/modernisation-platform
 **/networking.auto.tfvars.json @ministryofjustice/modernisation-platform
-/terraform/modules
+**/platform_*.tf @ministryofjustice/modernisation-platform
+/terraform/modules @ministryofjustice/modernisation-platform
```
davidkelliott, line 160: The /terraform/modules folder is for users to create their own modules without needing approval from us, so that can be left as it is with no codeowner

davidkelliott (dismissed):
Nice! I think you've caught everything, certainly a few things I hadn't thought about.  I spent a  while going through the name changes thinking they wouldn't work but then understanding the changes and coming to the conclusion that I think they will work!  You can manually run some of the scripts to test them if you wanted, you'll need to locally add a new environment file in the environments json, then run the scripts (you may need to modify the dir locations at the top) and it should create the files locally for you in a new application folder.

scripts/provision-member-directories.sh
ewastempel: updated

davidkelliott (approved):
LGTM

PR #3094, opened 2023-02-08
Turn on resolver logs in each customer VPC

dms1981:
* Add a new provider and data call to get ARN of CloudWatch Log Group in `core-logging` account
* Use a `for` expression in a `for_each` to get the VPC IDs in each workspace and apply the `r53-resolver-logs` module
* Remove an unsupported `tags` attribute from the `r53-resolver-logs` module


davidkelliott (approved):
👍

PR #3098, opened 2023-02-09
Fix logging location for route53 log files

SteveLinden:
The changes in the core-vpc earlier were not allowed due to log writing issues. This is to correct the name of the location to allow this access.

### DIFF
--- a/terraform/environments/core-logging/route53_resolver_logs.tf
+++ b/terraform/environments/core-logging/route53_resolver_logs.tf
@@ -15,7 +15,7 @@ data "aws_iam_policy_document" "route53-query-logging-policy" {
       "logs:PutLogEvents",
     ]
 
-    resources = ["arn:aws:logs:*:*:log-group:modernisation-platform-r53-resolver-logs:*"]
+    resources = ["arn:aws:logs:*:*:log-group:modernisation-platform-r53-resolver-logs"]
 
     principals {
       identifiers = ["route53resolver.amazonaws.com"]
--- a/terraform/environments/core-logging/route53_resolver_logs.tf
+++ b/terraform/environments/core-logging/route53_resolver_logs.tf
@@ -15,10 +15,10 @@ data "aws_iam_policy_document" "route53-query-logging-policy" {
       "logs:PutLogEvents",
     ]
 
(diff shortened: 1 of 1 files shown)


terraform/environments/core-logging/route53_resolver_logs.tf
```diff
@@ -15,7 +15,7 @@ data "aws_iam_policy_document" "route53-query-logging-policy" {
       "logs:PutLogEvents",
     ]
 
-    resources = ["arn:aws:logs:*:*:log-group:/aws/route53/*"]
+    resources = ["arn:aws:logs:*:*:log-group:route53-query-logging-policy"]
```
ewastempel, line 18: Looking at the available log groups in the core-logging account, I think you wanted to use: `modernisation-platform-r53-resolver-logs` log group.

so it would be something like this
`resources = ["arn:aws:logs:*:*:log-group:modernisation-platform-r53-resolver-logs:*"]`

However, I think it would be better if you narrow it down to that account as well, so this is not a complete answer, but a pointer

dms1981: @ewastempel - I think a condition that includes only accounts in our OU might be the best way to do this.

terraform/environments/core-logging/route53_resolver_logs.tf
```diff
@@ -15,7 +15,7 @@ data "aws_iam_policy_document" "route53-query-logging-policy" {
...
 
     principals {
       identifiers = ["route53.amazonaws.com"]
```
dms1981, line 21: Also this _might_ be incorrect, as we're asking to allow the route53resolver to do the logging, so `route53.amazonaws.com` might be the appropriate thing here.

SteveLinden: I'll amend it and post the update

dms1981: Apologies! I didn't spell it quite right - `route53resolver.amazonaws.com` should be what we're after, and I've amended my original comment to show this

dms1981 (approved):

PR #3104, opened 2023-02-10
Feature/route53 resolver logging

SteveLinden:
After consulting with AWS about the error we had connecting to VPCs it appears that the log file must be built in us-east-1.
This code changes the location (one on eu-west-2 has been removed) and builds it there. Once this is complete I will re-run the code to link to VPCs

### DIFF
--- a/terraform/environments/core-logging/route53_resolver_logs.tf
+++ b/terraform/environments/core-logging/route53_resolver_logs.tf
@@ -1,6 +1,16 @@
+provider "aws" {
+  alias  = "us-east-1"
+  region = "us-east-1"
+}
+
 resource "aws_cloudwatch_log_group" "modernisation-platform-r53-resolver-logs" {
   name              = "modernisation-platform-r53-resolver-logs"
   retention_in_days = 365
+}
+
+provider "aws" {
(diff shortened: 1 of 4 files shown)


terraform/environments/core-logging/route53_resolver_logs.tf
```diff
@@ -1,8 +1,18 @@
...
+  alias  = "us-east-1"
+  region = "us-east-1"
+}
+
 resource "aws_cloudwatch_log_group" "modernisation-platform-r53-resolver-logs" {
   name              = "modernisation-platform-r53-resolver-logs"
   retention_in_days = 365
 }
 
+provider "aws" {
```
davidkelliott, line 11: This provider is not used it can be removed

davidkelliott (changes requested):
This seems to be mixed in with changes that are already made?  Could you rebase from main or create a new branch from main and put your changes in there please?

terraform/environments/core-logging/route53_resolver_logs.tf
SteveLinden: Done and re-pushed

PR #3106, opened 2023-02-10
Move route53 log to Virginia

SteveLinden:
Move the route 53 log to us-east-1 (North Virginia) as suggested in the AWS documentation and suggested by AWS following a call


terraform/environments/core-logging/route53_resolver_logs.tf
```diff
@@ -1,3 +1,8 @@
+provider "aws" {
```
ewastempel, line 1: if needed, this should go to providers.tf

this change on its own will not have the desired effect. Note, you will need to reference it in your resources for it to have an effect

SteveLinden: Made amendments and pushed back up

dms1981:
Closing as not relevant given new fix branch

PR #3114, opened 2023-02-13
Refactor R53 resolver logs

dms1981:
* Remove CloudWatch log group and related resources from `core-logging`
* Add functionality to create CloudWatch log group to hold resolver logs on a per-VPC basis


terraform/modules/r53-resolver-logs/main.tf
```diff
@@ -1,10 +1,38 @@
+data "aws_caller_identity" "current" {}
```
ewastempel, line 1: perhaps it's worth creating a data.tf for data

terraform/modules/r53-resolver-logs/main.tf
```diff
@@ -1,10 +1,38 @@
...
+  name              = format("%s-r53-resolver-logs", var.vpc_id)
+  retention_in_days = 365
+}
+
+resource "aws_cloudwatch_log_resource_policy" "route53-query-logging-policy" {
+  policy_document = data.aws_iam_policy_document.route53-query-logging-policy.json
+  policy_name     = format("%s-r53-query-logging-policy", var.vpc_id)
+}
+
+data "aws_iam_policy_document" "route53-query-logging-policy" {
```
ewastempel, line 24: same, as my previous comment RE data.tf file

terraform/modules/r53-resolver-logs/main.tf
```diff
@@ -1,10 +1,38 @@
+data "aws_caller_identity" "current" {}
+
 resource "aws_route53_resolver_query_log_config" "main" {
-  name            = "dns_logs"
-  destination_arn = var.logging_destination_arn
+  name            = format("%s-r53-resolver-logs", var.vpc_id)
```
ewastempel, line 4: Perhaps `var.vpc_name` could be a more user friendly name? Although I think that would require more changes

dms1981: It would be friendlier, yeah. I think I can accommodate that. It should be a matter of supplying it through the `for_each` that I have in place.

ewastempel (approved):
nice one!

PR #3121, opened 2023-02-14
Add S3 bucket user for Jitbit

sobostion:


dms1981 (approved):
Approved - discussed rationale for IAM user rather than role assumption [here](https://mojdt.slack.com/archives/C032BQQHJE5/p1676451583936309?thread_ts=1675954567.723189&cid=C032BQQHJE5).

PR #3122, opened 2023-02-14
Initial Code for FQDN

markgov:
This is the Initial Code for the FQDN firewall rules


terraform/modules/firewall-policy/variables.tf
```diff
@@ -28,4 +28,8 @@ variable "rules" {
 variable "tags" {
   description = "A map of keys and values used to create resource metadata tags"
   type        = map(any)
+}
+
+variable "list_of_allowed_domains" {
```
dms1981, line 33: Could we use `fw_allowed_domains` to stay consistent with the naming of the other variables? We should probably have a `description` in the variable block too.

terraform/modules/firewall-policy/main.tf
```diff
@@ -49,6 +49,21 @@ resource "aws_networkfirewall_rule_group" "stateful" {
   }
 }
 
+resource "aws_networkfirewall_rule_group" "fqdn-stateful" {
+  capacity = 100
```
dms1981, line 53: The `capacity` value seems low to me. Should this be supplied with a variable, with a greater default value? The documented maximum for a rule group is `30000`, but the implied limit for a policy is `30000`, and that would be a combination of all the associated rule groups.

dms1981 (commented):
Also, consider the need to override the `$HOME_NET` variable - you'll want to read this: https://docs.aws.amazon.com/network-firewall/latest/developerguide/stateful-rule-groups-domain-names.html#stateful-rule-groups-domain-names-home-net

dms1981 (approved):

PR #3127, opened 2023-02-15
Remove start option for 6dg VPN tunnel startup

dms1981:
Based on the advice of AWS Support, I've removed the optional field that causes AWS to initiate tunnel establishment. From their investigation of the logs, the 6DG CGW is initiating connections.

From discussion with AWS Support they stated the following:
> Start: AWS initiates the IKE negotiation to bring the tunnel up. Only supported if your customer gateway is configured with an IP address. Supported if using IKEv2.

However, I'm pretty sure that we're using IKEv1 in other VPNs (eg, Nomis) so I don't know how trustworthy that statement is.


dms1981:
closing this as no longer appropriate

PR #3137, opened 2023-02-16
Code formatter pr creation

dms1981:
This now accurately creates a PR following a run of the code formatter.

I'm still not convinced that the step to check for changes is working successfully, so will dig into this separately.


markgov (approved):
LGTM

PR #3145, opened 2023-02-16
Complete configuration for new code formatter MVP

dms1981:
* Adjusted variable that supplies list of linters to use, as spaces are not supported in variables
* Added variable to exclude `.github` directory from yaml linter
* Adjusted `check for changes` to state if there are, or are not, changes detected


davidkelliott (approved):
👍

PR #3151, opened 2023-02-17
LAWS-3031: Add PagerDuty Alert for LAA in MP

vc13837:
Following https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/integrating-alarms-with-pagerduty-with-slack.html#3-create-a-pagerduty-service


ewastempel (approved):
LGTM

PR #3154, opened 2023-02-17
Reorganising default files - batch 4

ewastempel:
This PR is part of the #3090 implementation.

It updates default files with the latest templates files for:
- nomis
- oas
- oasys
- performance-hub
- refer-monitor
- remote-supervision
- tariff
- testing
- xhibit-portal

and makes sure that subnet_share.tf is not copied for core accounts (on top of the unrestricted accounts).

NOTE, these were tested locally with `terraform plan` for all of the accounts that the files have been updated for.


scripts/provision-environment-directories.sh
```diff
@@ -89,8 +89,8 @@ copy_templates() {
 
     if [ ${filename} != "platform_locals.tf" ] && [ ${filename} != "platform_providers.tf" ] && [ ${filename} != "platform_data.tf" ] && [[ "${filename}" !=  "member_"* ]]
     then
-      if [ ${filename} == "subnet_share.tf" ] && [ ${account_type} == "member-unrestricted" ]; then
```
davidkelliott, line 92: Alternatively you could test if the account_type != "member"

davidkelliott (approved):

scripts/provision-environment-directories.sh
ewastempel: thanks Dave, good to know

PR #3156, opened 2023-02-17
Update infrastructure support tag for environments

davidkelliott:
We want to start sending AWS health messages directly to the infrastructure support email addresses, rather than the current manual process of forwarding these on. Ensuring all accounts have these and they are up-to-date.


dms1981 (dismissed):

ewastempel (approved):
LGTM

PR #3157, opened 2023-02-17
Refactoring default files for sprinkler account

ewastempel:
This PR is part of the https://github.com/ministryofjustice/modernisation-platform/pull/3090 implementation.

It updates default files with the latest templates files for:
- sprinkler

NOTE,  it's been tested locally with terraform plan.
Unfortunately `sandbox` had to be hardcoded in locals and subnet share due to not being able to retrieve `sandbox` programmatically across all accounts (as it is now) and because of ssm patching relying on the environment value to be `development`


terraform/environments/sprinkler/platform_backend.tf
```diff
@@ -8,7 +8,7 @@ terraform {
     encrypt              = true
     key                  = "terraform.tfstate"
     region               = "eu-west-2"
-    workspace_key_prefix = "environments/accounts/sprinkler" # This will store the object as environments/sprinkler/${workspace}/terraform.tfstate
+    workspace_key_prefix = "environments/accounts/sprinkler" # This will store the object as environments/accounts/sprinkler/${workspace}/terraform.tfstate
```
davidkelliott, line 11: I notice this will change the location of the state file, but looking in the S3 bucket it's already in this location and last accessed in November, so Im not sure if it was changed accidentally to it's current location, but either way this should be fine.  You may want to delete the file in the old location to tidy up.

davidkelliott (approved):

terraform/environments/sprinkler/platform_backend.tf
ewastempel: oh yes, well noticed, weird, but this looks like the correct path, so yes, I will delete the other state and keep this one;
thanks Dave!

Issue #3069, opened 2023-02-06
Get AWS Health issues directly to users


davidkelliott:
## User Story

We often get health emails from AWS such as this:

Hello,

Athena engine auto-upgrade will begin in 60 days on March 9, 2023 for the EU-WEST-2 Region. Before upgrading a workgroup we perform a set of validation tests to confirm that its queries perform correctly and efficiently on engine version 3. Since our validation is performed on a best-effort basis, we recommend that all customers perform their own validation testing and follow the following practices to ensure all queries are executing as expected:

We then send these on to the user.

We need a better way of:
1. Getting these directly to the user
2. Get an overview of all of these on MP so that we can see which issues are not being actioned
3. A process to ensure these get addressed

## User Type(s)

MP engineer, MP user

## Value

Prevent SPOF with health notifications

## Definition of done

- [ ] users directly informed
- [ ] method of viewing across the platform implemented
- [ ] process in place for reviewing
- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

jakemulley:
A way to ensure users get these, as MP splits by account, is to take a user's `infrastructure-support` email and set the `OPERATIONS` contact using the [aws_account_alternate_contact](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/account_alternate_contact) resource.

PR #3165, opened 2023-02-20
Add alternative contacts for accounts

davidkelliott:
This will enable AWS health operations and security emails to go directly to the application support teams rather than to Jake -> MP -> App teams.

https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-update-contact.html

Tested on sprinkler development

Resolves https://github.com/ministryofjustice/modernisation-platform/issues/3069

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/account.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/account.tf
@@ -7,13 +7,3 @@ resource "aws_account_alternate_contact" "infrastructure_support_operations" {
   email_address = local.application_tags.infrastructure-support
   phone_number  = "+0000000000"
 }
-
-resource "aws_account_alternate_contact" "infrastructure_support_security" {
-
-  alternate_contact_type = "SECURITY"
-
-  name          = "Infrastructure Support"
-  title         = "Infrastructure Support Security"
-  email_address = local.application_tags.infrastructure-support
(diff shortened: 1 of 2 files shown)


terraform/environments/bootstrap/member-bootstrap/account.tf
```diff
@@ -0,0 +1,19 @@
...
+
+  name          = "Infrastructure Support"
+  title         = "Infrastructure Support Operations"
+  email_address = local.application_tags.infrastructure-support
+  phone_number  = "+0000000000"
+}
+
+resource "aws_account_alternate_contact" "infrastructure_support_security" {
+
+  alternate_contact_type = "SECURITY"
```
jakemulley: The `SECURITY` contact is already managed by [aws-root-account](https://github.com/ministryofjustice/aws-root-account/blob/main/management-account/terraform/alternate-contacts.tf#L10) for _all_ AWS accounts, so you won't need this here.

It is periodically run (manually) to capture any new accounts created 👍🏼

jakemulley (changes requested):

terraform/environments/bootstrap/member-bootstrap/account.tf
davidkelliott: Thanks! Removed :D

jakemulley (approved):

PR #3183, opened 2023-02-21
Amended the github-slug to point to the correct space

SteveLinden:
Corrected the github-slug to point to the correct place.


davidkelliott (approved):
👍

PR #3198, opened 2023-02-22
Granting athena query and dms replication task permissions on request from DE/DM

julialawrence:
* These permission requests are from the Data Engineering and Data Modelling teams who need permissions to run athena queries on an adhoc basis as well as save some named queries. 
* Data Modelling are requesting DMS permissions so they can execute adhoc replication tasks from their target databases.


davidkelliott:
Hi @julialawrence thanks for raising this, is there a reason why developers can't create their tasks via code https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dms_replication_task ? Thanks

PR #3201, opened 2023-02-22
Updating default files for PPUD

ewastempel:
This PR is part of the implementation of the https://github.com/ministryofjustice/modernisation-platform/issues/2521 ticket.

Updating default files to match the templates.

NOTE, I am not sure whether iam.tf should stay in this repo or should be migrated to the env one.


davidkelliott (approved):
IAM needs to stay, it contains things that the users don't have permissions to create via the environments repo.  We probably need to have a chat and get some documentation around this though, plus some pipelines to run this TF

PR #3215, opened 2023-02-23
added in permission to be able to run instances

markgov:
This PR add in permission for running instances this is to help with the ppud migration as AWS AMS does not seam to have permission to run the migrated instances


SteveLinden (approved):
Looks OK to me

PR #3266, opened 2023-03-01
Production and preproduction block

markgov:
This PR add a block to the ingress traffic to the prod and pre production CIDR range.


dms1981 (approved):
LGTM - but can we ensure we've notified people prior to implementing this?

PR #3312, opened 2023-03-07
remove r53resolver cloudwatch resource policies

dms1981:
Following on from the support case raised around r53 resolver logging issues, it seems prudent to remove these cloudwatch resource policies. They don't seem to add any necessary, extra, security to our logging, they're not visible through the AWS Console, and we're limited to 10 policies per account - this PR frees up 8 of them.


davidkelliott (approved):
👍 seems sensible, thanks

PR #3318, opened 2023-03-07
Updated code formatting action

dms1981:
Realised that the `check for changes` step was missing a commit message.


davidkelliott (dismissed):
👍 if it works it might be worth removing the 2nd commit as someone may come along later and remove this one thinking it's not needed

davidkelliott (approved):

PR #3325, opened 2023-03-08
Documentation/fqdn firewall documentation

markgov:
Documentation for new FQDN firewall rules


markgov:
while i was updating the documentation i have found that i have a missing output so i have added here as part of the update

SteveLinden (approved):

PR #3341, opened 2023-03-08
reviewed docs for Daniel the Manual Spaniel

dms1981:


davidkelliott (changes requested):
This will need to be updated here as well - https://github.com/ministryofjustice/modernisation-platform/blob/main/source/index.html.md.erb#L109

dms1981:
> This will need to be updated here as well - https://github.com/ministryofjustice/modernisation-platform/blob/main/source/index.html.md.erb#L109

Good catch. Thanks!

SteveLinden (approved):

PR #3346, opened 2023-03-09
Update rule options generation

dms1981:
When this code was previously written it was compliant with the AWS API. Since then, the way that the firewall rule option and settings are passed in has changed. This means that with our current code, every change has repeating iterations of the following:
```
                  - rule_option {
                      - keyword  = "sid" -> null
                      - settings = [
                          - "1",
                        ] -> null
                    }
                  + rule_option {
                      + keyword  = "sid:1"
                      + settings = []
                    }
```
This change will correct that by ensuring that the `keyword` and `settings` are separated.


ewastempel (approved):
nice!

PR #3355, opened 2023-03-09
Initial changes for us-east-1 policy and role.

ep-93:


davidkelliott (changes requested):
Policy looks good but I've messaged you about the location and pattern of implementation to keep things consistent with the other member infrastructure access role

terraform/environments/bootstrap/member-bootstrap/providers.tf
```diff
@@ -11,3 +11,11 @@ provider "aws" {
   alias  = "modernisation-platform"
   region = "eu-west-2"
 }
+# AWS provider (workspace): the workspace account. Required for assuming a role into an account for bootstrapping
```
davidkelliott, line 14: This provider isn't needed, the boot strapping won't be creating any resources in us-east-1, also I think this would create a dependancy loop if you used this provider as you are creating a role then trying to assume it in the same terraform

terraform/environments/bootstrap/member-bootstrap/iam.tf
```diff
@@ -229,4 +229,48 @@ resource "aws_iam_role_policy_attachment" "testing_member_infrastructure_access_
   count      = terraform.workspace == "testing-test" ? 1 : 0
   role       = aws_iam_role.testing_member_infrastructure_access_role[0].id
   policy_arn = aws_iam_policy.member-access[0].arn
-}
\ No newline at end of file
+}
+
```
davidkelliott, line 233:
```suggestion
```

davidkelliott (changes requested):

terraform/environments/bootstrap/member-bootstrap/providers.tf
ep-93: @davidkelliott Yeah I was looking at the output and realised that now, thanks

davidkelliott (approved):
LGTM 👍

PR #3367, opened 2023-03-10
Adding new collaborators

markgov:
Processing request from the ask channel for more collaborators for the ppud environments

### DIFF
--- a/collaborators.json
+++ b/collaborators.json
@@ -323,6 +323,14 @@
         {
           "account-name": "ppud-development",
           "access": "developer"
+        },
+        {
+          "account-name": "ppud-preproduction",
+          "access": "migration"
+        },
+        {
+          "account-name": "ppud-production",
+          "access": "migration"
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


collaborators.json
```diff
@@ -323,6 +323,14 @@
         {
           "account-name": "ppud-development",
           "access": "developer"
+        },
+        {
+          "account-name": "ppud-preproduction",
+          "access": "migration"
```
davidkelliott, line 329: This should be developer access

collaborators.json
```diff
@@ -323,6 +323,14 @@
...
+        },
+        {
+          "account-name": "ppud-production",
+          "access": "migration"
```
davidkelliott, line 333: This should be developer access

davidkelliott (changes requested):

collaborators.json
markgov: role updated

collaborators.json
markgov: role updated

davidkelliott (approved):

PR #3371, opened 2023-03-10
Allow traffic from CP to MP platforms developement

davidkelliott:
Allow traffic from MP to the platforms development subnets, this is to allow testing of traffic latency between CP and MP.


markgov (approved):
LGTM

PR #3445, opened 2023-03-16
Name tweak to delius-core ecr repo names for consistency

adeweetman-al:
Small name change to ensure consistency with other repos.


markgov (approved):
LGTM

PR #3462, opened 2023-03-17
initial changes for laa-ccms-infra-azure-ad-sso

dms1981:


dms1981:
> Seems to list all accounts not just developer. Is this what they want? Thought it was just development

You're right! Let me sort that.

ep-93 (approved):

PR #3475, opened 2023-03-20
Adding servicediscovery to the cicd memebr permissions

ewastempel:
Adding permissions to the cicd user as per a user's request [here](https://mojdt.slack.com/archives/C01A7QK5VM1/p1679316872741289).


adeweetman-al:
Shouldn't these new permissions be associated to the `MemberInfrastructureAccess` role as they're needed when we run the usual terraform plan/apply workflows?

Issue #2943, opened 2023-01-10
Update environments repo provider files so that users don't need to modify them to run local plans


davidkelliott:
## User Story

Do this to all provider files in the environments repo - https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/running-terraform-plan-locally.html#running-local-plans-without-file-modifications

conditional - there is an example in the documentation. check that this works for all users, SSO and collaborators.

## Definition of done

- [x] readme has been updated
- [x] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

davidkelliott:
Note we discussed this as something to do when we refactor to use terragrunt to save updating everything in multiple places

davidkelliott:
Template changes in sprinkler here - https://github.com/ministryofjustice/modernisation-platform-environments/pull/1350

seanprivett:
@davidkelliott to review after assessing new roles that have been created since ticket drafted

PR #3533, opened 2023-03-28
Update provider templates for better local plans

davidkelliott:
To run plans locally you used to have to comment and uncomment the provider files.  This change updates the templates so that the user you are using is detected and the appropriate provider settings applied.

Also updating documentation around this.

Issue https://github.com/ministryofjustice/modernisation-platform/issues/2943


ewastempel (approved):
nice work

PR #3549, opened 2023-03-30
Refactor provisioning script

davidkelliott:
There are too many differences between the terraform files in the modernisation platform repo and the environments repo to keep using the same templates.  This splits the templates into two folders, one for each repo.  The scripts have been updated and simplified to account for this.  Now we can make changes to one set of files without impacting the other.


ewastempel (approved):
Looks good, tested locally in mp and env repo using `apex` as an example account to be generated.

PR #3560, opened 2023-04-03
Enabling security-audit and read-only permission sets

julialawrence:
- Adding security-audit and read-only permission options, updating rego.
- Switching permission_set_arn for instance-management.
- Documentation update

SecurityAudit and ReadOnly permission sets are already defined in the root account, so this is code to enable these permissions for MP and MP users.


ewastempel (approved):
LGTM

markgov (approved):
LGTM

PR #3562, opened 2023-04-03
Further provision script simplification

davidkelliott:
This removes the need for separate sed commands converting the backends.
Also removing the original template files which are no longer needed.


ewastempel (approved):
Looks good, also tested locally using apex account as an example. TF plans out clean.

PR #3563, opened 2023-04-03
Add script to replace platform files

davidkelliott:
This is a helper script to quickly replace all the platform_files in the environments repo to match the environment repo templates.

Note if you keep your environments folder in a different location you will need to update the environments_folder location.


ewastempel (approved):
nice one

PR #3577, opened 2023-04-05
Fix template location on workspace script 

davidkelliott:
The template location was updated but not changed here, fixing this now and updating some broken links in documentation


markgov (approved):
LGTM

PR #3583, opened 2023-04-05
Rename main repo files to remove platform_ prefix

davidkelliott:
This prefix is used in the environment repo so that providers know that these files are managed by the MP team and are programatically overwritten.  Some of the platform_ files in the mp repo are different though from the environment repo platform_ files.  Because this is such a unique name it encourages the thought that they are the same.  This PR is reverting the MP files back to standard terraform naming conventions, so that it is clear that these files are different, and should not be replaced with the ones from the environment repository templates.


ewastempel (commented):
looks good :)

ewastempel (approved):

PR #3619, opened 2023-04-12
New section to archive the modernisation-platform-terraform-ecs repo

SteveLinden:
Added code to archive but we want to delete it instead so code will be amended. It automatically archives when you delete.

Had to remove a line of code from teams.tf (module.terraform-module-ecs.repository.name) as it was failing when it hit this.

### DIFF
--- a/terraform/github/repositories.tf
+++ b/terraform/github/repositories.tf
@@ -150,6 +150,7 @@ module "terraform-module-ecs" {
   name        = "modernisation-platform-terraform-ecs"
   type        = "module"
   description = "Module for creating ECS cluster (Linux/Windows) solely for EC2 launch type"
+  archived = true
   topics = [
     "aws",
     "ecs",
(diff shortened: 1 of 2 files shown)


terraform/github/repositories.tf
```diff
@@ -395,3 +395,8 @@ module "modernisation-platform-incident-response" {
     "helm"
   ]
 }
+
+resource "github_repository" "modernisation-platform-terraform-ecs" {
```
ewastempel, line 399: This block of code will add a new repository and archive it. However, such resource already exists and needs updating. Do a search in code for the repo name and you will find it.

ewastempel (changes requested):

davidkelliott:
Just noticed, we have a setting here set to archive repos when they are removed from terraform, see code here - 
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/github/modules/repository/main.tf#L24

And what the setting does here - 

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository#archive_on_destroy

So I think you just need to remove the repo from code and it will automatically be archived.

Issue #3452, opened 2023-03-16
Switch to shared workflows


davidkelliott:
## User Story

Nomis have been trialing reusable workflows - https://github.com/ministryofjustice/modernisation-platform-environments/pull/1482, test these new workflows on MP applications and if there are no issues switch all applications to these.

## User Type(s)

MP users

## Value

Cleans up workflow code, dries it out and adds new features

## Definition of done

- [x] switch MP apps to new workflows and test
- [x] switch user apps to new workflows
- [x] let users know about the changes
- [x] user docs have been updated
- [x] another team member has reviewed

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

ewastempel:
This has been implemented and documented with the PR links attached.
The update channel received a post on the changes.

This is done now.

PR #3629, opened 2023-04-12
Updating workflow template to include the reusable pipelines

ewastempel:
This PR is part of the https://github.com/ministryofjustice/modernisation-platform/issues/3452 ticket.

This is to update the workflow-template.yml to use reusable strategy and tf plan/apply pipelines.


.github/workflows/templates/workflow-template.yml
```diff
@@ -22,238 +24,43 @@ on:
...
+      fail-fast: false
+      matrix: ${{ fromJson(needs.strategy.outputs.matrix) }}
+    uses: ./.github/workflows/reusable_terraform_plan_apply.yml
+    with:
+      application: "${{ github.workflow }}"
+      environment: "${{ matrix.target }}"
+      action: "${{ matrix.action }}"
+    secrets:
+      modernisation_platform_environments: "${{ secrets.MODERNISATION_PLATFORM_ENVIRONMENTS }}"
+      pipeline_github_token: "${{ secrets.DSO_GITHUB_AUTOMATION_PAT }}"
```
davidkelliott, line 54: What is this used for? Should we be using our token here?

ewastempel: Good spot, this is for posting comments in a PR (tf plan/apply output) I believe (that's what we used it for in DSO in the past)

ewastempel: this PAT is global to organisation, I would be tempted to re-use it (once AL goes away, it would be nice to keep it), but if we have one as well then it makes sense to use ours

ewastempel: see our new shiny secret please :)

.github/workflows/templates/workflow-template.yml
```diff
@@ -22,238 +24,43 @@ on:
...
+      fail-fast: false
+      matrix: ${{ fromJson(needs.strategy.outputs.matrix) }}
+    uses: ./.github/workflows/reusable_terraform_plan_apply.yml
+    with:
+      application: "${{ github.workflow }}"
+      environment: "${{ matrix.target }}"
+      action: "${{ matrix.action }}"
+    secrets:
+      modernisation_platform_environments: "${{ secrets.MODERNISATION_PLATFORM_ENVIRONMENTS }}"
+      pipeline_github_token: "${{ secrets.MODERNISATION_PLATFORM_CI_USER_ENVIRONMENTS_REPO_PAT }}"
```
ewastempel, line 54: this token is updated now to MODERNISATION_PLATFORM_CI_USER_ENVIRONMENTS_REPO_PAT

davidkelliott (approved):

PR #3654, opened 2023-04-14
Allow traffic from I2N to HMPPS prod and preprod

davidkelliott:
This is required for delius iaps, see request here - https://mojdt.slack.com/archives/C01A7QK5VM1/p1681394593957699

### DIFF
--- a/terraform/environments/core-network-services/production_rules.json
+++ b/terraform/environments/core-network-services/production_rules.json
@@ -145,5 +145,19 @@
     "destination_ip": "10.27.64.0/21",
     "destination_port": "ANY",
     "protocol": "TCP"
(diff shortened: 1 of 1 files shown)


SteveLinden (commented):
Are you sure about the preprod IP address? .0.0/21

davidkelliott:
> Are you sure about the preprod IP address? .0.0/21

@SteveLinden yes that's correct
https://github.com/ministryofjustice/modernisation-platform/blob/main/environments-networks/hmpps-preproduction.json#L5
https://github.com/ministryofjustice/modernisation-platform/blob/main/cidr-allocation.md?plain=1#L67

PR #3663, opened 2023-04-17
fix up global protect rules for hmpps VPCs

drobinson-moj:
Fixed / aligned global protect firewall rules for hmpps VPCs
- add access to development for testing
- add access to preprod
- fixed the production rules which were typoed.  It is safe to change as the typoed 10.40/18 subnet is already covered by later rules in the same file
- updated the rule name in hmpps-test

### DIFF
--- a/terraform/environments/core-network-services/production_rules.json
+++ b/terraform/environments/core-network-services/production_rules.json
@@ -1,14 +1,14 @@
 {
   "gp_to_mp_nomis_production_http": {
     "action": "PASS",
-    "source_ip": "10.40.0.0/18",
+    "source_ip": "10.184.0.0/16",
     "destination_ip": "10.27.8.0/21",
     "destination_port": "80",
     "protocol": "TCP"
   },
   "gp_to_mp_nomis_production_https": {
     "action": "PASS",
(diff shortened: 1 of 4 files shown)


terraform/environments/core-network-services/production_rules.json
```diff
@@ -1,14 +1,14 @@
 {
   "gp_to_mp_nomis_production_http": {
```
ewastempel, line 2: this one could be renamed to hmpps as well

terraform/environments/core-network-services/production_rules.json
```diff
@@ -1,14 +1,14 @@
...
     "action": "PASS",
-    "source_ip": "10.40.0.0/18",
+    "source_ip": "10.184.0.0/16",
     "destination_ip": "10.27.8.0/21",
     "destination_port": "80",
     "protocol": "TCP"
   },
   "gp_to_mp_nomis_production_https": {
```
ewastempel, line 9: and this one as well could rename nomis to hmpps

ewastempel (approved):
looks good, thanks for fixing it! useful as always :)

PR #3668, opened 2023-04-18
Fixing creation/update of environments settings

ewastempel:
Recently our newly created accounts do not have workflow protection, which means that apply gets deployed automatically if the plan passes OK. This can cause a lot of problems, e.g. unwanted destruction of resources in a test environment if a dev environment passes.

This PR is to fix the problem. It's been already tested and applied on `cooker` and on the `hmpps-intelligence-management`.

### DIFF
--- a/environments/cooker.json
+++ b/environments/cooker.json
@@ -3,7 +3,12 @@
   "environments": [
     {
       "name": "development",
-      "access": []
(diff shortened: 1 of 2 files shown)


davidkelliott (approved):
👍

PR #3700, opened 2023-04-24
Move logging out of firewall policy module

dms1981:
At present the `firewall-policy` module introduces a cyclic dependency when it's used with a new firewall.
* A new firewall resource requires a policy
* The `firewall-policy` module requires a firewall ARN for its logging configuration

This PR separates out the logging into its own small module, and includes a random prefix to both ensure the new log groups are unique, and to prevent a clash with the existing log groups before they are destroyed.


davidkelliott (approved):
👍 :shipit:

PR #3701, opened 2023-04-24
Feature/cloudwatch kms

markgov:
Added new kms allow policy for cloudwatch at the request of users following the bellow comment in the ask channel
Afternoon team, would it be possible to give cloudwatch access to the business unit general CMK. so we can have encrypted SNS topics.


dms1981:
More context for this PR can also be found here: https://repost.aws/knowledge-center/cloudwatch-receive-sns-for-alarm-trigger

PR #3722, opened 2023-04-25
Refactor to names not CIDR ranges

davidkelliott:
This moves the definition of the CIDR ranges in to one file, and then the firewall files can use words rather than ranges to define the firewall rules.

This helps in a number of ways -

1. Easier to read and create
2. If CIDR ranges change we only update them in one place
3. We automatically get all of the MP ranges for the general subnet sets so no need to look them up.

This PR is an initial draft to demo this.  Other firewall files will be added if it is agreed to move forward.

Note the plan shows no changes.


terraform/environments/core-network-services/cidr-ranges.tf
```diff
@@ -0,0 +1,57 @@
...
+    analytical-platform-airflow-dev  = "10.200.0.0/16"
+    analytical-platform-airflow-prod = "10.201.0.0/16"
+    global-protect                   = "10.184.0.0/16"
+    psn                              = "51.0.0.0/8"
+    psn-ppud                         = "51.247.2.115/32"
+    moj-core-azure-1                 = "10.50.25.0/27"
+    moj-core-azure-2                 = "10.50.26.0/24"
+    i2n                              = "10.110.0.0/16"
+
+    # azure cidr ranges
```
ewastempel, line 22: all of these azure cidr ranges belong to hmpps, so perhaps worth mentioning it

terraform/environments/core-network-services/firewall-rules/preproduction_rules.json
```diff
@@ -1,134 +1,134 @@
 {
   "laa_stage_to_mp_laa_preproduction_http": {
     "action": "PASS",
-    "source_ip": "10.204.0.0/20",
-    "destination_ip": "10.27.72.0/21",
+    "source_ip": "${laa-lz-staging}",
+    "destination_ip": "${laa-preproduction}",
```
ewastempel, line 5: can't see it defined in the cidr ranges

terraform/environments/core-network-services/firewall-rules/preproduction_rules.json
```diff
@@ -1,134 +1,134 @@
...
     "destination_port": "80",
     "protocol": "TCP"
   },
   "gp_to_hmpps_preproduction_http": {
     "action": "PASS",
-    "source_ip": "10.184.0.0/16",
-    "destination_ip": "10.27.0.0/21",
+    "source_ip": "${global-protect}",
+    "destination_ip": "${hmpps-preproduction}",
```
ewastempel, line 12: can't see it defined in the cidr ranges

terraform/environments/core-network-services/firewall-rules/development_rules.json
```diff
@@ -1,50 +1,50 @@
 {
   "gp_to_hmpps_development_http": {
     "action": "PASS",
-    "source_ip": "10.184.0.0/16",
-    "destination_ip": "10.26.24.0/21",
+    "source_ip": "${global-protect}",
+    "destination_ip": "${hmpps-development}",
```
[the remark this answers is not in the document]
davidkelliott: These are defined here - https://github.com/ministryofjustice/modernisation-platform/blob/a4fd7fefa73992c4e8c7beaaea41136f0da9037a/terraform/environments/core-network-services/cidr-ranges.tf#L7 these are our MP ranges, and are generated from the environments-networks files so they are not repeated, also they won’t need to be updated if we add more business units.
They always follow the same format eg laa-development businessunit-environment so it should be fairly easy to work with (or check the environment-network files if not sure)

ewastempel (approved):
Looks great, well done! 😍

PR #3726, opened 2023-04-26
Fix broken KMS policy

dms1981:
I missed an `=` sign when reviewing a prior PR; this branch introduces the missing operator


davidkelliott (commented):
Is this the correct syntax, is it not something like:

```
principals {
      type        = "Service"
      identifiers = "cloudwatch.amazonaws.com"
    }
```

davidkelliott (approved):

PR #3729, opened 2023-04-26
Update incorrect ranges

davidkelliott:
10.40.0.0/16 this will be changed to noms-live-vnet (10.40.0.0/18)
10.27.8.0/16 this will be changed to hmpps-production (10.27.8.0/21)

The other IPs stay the same just converting to using the locals.

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/production_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/production_rules.json
@@ -15,7 +15,7 @@
   },
   "azure_noms_to_mp_nomis_production_db": {
     "action": "PASS",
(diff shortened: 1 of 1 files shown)


terraform/environments/core-network-services/firewall-rules/production_rules.json
```diff
@@ -15,7 +15,7 @@
   },
   "azure_noms_to_mp_nomis_production_db": {
     "action": "PASS",
-    "source_ip": "10.40.0.0/16",
```
ewastempel, line 18: just checked and NOMS-Live CIDR range is 10.40.0.0/18

ewastempel (approved):

Issue #3547, opened 2023-03-30
Implement in-line firewall endpoints for scanning of egress traffic


dms1981:
## User Story

As a Modernisation Platform Engineer 
I want to pass traffic leaving the Modernisation Platform through our Network Firewall
So that we can permit HTTP traffic only to preselected destinations

## User Type(s)

Modernisation Platform Customer

## Value

We previously discussed consolidating our egress VPCs in this issue: [Consolidate Egress VPCs](https://github.com/ministryofjustice/modernisation-platform/issues/3000). However, after some discussion in the team it was felt that preserving our separation of traffic was of sufficient importance that we should look into an alternative way to scan our egress traffic.

This means that our approach needs to change to use in-line firewall endpoints in our egress VPCs. We can accommodate this by provisioning a firewall endpoint in each availability zone in the egress VPCs. We will route traffic bound for the internet from the transit gateway endpoint subnets via the new firewall endpoints. We will then route traffic bound for the internet from the firewall endpoint subnets via the existing NAT Gateways. We will then route traffic bound for the Modernisation Platform back from the NAT Gateway subnets via the new firewall endpoints, and then from the firewall endpoint subnets via the transit gateway.

## Definition of done

- [x] firewall endpoints provisioned in non_live egress VPC
- [x] firewall prepared to allow traffic out to HTTPS without further inspection
- [x] firewall prepared to allow traffic out to HTTP *with* FQDN inspection
- [x] routing changed to flow through firewall endpoints in non_live egress VPC
- [x] traffic tested successfully through firewall to HTTPS internet destination
- [x] traffic tested successfully through firewall to permitted HTTP internet destination
- [x] traffic tested unsuccessfully through firewall to non-permitted HTTP internet destination
- [x] firewall endpoints provisioned in live egress VPC
- [x] routing changed to flow through firewall endpoints in live egress VPC
- [x] diagrams updated
- [x] documentation updated
- [x] tests updated

## Reference

* [How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)
* [core-networking terraform](https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/core-network-services/vpc.tf)

SimonPPledger:
@davidkelliott can we discuss how we do this - but, ideally, we bring it into next sprint

dms1981:
The following PRs are relevant to this story:
[...]

PR #3762, opened 2023-05-02
Route traffic in egress VPCs through inline firewalls

dms1981:
This PR is the final piece of the puzzle for #3547 .
It introduces the following changes:

* `aws_route_table.public` in the `vpc-hub` module becomes conditional
* when `inline_inspection = true`, public route tables are created per-availability-zone so that traffic can be routed via firewall endpoints
* new outputs are created to map route tables IDs to availability zones
* some existing outputs are expanded to account for the possibility of one or more route tables

* An output in `core-network-services` is temporarily commented out and will be revisited in a fix branch as its current output can't be computed (`terraform console` shows `known after apply` for `module.vpc_hub[key].public_igw_route`)

* Because of the inherent dependency of the inspection firewalls on resources created in the `vpc-hub` module, routes that then make reference to the `aws_networkfirewall_firewall` resources are created in `core-network-services`, although this could be discussed further.
* In brief, these routes cover `transit_gateway > inspection_endpoint`, `inspection_endpoint > nat_gateway`, `nat_gateway > inspection_endpoint`, and `inspection_endpoint > nat_gateway` allowing traffic to flow through the egress VPC via the firewall endpoints to the internet, and back from the internet via the firewall endpoints to the transit gateway.

### DIFF
--- a/terraform/modules/vpc-hub/main.tf
+++ b/terraform/modules/vpc-hub/main.tf
@@ -310,7 +310,14 @@ resource "aws_route_table_association" "public-inspection" {
 }
 
 # Public route
-resource "aws_route" "public-internet-gateway" {
+resource "aws_route" "public-internet-gateway-transit" {
+  count                  = var.gateway == "transit" ? 1 : 0
+  destination_cidr_block = "0.0.0.0/0"
+  gateway_id             = aws_internet_gateway.default.id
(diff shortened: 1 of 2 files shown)


dms1981:
Superseded by https://github.com/ministryofjustice/modernisation-platform/pull/3834

PR #3777, opened 2023-05-03
Amended private-public to shared-private on platform_data.tf

SteveLinden:
Only place this code appears on modernisation-platform so I have amended this. It will need to be merged around the same time as the change for modernisation-platform-environments.


ep-93 (approved):

davidkelliott (approved):

ewastempel (approved):
nice one!

PR #3819, opened 2023-05-09
Inline inspection module

dms1981:
This PR bundles up previous work to create a module which can be used to create an inline inspection VPC.
The VPC contains the following::
* AWS VPC with three types of subnet; `transit-gateway`, `inspection`, and `private`
* AWS Network Firewall with endpoints spread across the `inspection` subnets
* AWS Route tables and routes to flow traffic out through the firewall endpoints and back in via the transit gateway
* AWS NAT gateways with Elastic IP addresses
* AWS Network ACLs
* Outputs to be used in the creation of tests


dms1981:
Relevant to #3547 and our discussion on a more strategic approach to solving the need, rather than implementing tactical solutions.

ep-93 (approved):

PR #3834, opened 2023-05-10
Create two inline_inspection VPCs

dms1981:
This PR creates two VPCs from the `vpc-inspection` module, and attaches them to the MP transit gateway.
A future PR will be needed to route traffic through these new VPCs and remove the old VPCs, as well as to introduce tests to ensure that VPCs remain functional.


davidkelliott (approved):
Looks good and thanks for the cidr range explanation 👍

PR #3837, opened 2023-05-10
Change MRLA route53

SteveLinden:
Code has been planned and will work with production OK.

### DIFF
--- a/terraform/environments/core-network-services/route53.tf
+++ b/terraform/environments/core-network-services/route53.tf
@@ -6,7 +6,7 @@ locals {
     equip    = "equip.service.justice.gov.uk",
     ccms-ebs = "ccms-ebs.service.justice.gov.uk",
     mojfin   = "laa-finance-data.service.justice.gov.uk",
-    mlra     = "mlra.service.justice.gov.uk"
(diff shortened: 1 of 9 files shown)


terraform/environments/core-network-services/route53.tf
```diff
@@ -102,6 +102,22 @@ resource "aws_route53_record" "bichard7" {
   ]
 }
 
+# MRLA/MAAT location
+resource "aws_route53_record" "mrla" {
```
davidkelliott, line 106: This record isn't needed, this is creating a name server record which would send all DNS queries to the ns records detailed in this record.  We don't need to do this here as we want to have control of the domain.  What we do need to do is once this hosted zone is created, get the NS records for this new zone and give them to ops engineering so that they can create a record like this one to forward queries on to us

davidkelliott (changes requested):
The route 53 record isn't needed and the extra files need to be removed

SteveLinden:
Change was in another request (https://github.com/ministryofjustice/modernisation-platform/pull/3833) which has been merged. Closing this PR

PR #3839, opened 2023-05-10
Move inspection VPC TGW attachment into module

dms1981:
* Move the `aws_ec2_transit_gateway_vpc_attachment` resource into the `vpc-inspection` module
* Update outputs to remove use of legacy splat syntax
* Remove unused output from `vpc-inspection` module
* Added version constraints to align with good practice


ep-93 (approved):

ep-93:
Since this is blocking other tickets, the analysis can be dealt with after.

PR #3842, opened 2023-05-11
Fix policy name

davidkelliott:
The sso permission set provisioning was failing as the policy name didn't match.  Changing to match the name defined here:

https://github.com/ministryofjustice/aws-root-account/blob/efd728ed376d8ae244c4fb2b1cc99005d3033cc3/management-account/terraform/sso-admin-permission-sets.tf#LL304C13-L304C39


markgov (approved):
LGTM

PR #3848, opened 2023-05-11
Associate inspection VPCs with TGW route tables

dms1981:
In preparation for migration, the new inspection VPCs need to be associated with the correct TGW route tables.

As the route tables are defined in `core-network-services`, I think this is the most appropriate place to put the route table association; I could feed it into the module (which is probably the better approach) but starts to blur the line on where the associations are done and for the sake of comprehensibility I'm going to leave the association in its current location.

* Add output of TGW attachment ID to `vpc-inspection` module
* Consume output in `core-network-services`


markgov (approved):
LGTM

PR #3849, opened 2023-05-11
Route traffic to internet through new inline-inspection VPCs

dms1981:
* Amend default routes for `live_data` and `non_live_data` route tables to flow traffic out through new inline-inspection VPCs


markgov (approved):
LGTM

davidkelliott (approved):

PR #3850, opened 2023-05-11
Disconnect old egress VPCs

dms1981:
As the old egress VPCs are no longer required, the first step in removing them is disconnecting them from the Modernisation Platform transit gateway.

This PR also updates the `Name` tags of the attachments to the new egress VPCs so that they match the previous syntax, and are monitored by our CloudWatch alarms.


ep-93:
Keeping them vars tidy. LGTM.

ep-93 (approved):

PR #3884, opened 2023-05-17
Tests for new inspection VPCs

dms1981:
This PR recreates the existing tests for the old egress VPCs and rewrites them for the new inline-inspection egress VPCs.


davidkelliott (approved):
👍

PR #3917, opened 2023-05-22
Application zone for tipstaff.service.justice.gov.uk

dms1981:


terraform/environments/core-network-services/route53.tf
```diff
@@ -5,8 +5,9 @@ locals {
   application-zones = {
     equip    = "equip.service.justice.gov.uk",
     ccms-ebs = "ccms-ebs.service.justice.gov.uk",
+    mlra     = "maat-libra-administration-tool.service.justice.gov.uk",
     mojfin   = "laa-finance-data.service.justice.gov.uk",
-    mlra     = "maat-libra-administration-tool.service.justice.gov.uk"
+    tipstaff = "tipstaff.service.justice.gov.uk"
```
ewastempel, line 10: could we use more meaningful name here?

davidkelliott (approved):
See discussions on slack regarding name, agreeing to go ahead as is.

PR #3933, opened 2023-05-23
Add apigateway permissions to the member pipeline

davidkelliott:
This will allow MP users to create api gateways, following a user request in the ask channel here - https://mojdt.slack.com/archives/C01A7QK5VM1/p1684838327099879


ewastempel (approved):
LGTM

PR #3938, opened 2023-05-23
Ignore rotation warnings

davidkelliott:
We either can't, don't want to or currently do rotate these via a different mechanism.


ewastempel (approved):
LGTM

Issue #3290, opened 2023-03-03
Firewall monitoring


davidkelliott:
## User Story

As an MP engineer
I need to know when the firewall is unexpectedly dropping traffic
So that I can ensure rules are correct and that we are not being attacked.

## User Type(s)
MP engineers

## Value

Know when our firewall is blocking it shouldn't or if we are under attack.

### Proposal

If we create a cloudwatch alarm with anomaly detection on traffic dropped from the firewall, if we get a sudden spike in dropped traffic we'll know either something has started blocking unexpectedly or we are under attack.

## Definition of done

- [ ] firewall monitoring in place
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

ep-93:
Need to write a cloudwatch alarm for DroppedPackets

 https://docs.aws.amazon.com/network-firewall/latest/developerguide/monitoring-cloudwatch.html

ep-93:
Waiting on - https://github.com/ministryofjustice/modernisation-platform/pull/4046

Once merged, completed. Please review the alarm though.

PR #3972, opened 2023-05-26
Firewall Alarm Dropped Packets

ep-93:
Added firewall cloudwatch alarm with no sns topic attached. Already applied, and works.

Related to this card. https://github.com/ministryofjustice/modernisation-platform/issues/3290


markgov (approved):
LGTM

PR #4087, opened 2023-06-01
network approach

markgov:
Updating the network approach to include a mention of network firewalls

### DIFF
--- a/source/concepts/networking/networking-approach.html.md.erb
+++ b/source/concepts/networking/networking-approach.html.md.erb
@@ -1,7 +1,7 @@
 ---
 owner_slack: "#modernisation-platform"
 title: Networking Approach
-last_reviewed_on: 2022-11-24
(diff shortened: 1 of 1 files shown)


source/concepts/networking/networking-approach.html.md.erb
```diff
@@ -76,6 +76,10 @@ NACLs allow traffic in and out of the Modernisation Platform (North/South), but
 
 Traffic within a VPC is not restricted.
 
+#### Network Firewall
+
+Network Fireall trafficis controlled via code by amending the following the documentation [Network Firewall](network-firewall.html)
```
dms1981, line 81: How about...
```AWS [Network Firewalls](network-firewall.html) provide additional controls to traffic entering and exiting the Modernisation Platform.```

dms1981 (approved):

PR #4091, opened 2023-06-01
add secret manager perms for instance mgmt role

sobostion:


davidkelliott (approved):
Approved taking the view that if someone is trusted with OS level access they will need access to passwords in order to do their job.  The ability to change passwords remains with the developer role.

Issue #3876, opened 2023-05-17
Use AWS LicenseManager to create grant across OU


dms1981:
## User Story

As a Modernisation Platform Engineer
I want to allow all accounts in our OU to make use of a marketplace license agreement
So that our customers can use marketplace AMIs in their accounts

## User Type(s)

Modernisation Platform Engineer
Modernisation Platform customer

## Value

In [this slack conversation](https://mojdt.slack.com/archives/C01A7QK5VM1/p1684254308541869) the Nomis team requested that we help them with an AWS Marketplace license. We accepted the license in our main `Modernisation Platform` account but on going through the License Manager pane could not see a way to share the license with our OU, only individual accounts

### Hypothesis 
If we use AWS License Manager in our parent account to share licenses
Then we will prevent us having to accept a license in multiple accounts, making it easier for us to administer licenses and understand what licenses we have accepted

## Definition of done

- [ ] license shared across Modernisation Platform OU
- [ ] license confirmed as available in a subordinate account
- [ ] Nomis team informed

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)
[Sharing subscriptions in an organization](https://docs.aws.amazon.com/marketplace/latest/buyerguide/organizations-sharing.html)

markgov:
Looking at the following documentation 
https://docs.aws.amazon.com/data-exchange/latest/userguide/organizations-sharing.html#view-all-licenses
is is possible to share the licences with other accounts but it will need root account access to do this function

markgov:
i think i have an idea how this works but have reached out to craig on the aws team to see if there are any sessions that can be setup around this subject to explain it better.

dms1981:
Moved this back to in-progress while we resolve the need to implement service roles for License Manager

dms1981:
Further details can be seen here: https://github.com/ministryofjustice/modernisation-platform/issues/3511

In short, this one isn't sensibly possible as we can't control the creation of the service role assumed by licensemanager as it is also created through an organization-level marketplace integration setting.

As such we've chose to manually accept licenses in the accounts which require them, pending a better solution.

PR #4093, opened 2023-06-01
Grant RHEL-6 ELS license to all member accounts

dms1981:
As part of #3876 , this PR creates a module to both grant & accept an AWS LicenseManager grant based on a supplied `source_license_arn` value.
The goal of this PR is to allow us to accept an AWS Marketplace agreement in our main `modernisation-platform` account by hand, and then grant that to member accounts through code, ensuring that our LicenseManager grants are managed in Terraform and not through ClickOps.


davidkelliott (approved):
Looks good, thanks!

PR #4130, opened 2023-06-05
Switch to quiet and add workflow dispatch for SCA

davidkelliott:
Currently to find the errors you need to scroll through all the successful checks, switching to quiet mode will only show the failed checks.

Also adding a workflow dispatch trigger so that this workflow can be manually be run following corrections.


markgov (approved):
LGTM

dms1981 (approved):

PR #4138, opened 2023-06-05
Remove reference to business unit for core-network-services

dms1981:
With the reference to `platforms` in the networking.auto.tfvars.json, we find that our new member environment workflow tries to remove this value. If we choose to investigate this, we'll do so as part of a bug ticket that considers the root cause in a wider context.
With the removal of the reference, there's also a slight change to a data call for a KMS key which is taken into account here.


markgov (approved):
LGTM

davidkelliott (approved):

PR #4156, opened 2023-06-06
Remove SMTP Submission changes

dms1981:
After further discussion with the PPUD migration team the changes to allow hosts in private subnets to send SMTP Submission traffic out to the internet is no longer needed. This PR removes those changes, and re-orders the firewall rules for readability.


ep-93 (approved):
LGTM, its more than just a revert though

Issue #3926, opened 2023-05-22
Allow Route53 query logging in cloudwatch


drobinson-moj:
## User Story

As a webops engineer
I need to capture Route53 query logs
So I can investigate Route53 queries when diagnosing incidents / security incidents.

## User Type(s)

WebOps Engineers

## Value

Security best practice.  Resolves checkov recommendation CKV2_AWS_39

## Assumptions / Hypothesis / Questions / Unknowns

See https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/query-logs.html
Attempting to create using us-east-1 provider results in following error 
```
Error: creating CloudWatch Logs Log Group (route53/development.nomis.service.justice.gov.uk): 
AccessDeniedException: User: arn:aws:sts::***:assumed-role/MemberInfrastructureAccessUSEast/aws-go-sdk-1684769053606027923 is not authorized to perform: logs:CreateLogGroup on resource: arn:aws:logs:us-east-1:***:log-group:route53/development.nomis.service.justice.gov.uk:log-stream: because no identity-based policy allows the logs:CreateLogGroup action
```
Allow `logs:*` or `CreateLogGroup,DeleteLogGroup,DescribeLogGroup,ListTagsLogGroups,TagLogGroup` in us-east-1 

## Definition of done

- [x] another team member has reviewed
- [x] tested route53 query logs can be created in us-east-1 using terraform us-east-1 member provider

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

markgov:
https://github.com/ministryofjustice/modernisation-platform/pull/4162

markgov:
informed Dominic of the change and he is going to test

PR #4162, opened 2023-06-07
Adding extra permissions to create logs

markgov:
As part of [Allow Route53 query logging in cloudwatch#3926](https://github.com/ministryofjustice/modernisation-platform/issues/3926) users needed the ability to create logs for route53 tracking this pr expands on an existing role that has been assigned to the nomis team to include the ability to create logs in the us-east-1 region

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/iam.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/iam.tf
@@ -260,7 +260,7 @@ data "aws_iam_policy_document" "member-access-us-east" {
     #checkov:skip=CKV_AWS_356: Needs to access multiple resources
     effect    = "Allow"
     actions   = ["acm:*",
-      "log:*"
+      "logs:*"
     ]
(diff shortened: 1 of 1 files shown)


terraform/environments/bootstrap/member-bootstrap/iam.tf
```diff
@@ -259,7 +259,9 @@ data "aws_iam_policy_document" "member-access-us-east" {
     #checkov:skip=CKV2_AWS_40
     #checkov:skip=CKV_AWS_356: Needs to access multiple resources
     effect    = "Allow"
-    actions   = ["acm:*"]
+    actions   = ["acm:*",
+      "log:*"
```
ep-93, line 263: is it not logs rather than log?

markgov: you are correct

ep-93 (approved):

PR #4163, opened 2023-06-07
Remove 6DG vpns

dms1981:
Following the migration of CCMS-EBS to the Modernisation Platform, our VPN connections to Six Degrees are no longer required and can be removed. We've confirmed this with the migration team [here](https://mojdt.slack.com/archives/C04EZ78B6US/p1686063147217849).


ewastempel (approved):
whoop whoop!

PR #4168, opened 2023-06-08
add data platform athena load ecr

LavMatt:
adding ecr to contain image for lambda function code being used in data platform's data product development work


markgov (approved):
LGTM

PR #4180, opened 2023-06-09
Only run slack failure when on main

davidkelliott:
This was running for PRs as well which we don't need to know about via slack.  Now we will only get alerted if it fails when merging.


dms1981 (approved):

markgov (approved):
LGTM

PR #4207, opened 2023-06-13
Try to fix slack message triggering

davidkelliott:
Attempting slightly different syntax as per docs https://docs.github.com/en/actions/learn-github-actions/contexts#github-context


markgov (approved):
LGTM

dms1981 (approved):

Issue #4211, opened 2023-06-13
LAA_shared service VPC not connecting to MP VPC


tajewole-moj:
Hi Team,

 1.  created a local TGW attachment for a local VPC in LAA_shared service. (10.200.32.0/19 & 10.200.68.0/22).  and its been accepted/propagated
 2.   routes  added the  tgw attachment as **targets** and the required VPC  as **destinations**( prd vpc  10.27.64.0/21. and dev vpc 10.26.56.0/21) -- these are different aws accounts
I can connect from  local VPC (10.200.32.0/19 & 10.200.68.0/22) to (dev vpc 10.26.56.0/21) fine just cant connect to prd vpc (10.27.64.0/21)
is there a  virtual appliance fw thats blocking access to the prd vpc?

I was told the fw rules are in https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/core-network-services/firewall-rules/production_rules.json

Regards
Tunde

PR #4224, opened 2023-06-13
LAA shared services firewall rules

dms1981:
As part of https://github.com/ministryofjustice/modernisation-platform/issues/4211, this introduces rules allowing access from the VPC for LAA Appstream to the LAA development and production VPCs on the Modernisation Platform


SteveLinden (approved):
LGTM

PR #4228, opened 2023-06-14
Update versions.tf

ep-93:
Upgrading example to use v5


SteveLinden (approved):
or should we leave is as >= 4.0.0, <6.0.0 - as a just in case scenario?

PR #4253, opened 2023-06-15
add lambda function to retrieval policy for repo

LavMatt:
adding the lambda function arn so that it can pull the image from the ecr rep


markgov (approved):
LGTM

PR #4273, opened 2023-06-15
Change to creating-accounts-for-end-users.html.nd.erb

SteveLinden:
Adding comments relating to creating member-unrestricted accounts. Some steps are not required when creating these accounts so this is to clarify the documentation. Also updated the refresh date


source/runbooks/creating-accounts-for-end-users.html.md.erb
```diff
@@ -50,6 +50,8 @@ This will run the [`new-environment` workflow](https://github.com/ministryofjust
 
 This will also run the [`new-environment-files` workflow](https://github.com/ministryofjustice/modernisation-platform/blob/main/.github/workflows/new-environment-files.yml) which generates two new pull requests if this is the first environment of the new application:
 
+If the environment is being created for an account type of member-unrestricted it will not produce the change to run on the modernisation-platform-environments repository and this step can be ignored.
```
davidkelliott, line 53:
```suggestion
If the environment is being created for an account type of member-unrestricted it will not create the PR in the modernisation-platform-environments repository and this step can be ignored.
```

markgov (dismissed):
LGTM

PR #4319, opened 2023-06-16
Update iam.tf , adding cloudtrail permissions to member-access

ep-93:
Related to this request. 

https://mojdt.slack.com/archives/C01A7QK5VM1/p1686923459069869


davidkelliott (changes requested):
We don't want to add all permissions for cloudtrail as they could then delete their account cloudtrail

davidkelliott (approved):

Issue #3970, opened 2023-05-26
Upgrade to AWS Terraform provider 5.0.0


davidkelliott:
## User Story

The latest version of the AWS Terraform provider is out. https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.0.0
There are breaking changes so we should assess and update as appropriate in a controlled manner.
We should provide guidance to users on how to update.

## User Type(s)

MP Engineers
MP Users

## Value

Keeping the platform up to date and secure

## Definition of done

- [ ] update to core mp repos completed
- [ ] users provided with update guidance if needed
- [ ] update to environments repo completed
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

PR #4366, opened 2023-06-20
Update to modules and core-* accounts to v5.

ep-93:
Required for ticket - https://github.com/ministryofjustice/modernisation-platform/issues/3970 

Myself and @ASTRobinson have run a terraform plan before the upgrade, and then after, making sure no resources are to be changed. We have upgraded any module usage to use the new tagged version running on v5. 

Only changes to the code have been non-related to the upgrade, where terraform would not apply even before upgrading from v4 to v5.


davidkelliott (approved):
:shipit:

PR #4383, opened 2023-06-22
Allow shared VPCs to be associated with private zones

dms1981:
Where customers have a requirement for application-local AWS Route53 Private zones, we allow them to create those zones but not to associated them with the relevant business-unit VPC that is shared with their account.

This PR will add the ability for the role assumed by the `provider.core-vpc` used by customers to associate a VPC with a private Route53 zone.


davidkelliott (approved):
👍 thanks for this

PR #4390, opened 2023-06-22
Update Cloud Trail Permissions

harichintala1:


terraform/environments/bootstrap/member-bootstrap/iam.tf
```diff
@@ -40,6 +40,10 @@ data "aws_iam_policy_document" "member-access" {
       "cloudtrail:AddTags",
       "cloudtrail:CancelQuery",
       "cloudtrail:Create*",
+      "cloudtrail:StartLogging",
+      "cloudtrail:StopLogging",
```
davidkelliott, line 44:
```suggestion
```
Could we remove the StopLogging action please?  I'm wary of giving permission to turn off Cloudtrail logs! Thanks

harichintala1:
@davidkelliott  I have removed stoptrail - however I have addded,

  "cloudtrail:DeleteTrail" - in view of when we remove resource from Terraform, it may fail to delete the Cloud trail we created ?

terraform/environments/bootstrap/member-bootstrap/iam.tf
harichintala1: removed 

      "cloudtrail:StopLogging",

davidkelliott:
> @davidkelliott I have removed stoptrail - however I have addded,
> 
> "cloudtrail:DeleteTrail" - in view of when we remove resource from Terraform, it may fail to delete the Cloud trail we created ?

@harichintala1 I don't think we can have this either I'm afraid, we need to ensure that there is no way a Cloudtrail can be stopped or deleted for security reasons.  I think it will have to be if you need to remove a cloudtrail once created you will have to ask one of the MP engineers to manually do it.

harichintala1:
> > @davidkelliott I have removed stoptrail - however I have addded,
> > "cloudtrail:DeleteTrail" - in view of when we remove resource from Terraform, it may fail to delete the Cloud trail we created ?
> 
> @harichintala1 I don't think we can have this either I'm afraid, we need to ensure that there is no way a Cloudtrail can be stopped or deleted for security reasons. I think it will have to be if you need to remove a cloudtrail once created you will have to ask one of the MP engineers to manually do it.

Hi - No problem, I have removed that now !

harichintala1:
@davidkelliott could you review this for latest changes please !

davidkelliott (approved):

PR #4405, opened 2023-06-23
IAM permission changes for RDS

markgov:
Request from the laa team as follows:
Hi MP Team, for disaster recovery planning we are looking into point-in-time restoration for the RDS. Can this permission be provided permanently for us to do this ourselves on the AWS Console - specially I think it’s rds:RestoreDBInstanceToPointInTime? To remove this database manually afterwards I suppose we can request this from your team after the recovery is done.

added to the developer role


davidkelliott (approved):
Sound reasonable 👍

PR #4419, opened 2023-06-26
Update README.md with latest modules

davidkelliott:


ep-93 (changes requested):
`Module for creating an Lambda Functuction`

Should it be `Module for creating a Lambda Functuction`

ep-93 (approved):

PR #4426, opened 2023-06-27
Add additional CIDR ranges for HMPPS - Temp only

davidkelliott:
These are temporary changes to allow moving the Nomis T3 environment and should be reverted afterwards.


dms1981 (dismissed):

dms1981 (commented):
The problem with this PR is that, from the order in the NACLs module, an earlier rule blocking east-west traffic will prevent the new rules from being matched.

PR #4444, opened 2023-06-29
Nuke/data insights change

markgov:
user has requested that there sandbox environment is not rebuilt change setting to exclude env from rebuild a new pr should be raised on monday to add the env back to the rebuild


ASTRobinson (approved):
LGTM

PR #4450, opened 2023-06-30
Changes to set-output code as shown in link on the call log

SteveLinden:
This is the initial code change for the set-output lines on the modernisation-platform repo.

It is based on the code shown in the call log and other research I have done.

It cannot be tested so it will only be "hit" when we create a new environment. Testing will take place when this is run as the code is in new-environment.yml file under the github workflows.

Reference to changes from the call - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/new-environment.yml
+++ b/.github/workflows/new-environment.yml
@@ -169,7 +169,7 @@ jobs:
       - name: get new account(s)
         id: new_account
         run: |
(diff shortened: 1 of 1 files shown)


davidkelliott (changes requested):
I'm don't think this will work, looking at the example here - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

```
- name: Set output
run: echo "::set-output name={name}::{value}"
```

Becomes:

```
- name: Set output
run: echo "{name}={value}" >> $GITHUB_OUTPUT
```

So it looks like you will need to remove the set-output part as well

SteveLinden:
I did this on the second release

davidkelliott (approved):

PR #4457, opened 2023-06-30
Testing version upgrades

davidkelliott:
Testing syntax for allowing any minor and patch version upgrades


davidkelliott:
Didn't work :(

PR #4459, opened 2023-07-03
Commit hash fix.

ep-93:
Missed in two spots.


davidkelliott (commented):
Could you add the version in the comment after please?  Cheers!

SteveLinden (approved):
May want to comment the version number as previously but otherwise approved

SteveLinden (approved):

PR #4477, opened 2023-07-05
addressed checkov failures

dms1981:
* Use of commit hashes is not supported for Terraform Module registry
* See [this terraform issue](https://github.com/hashicorp/terraform/issues/29867) for more detail
* Added skip around `aws_eip` resources as these are used by NAT gateways


ewastempel (approved):
thank you ❤️

PR #4480, opened 2023-07-05
Architectural Decision record following DNS discussion

dms1981:
This ADR captures the outcome of a discussion on DNS and specifically the creation of private hosted zones for customers.


markgov (approved):
LGTM

PR #4506, opened 2023-07-10
Update SQL WORK BENCH permissions

harichintala1:
Update sqlworkbench permissions


dms1981 (approved):
Discussed this one with the team; it's possible for other users of the role to attach a policy that denies this, so with that in mind we're OK to expand this access.

Issue #4470, opened 2023-07-04
Environment Request for Ncas


matt-k1998:
### Application Name

ncas

### Description of application

The Neutral Citation Allocation System (NCAS) is a business critical application that allows The Royal Courts of Justice Courts Recording and Transcription Unit staff (CRATU) to create citation entries for cases heard within one of the Royal Courts of Justice (RCJ) courts. CRATU staff (staff users) are able to create, search and view previous citation entries. Judges and judges clerks are able to view citations where these are available for new cases allocated to a named judge.

### GitHub team slug

dts-legacy

### Environments

| Environment   | Access level |
| ---     | --- |
|<ul><li>[X] Development</li></ul> | <ul><li>[ ] view-only</li><li>[ ] developer</li><li>[X] sandbox</li></ul> |
|<ul><li>[ ] Test</li></ul>   | <ul><li>[ ] view-only</li><li>[ ] developer</li></ul> |
|<ul><li>[X] Preproduction</li></ul>| <ul><li>[ ] view-only</li><li>[X] developer</li></ul> |
|<ul><li>[X] Production</li></ul> | <ul><li>[ ] view-only</li><li>[X] developer</li></ul> |

### Tags

tag | value
--- | ---
application | Ncas
business-unit | HMCTS
infrastructure-support | dts-legacy-apps-support-team@hmcts.net
owner | dts-legacy-applications:dts-legacy-apps-support-team@hmcts.net

### Subnet sets

- [X] General - this application will use the same subnet as other applications in your business unit
- [ ] Isolated - this application has highly sensitive data which must have its own isolated subnet with no network connectivity

### How do users connect to the application?

- [X] Over the public internet
- [ ] With a purple cabled device (please give details)
- [X] With a MoJ Official device

## Additional features

- [ ] Additional VPC Endpoints
- [ ] Extended DNS Zones
- [ ] Other - please specify

## Definition of done

- [ ] account created
- [X] customer informed

ASTRobinson:
Environment setup completed and customer informed.

PR #4507, opened 2023-07-10
Create NCAS environment files

ASTRobinson:
Creating new account as per #4470

### DIFF
--- a/policies/environments/expected.rego
+++ b/policies/environments/expected.rego
@@ -38,6 +38,7 @@ expected =
     "maatdb",
     "mlra",
     "mojfin",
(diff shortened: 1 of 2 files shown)


policies/environments/expected.rego
```diff
@@ -38,6 +38,7 @@ expected =
     "maatdb",
     "mlra",
     "mojfin",
+    "ncas",
```
ewastempel, line 41: I don't think it needs to be ordered alphabetically, but since we've kept it that way so far, it's probably worth continuing

ewastempel (approved):
LGTM

PR #4545, opened 2023-07-12
📝 Update to creating new environments for end users

ASTRobinson:
Updated section 6 as the editing of the PR to comment out any environments that are not used in the GitHub Actions workflow file is no longer required.


ep-93 (approved):

dms1981 (approved):

dms1981:
I did actually wonder about the matrix list that ends up in an account workflow, and can see that it's created dynamically by pulling the contents of the `environments.json` and using `jq` to query for environments.

PR #4551, opened 2023-07-13
Update new story template

dms1981:
Wrapped the definition of done into comments so that we deliberately write a definition as part of a story, instead of just accepting the template that may or may not be appropriate.


markgov (approved):
LGTM

PR #4568, opened 2023-07-14
Add rule allowing ldap from cp to hmpps dev

adeweetman-al:


dms1981 (commented):
@adeweetman-al - do you also need tcp/636 for ldaps?

adeweetman-al:
@dms1981  - no, just tcp/389.

dms1981 (approved):

PR #4590, opened 2023-07-18
Automate changes to member environments

dms1981:
This PR (to be converted to a draft) is a rework of an existing attempt (https://github.com/ministryofjustice/modernisation-platform/pull/4579).
Given that I can't test this locally, this PR has been raised to confirm that the `git diff` step works correctly. If so, the job output for `find-environments` can be set and used in following jobs as part of the whole action.
When converted from a draft, this PR should do the following:
* Get the directories with changes
* Initialise terraform, then loop through workspaces and conduct a terraform plan, skipping the job if the workspace does not exist (eg, running a plan in the case of `cooker-development`, but skipping the job in the case of `cooker-test`.
* Initialise terraform, then loop through `development` and `test` workspaces and conduct a terraform plan, outputting a `.tfplan` file that can be used by a following terraform apply step.


davidkelliott (approved):
👍 looks really good, thanks for this.

PR #4599, opened 2023-07-19
Lambda snapshot delete

ep-93:
WIP.

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/instance-scheduler.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/instance-scheduler.tf
@@ -20,9 +20,12 @@ data "aws_iam_policy_document" "instance-scheduler-access" {
     #checkov:skip=CKV_AWS_356: Resources not known in advance
     effect = "Allow"
     actions = [
-      "ec2:DeleteSnapshot"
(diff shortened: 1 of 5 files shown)


SteveLinden (approved):

terraform/environments/bootstrap/member-bootstrap/instance-scheduler.tf
```diff
@@ -23,8 +23,7 @@ data "aws_iam_policy_document" "instance-scheduler-access" {
       "ec2:DescribeInstances",
       "ec2:DescribeInstanceStatus",
       "ec2:DescribeTags",
-      "ec2:StartInstances",
-      "ec2:StopInstances"
```
dms1981, line 27: Did you mean to remove the ability for the instance scheduler to stop instances?

dms1981 (changes requested):
Are we sure this is ready to merge?

terraform/environments/bootstrap/member-bootstrap/instance-scheduler.tf
SteveLinden: Edd put together this release while I was off. I assumed this is added as part of the later release when the IAM changes are made but not 100% sure

PR #4607, opened 2023-07-21
add enable, disable event permissions to data engineer policy

LavMatt:
Have added these permissions as are required to disable and enable object created event triggers in eventbridge, during a migration exercise


dms1981 (approved):
LGTM

Issue #3883, opened 2023-05-17
Allow developers to delete account-local ECR images.


dms1981:
## User Story

As a Modernisation Platform developer
I want to be able to manually delete ECR images in my account
So that I don't need to rely on MP engineers to do this for me
https://mojdt.slack.com/archives/C01A7QK5VM1/p1684324129674539
## User Type(s)

Modernisation Platform Customer

## Value

Customers who maintain ECR repositories in **their** accounts sometimes need to delete those images (specifically, `MLRA`). If we add the permission to delete ECR images to the `developer` role, customers will be able to delete images before tearing down a development account.

## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

ASTRobinson:
https://github.com/ministryofjustice/modernisation-platform/pull/4608

PR #4608, opened 2023-07-21
Add ECR delete permissions to developer role

ASTRobinson:
As per ticket #3883 this PR grants the developer role permission to delete an existing image repository (AWS docs located [here](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonelasticcontainerregistry.html)).


SteveLinden (approved):
Assuming they can only delete local ECR images and not others it looks fine.

PR #4610, opened 2023-07-21
Allow customers to send outbound email through egress VPCs

dms1981:
As discussed in [this slack thread](https://mojdt.slack.com/archives/C01A7QK5VM1/p1689849118728639), SES is not always an appropriate choice for customers to relay outbound email through. While we would generally prefer that SES be used, this PR will allow customers in the same situation as the JitBit application to send traffic out over the [SMTP submission port](https://www.cloudflare.com/en-gb/learning/email-security/smtp-port-25-587/#:~:text=The%20official%20default%20port%20for%20SMTPS%20is%20port%20587.%20SMTPS%20connections%20start%20with%20a%20%22STARTTLS%22%20command%20to%20let%20the%20mail%20server%20know%20that%20the%20SMTP%20traffic%20will%20be%20sent%20over%20TLS.)

Further to this, this PR removes the now-unneeded SES users that were created previously.

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/inline_fqdn_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/inline_fqdn_rules.json
@@ -9,6 +9,7 @@
     "ntservicepack.microsoft.com",
     "stats.microsoft.com",
     "saas40.kaseya.net",
-    ".dl.delivery.mp.microsoft.com"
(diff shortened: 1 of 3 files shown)


SteveLinden (approved):
LGTM

davidkelliott (approved):

PR #4626, opened 2023-07-24
Add discovery permissions to migration role

ASTRobinson:
As per Slack conversation [here](https://mojdt.slack.com/archives/C013RM6MFFW/p1690193214354109) this PR adds the discovery:* permissions to the migration role. If this fails to resolve the issue for the users I believe the next step would be to add the 'AWSApplicationDiscoveryServiceFullAccess' policy to the migration role as per David E. suggestion.


dms1981 (approved):

dms1981:
From what I can see, this gets the permissions set to the same place that `AWSApplicationDiscoveryServiceFullAccess` is at so there may not be much value in applying it in a later PR.

Issue #4538, opened 2023-07-12
Connectivity from MoJ Official devices to DIH Redshift


davidkelliott:
## User Story

Member request from Stephen Jobling direct via slack
Use Power BI on an MoJ Official device to access data in the Redshift cluster in data-and-insights-hub-wepi accounts

I believe the following is required - 

- firewall rule allowing traffic from global protect to hq-<environment> on port 5439
- VPC endpoint for Redshift

Routing from moj official devices to MP TGW should already be in place.

## User Type(s)

MP user

## Value

Allow users to securely access Redshift clusters locally.

## Definition of done

- [x] connectivity established
- [ ] Stephen Jobling has tested and able to access as required
- [x] another team member has reviewed
- [x] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

SimonPPledger:
@davidkelliott to add requestor details

dms1981:
Raised https://github.com/ministryofjustice/modernisation-platform/pull/4630, but this requires further discussion.

dms1981:
Raised https://github.com/ministryofjustice/modernisation-platform/pull/4635 which should create the necessary endpoints and allow access from within the HQ VPCs (will need a further update to the VPC module to update the VPC endpoint security groups).

dms1981:
Suggested customer use something like this to get traffic from AWS LB to Redshift VPC endpoints:
```
data "aws_vpc_endpoint" "redshift-data" {
  vpc_id       = data.aws_vpc.shared.id
  service_name = "com.amazonaws.eu-west-2.redshift-data"
}

data "aws_network_interface" "redshift-data" {
  for_each = data.aws_vpc_endpoint.redshift-data.network_interface_ids
  id       = each.value
}

resource "aws_lb" "redshift-data" {}
resource "aws_lb_target_group "redshift-data" {}
resource "aws_lb_target_group_attachment" "redshift data" {
  for_each         = data.aws_network_interface.redshift-data
  target_group_arn = aws_lb_target_group.redshift-data.arn
  target_id        = each.value.private_ip
  port             = 5439
}
```

dms1981:
[...]

PR #4635, opened 2023-07-25
Allow access to Redshift VPC endpoint from Global Protect users

dms1981:
As part of https://github.com/ministryofjustice/modernisation-platform/issues/4538, this PR allows traffic from MOJ Global Protect users to access a redshift VPC endpoint.
There would also need to be a corresponding change to the [modernisation-platform-terraform-member-vpc](https://github.com/ministryofjustice/modernisation-platform-terraform-member-vpc/blob/d568b0d83bb7cec82ee9eddb7c264cf8c02757a4/main.tf#L317) module to amend the security group, and a method for an AWS Load Balancer target group attachment to discover the relevant IP address (probably a combination of a data call and a for_each expression).

### DIFF
--- a/environments-networks/hq-development.json
+++ b/environments-networks/hq-development.json
@@ -10,7 +10,7 @@
   "options": {
     "bastion_linux": true,
     "additional_cidrs": [],
-    "additional_endpoints": [],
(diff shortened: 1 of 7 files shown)


markgov (approved):
LGTM

PR #4637, opened 2023-07-26
Update redshift VPC endpoint type

dms1981:
From further research, it appears that the `redshift-data` endpoint is the appropriate option for RedShift Serverless querying.


markgov (approved):
Approved

PR #4659, opened 2023-07-27
Fix inappropriate triggering of `terraform-member-environments`

dms1981:
According to the [GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths) on path filtering, I believe that the current path options are causing the `terraform-member-environments` action to trigger inappropriately when changes are made to terraform files in `core-*`.

Although the ordering is correct, I think that the use of `**/*.tf` is being matched and as it is more specific than `!core-*`, exclusions are not being applied as originally anticipated.

### DIFF
--- a/.github/workflows/terraform-member-environment.yml
+++ b/.github/workflows/terraform-member-environment.yml
@@ -7,14 +7,14 @@ on:
       - main
     paths:
       - 'terraform/environments/**/*.tf'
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

PR #4667, opened 2023-07-27
LAWS-3284-extend-dns-zone: locals private application zones and priva…

tmahmood72:
local private application zones an private dns zone extend added to the vpc.tf file


markgov:
Issue with module leave it with me will take a look

markgov (approved):
merging so i can test on a new branch

PR #4678, opened 2023-07-28
Exclude bootstrap subdirectories from member-environments updates

dms1981:
As the `terraform-member-environments` workflow did not exclude the bootstrap directories, changes here caused an inappropriate triggering of the workflow.


markgov (approved):
LGTM

davidkelliott (approved):

PR #4694, opened 2023-07-28
Reworked private zone association module

dms1981:
This PR reworks the private zone association module. 

* Some renaming for consistency with existing code has been conducted.
* The required zones to associate with a VPC are now specified in `environments-networks/$business_unit-$environment.json` as `additional_private_zones`.
* Removed `aws.prd.legalservices.gov.uk` as this zone does not currently exist.
* The module call in `core-vpcs/vpc.tf` has been reworked to create a map of private zones from the relevant json and supply them to the module as a map.
* The module source now uses `for_each` expressions to take in the possibility of multiple private zones, and create associations for each.


davidkelliott (approved):
Nice, I like this approach, we could also consider having application public zones stored in the app json.  That also may make it easier to increase the security around them.

PR #4697, opened 2023-07-31
Updated docs for private zone association

dms1981:
Following on from https://github.com/ministryofjustice/modernisation-platform/pull/4694, this PR updates the documents to match the revised approach


ewastempel (approved):
LGTM

Issue #4516, opened 2023-07-11
Deployment reviewers disappearing


dms1981:
## Expected Behavior

When I manually add a reviewer to an environment in the [modernisation-platform-environments](https://github.com/ministryofjustice/modernisation-platform-environments) repository, that reviewer should remain able to review deployments to that environment.

## Actual Behavior

After manually adding a collaborator to review deployments to `ppud-development`, or any other `ppud` environment, the collaborator is later removed through an unknown method.

## Steps to Reproduce the Problem

Unknown - but I suspect that when new accounts are created, the automation in place also corrects any manual changes to environments.

## Other information

[ppud-preproduction](https://github.com/ministryofjustice/modernisation-platform-environments/settings/environments/866653520/edit)

dms1981:
I think this is happening when the `git-create-environments.sh` runs as part of the `.github/workflows/terraform-github.yml` action.

ASTRobinson:
PR - https://github.com/ministryofjustice/modernisation-platform/pull/4725/

ASTRobinson:
Work in progress...

```
create_environment() {
  environment_name=$1
  github_teams=$2
  echo "Creating environment ${environment_name}..."

  # Function to add a maximum of 6 reviewers (individuals and teams)
  function add_reviewers {
    local reviewers_to_add=$1
    if [ -n "$existing_reviewers" ]; then
      updated_reviewers="${existing_reviewers},$reviewers_to_add"
    else
      updated_reviewers="$reviewers_to_add"
    fi
  }

  # Fetch existing reviewers for the environment
  existing_reviewers=$(curl -L -s \
    -H "Accept: application/vnd.github.v3+json" \
    -H "Authorization: Bearer ${secret}" \
    -H "X-GitHub-Api-Version: 2022-11-28" \
    https://api.github.com/repos/${repository}/environments/${environment_name} | jq -r '.reviewers | map(.login) | join(",")')

  # Fetch existing teams for the environment
  existing_teams=$(curl -L -s \
    -H "Accept: application/vnd.github.v3+json" \
    -H "Authorization: Bearer ${secret}" \
    -H "X-GitHub-Api-Version: 2022-11-28" \
    https://api.github.com/repos/${repository}/environments/${environment_name}/teams | jq -r '.teams | map(.slug) | join(",")')

  # Check if the total number of reviewers exceeds 6 (teams + individuals)
  reviewers_array=()
[...]

PR #4725, opened 2023-08-02
Deployment Reviewers Disappearing Fix

ASTRobinson:
As per ticket [Deployment reviewers disappearing#4516](https://github.com/ministryofjustice/modernisation-platform/issues/4516) this PR updates the `create_environment` function to fetch a list of GitHiub reviewers and then updates the list rather than overwriting it.

### DIFF
--- a/scripts/git-create-environments.sh
+++ b/scripts/git-create-environments.sh
@@ -48,22 +48,38 @@ create_environment() {
   environment_name=$1
   github_teams=$2
   echo "Creating environment ${environment_name}..."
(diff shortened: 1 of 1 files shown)


ewastempel (commented):
Nice work, I really like the approach, but a couple of questions:
1. have you seen it working? I cannot find a relevant pipeline that would confirm that.
2. should the existing  reviewers be added first, or the GH teams (from the env file)?
3. what about the limit of 5 reviewers? I don’t see it being taken into account

PR #4731, opened 2023-08-03
Temporarily add aws dev legalservices hostedzone to dns iam policy

ASTRobinson:
As per slack [request](https://mojdt.slack.com/archives/C01A7QK5VM1/p1691055181067629) i am adding the hosted zone to the policy as a temporary workaround while the bug/issue is investigated.


ewastempel (approved):
This is OK only as a temp fix.

Issue #4139, opened 2023-06-05
Look into if networking.auto.tfvars.json files for core accounts are needed


davidkelliott:
## User Story

We have networking.auto.tf.vars.json for our core accounts, this file is used for RAM sharing subnet sets from the core-vpc accounts to member accounts.  It is possible that this file was generated as part of automation but is no longer needed and not used.

The business unit is not populated on this file and has caused confusion as to why this is different from the standard way of doing this.  (see here - https://github.com/ministryofjustice/modernisation-platform/pull/4138) 

This ticket is to work out if this is the case and either remove it or use it properly.

## User Type(s)

MP engineers

## Value

Clean code means less confusion and errors

## Definition of done

- [ ] automations updated to either not generate or generate with business unit
- [ ] files removed if not needed
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
Adding the core accounts to `platforms-production.json` and re-running [provision-environment-directories.sh](https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/provision-environment-directories.sh#L69) adds the missing keys to the `networking.auto.tfvars.json` files.

dms1981:
However, given that core accounts don't need a `networking.auto.tfvars.json` to be fed into [member-account-ram-association.sh](https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/member-account-ram-association.sh), or used in the [ram-principal-association](https://github.com/ministryofjustice/modernisation-platform/tree/main/terraform/modules/ram-principal-association) module, I'll next look at excluding core accounts from having this file autogenerated.

PR #4743, opened 2023-08-08
Conditionally generate tfvars for non-core accounts

dms1981:
This PR relates to https://github.com/ministryofjustice/modernisation-platform/issues/4139. As we make use of `networking.auto.terraform.tfvars` to share resources, it's not relevant to our core accounts and as such this PR makes its creation conditional.
* Updates the scripts that output this file to conditionally skip the creation when we read the account type from `environments/*.json` and see that they account-type value is core.
* Removes `networking.auto.terraform.tfvars` from core accounts.
* Adds defaults to `base_variables.tf` for both core-sandbox and core-vpc accounts to accommodate removal of tfvars.
* Tidies up some of core-sandbox account to remove some unused code and fix a broken data call.


markgov (approved):
LGTM

PR #4757, opened 2023-08-08
Add data block for use in portal-dev dns IAM policy

ASTRobinson:
Following slack conversation [here](https://mojdt.slack.com/archives/C01A7QK5VM1/p1691484375454639) as the hosted zone id changing following an update this PR uses s `data` block to obtain the zone_id.


dms1981 (approved):
You might be able to do it a bit more directly by referencing the key in the resource that creates the zone, but this is fine in the short term.

Issue #4578, opened 2023-07-17
Issue on >> GITHUB_OUPUT for files


SteveLinden:
## Expected Behavior

When creating files with github output (echo "files=$(git diff --no-commit-id --name-only --diff-filter=AM -r @^ | awk '{print $1}' | grep ".json" | grep -a "environments//*"  | uniq | cut -f2-4 -d"/" | sed 's/.\{5\}$//')" >> $GITHUB_OUTPUT) it will not work for instances where there are multiple changes to the environments/*.json files

It needs to be amended to allow multiple files to be created. Information on this is shown in https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

## Actual Behavior

An example of the error received when there were changes (new environment) and another change to amend the tags in the same job. Shown in https://github.com/ministryofjustice/modernisation-platform/actions/runs/5544470404/jobs/10122623956

Run echo "files=$(git diff --no-commit-id --name-only --diff-filter=AM -r @^ | awk '{print $1}' | grep ".json" | grep -a "environments//*"  | uniq | cut -f2-4 -d"/" | sed 's/.\{5\}$//')" >> $GITHUB_OUTPUT
Error: Unable to process file command 'output' successfully.
Error: Invalid format 'wardship'

## Steps to Reproduce the Problem

  1. Change two json files in the environments folder and then run the PR

PR #4759, opened 2023-08-08
Wrap git diffs into delimiters to handle multiline outputs

dms1981:
In line with the guidance offered by GitHub [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings), this PR uses field delimiters to allow the output of multiline strings into the `$GITHUB_OUTPUT`.

### DIFF
--- a/.github/workflows/new-environment.yml
+++ b/.github/workflows/new-environment.yml
@@ -169,7 +169,10 @@ jobs:
       - name: get new account(s)
         id: new_account
         run: |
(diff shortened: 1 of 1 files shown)


dms1981:
NB. This PR should resolve https://github.com/ministryofjustice/modernisation-platform/issues/4578

markgov (approved):
LGTM

PR #4763, opened 2023-08-09
✨Add Connor to GitHub Team

connormaglynn:


terraform/github/locals.tf
```diff
@@ -15,7 +15,8 @@ locals {
     "ewastempel",
     "jakemulley",
     "SteveMarshall",
-    "davidkelliott"
+    "davidkelliott",
+    "connormaglynn"
```
connormaglynn, line 19: ☝️ I'm an Organisation owner, so I need to be on the maintainers' list 🔧

ewastempel: ok

PR #4764, opened 2023-08-09
allow migration role to be passed to AWS service role

dms1981:
This PR resolves an issue in the use of the Migration role and the AWS Migration service. See [this](https://mojdt.slack.com/archives/C01A7QK5VM1/p1691569159990639) Slack conversation for the full context.


markgov (approved):
LGTM

ASTRobinson (approved):

PR #4770, opened 2023-08-10
new secret for environment

markgov:
creating a new secret outside of the environments repo so that is is not deleted


SteveLinden (approved):
Looks OK based on the code above so approved.

PR #4773, opened 2023-08-11
adjusted secrets for pra-register to fix SCA errors

dms1981:
* Added KMS call to encrypt secret
* Tidied up naming as it doesn't need to be self-referential


markgov (dismissed):
LGTM

markgov (approved):
LGTM

PR #4786, opened 2023-08-11
Add ec2 CreateVpcEndpointServiceConfiguration permissions

ASTRobinson:
As per Slack [conversation](https://mojdt.slack.com/archives/C01A7QK5VM1/p1691760358946219) this PR adds the ec2:CreateVpcEndpointServiceConfiguration to the developer role.


dms1981 (approved):

dms1981 (approved):
Discussed with team; this should allow only the creation of services to be used with privatelink rather than creation of endpoints in general, so this is an appropriate set of permissions.

PR #4799, opened 2023-08-15
Initial change to see if VPN amendment works

SteveLinden:
This is the initial test to see if adding new addresses will clear an issue outstanding on NOMS. They have lost access for some time due to losing both VPNs.
The change includes a  new VPN (which may not work, unknown - noms_dev_vpn_static_routes) and adding another address to the current noms_live_vpn_static_routes based on a value provided for the azure system.
The former is "10.102.1.79" (dev/test) and the latter "10.40.129.79" (prod). Have been told the dev/test azure location is not an issue if that breaks.

A test pull request will be undertaken to make sure there are no obvious issues.

A pull gave me issues which I sorted out but only created the production line I added. I need to put some additional code in a json so I will do this once I decide what we need. Various other changes in here to correct initial plan issues.


terraform/environments/core-network-services/vpn_attachments.json
```diff
@@ -47,6 +47,30 @@
     "tunnel_dpd_timeout_seconds": "45",
     "tunnel_startup_action": "start"
   },
+  "NOMS-Transit-Dev-VPN-VNG_1": {
```
dms1981, line 50: What purpose do these VPNs serve?

terraform/environments/core-network-services/locals.tf
```diff
@@ -58,7 +58,11 @@ locals {
     "10.102.0.0/16",
     "10.47.0.0/26",
     "10.47.0.64/26",
-    "10.47.0.128/26"
+    "10.47.0.128/26",
+    "10.40.129.79/32"
```
dms1981, line 62: Assuming we only want to check against a single host in devtest, according to the story comments we only need `10.102.1.79/32`, so we don't need to make any changes to this local.

terraform/environments/core-network-services/locals.tf
```diff
@@ -58,7 +58,11 @@ locals {
...
+  ]
+  noms_dev_vpn_static_routes = [
```
dms1981, line 64: LGTM - you'll need to route this through both the `NOMS-Transit-Live-VPN-VNG_1` and `NOMS-Transit-Live-VPN-VNG_2` VPNs.

dms1981:
See this PR - this one won't be needed.

PR #4800, opened 2023-08-16
New permissions on  MemberInfrastructureAccess to add fsx:*

SteveLinden:
Adding permission to solve issue in the Ask channel


davidkelliott (changes requested):
Line looks good but this is in the wrong policy, it needs to be in the member-access policy

davidkelliott (approved):

PR #4846, opened 2023-08-21
OIDC Step 2

ep-93:
Now I have added all of the changes to the environments json this adds step 2, the configuration needed for them to be used.

After I'll need to change workflows to 

```role-to-assume: "arn:aws:iam::${{ env.ACCOUNT_NUMBER }}:role/modernisation-platform-oidc-cicd"```

### CODE REVIEW

**Comment by dms1981** [comment]
Related to which card @ep-93 ?

**Review by dms1981 — DISMISSED** [review]

File: environments/cooker.json [hunk 1]
```diff
@@ -16,5 +16,7 @@
     "business-unit": "Platforms",
     "infrastructure-support": "modernisation-platform@digital.justice.gov.uk",
     "owner": "Modernisation Platform: modernisation-platform@digital.justice.gov.uk"
-  }
+  },
+  "go-live-date": "",
```
dms1981, line 20: If this is blank, does it serve any purpose? Could we just set the same go-live date as other platform accounts here?

environments/cooker.json
```diff
@@ -16,5 +16,7 @@
...
+  "github-oidc-team-name": "github-oidc-repositories"
```
dms1981, line 21: If we're potentially having multiple OIDC values, maybe something like...
```
  "oidc" : {
    "github-team-name": "github-oidc-repositories"
  }
```

environments/cooker.json
ep-93: Well, it's not live so doesn't have a go live date, same with example.

davidkelliott (changes requested):
I don't think this will work as it is, it will pass in a full list of all the repos in all the files won't it?  I think you need to filter the files using application name so you only get the repos for that application

environments/cooker.json
ep-93: Ive gone with github-oidc-team-repositories"

ASTRobinson (approved):

PR #4883, opened 2023-08-24
🪛 Fix member-environments bespoke changes workflow to avoid accidental recursive matching

dms1981:
The current path filtering still appears to be triggering inappropriately. This change should prevent changes anywhere but in customer environments from triggering the workflow


ep-93 (approved):

markgov (approved):
LGTM

PR #4900, opened 2023-08-29
Dynamically populate list of private zones

dms1981:
This PR does the following
* Replaces individual data calls for private hosted zones with single call fed by `for_each`
* Replaces individual zone strings in IAM resources for `dns` policy with `for` expression to filter output of new data call

In short, this PR will remove the need to create individual data calls for each new private zone we introduce.


SteveLinden (approved):
Looks OK to me

davidkelliott (approved):
🙌  nice!

PR #4906, opened 2023-08-30
Update modernisation-platform-terraform-baselines to v6.0.0

SteveLinden:
Contains the correct reference to the new release.


davidkelliott (changes requested):
This isn't backing it out this is implementing the changes and moving to v6

PR #4918, opened 2023-08-31
Update OIDC role module version

dms1981:
* Update to v2.0.1 of `modernisation-platform-github-oidc-role` module.
* Remove wildcard from value in data-platform.json as not needed with new module version.


ewastempel (approved):
LGTM

PR #4931, opened 2023-09-04
Enrolling Mark and Aaron onto MP on-call rota

ewastempel:
As discussed within the team, adding Mark and Aaron to our MP on-call rota.
Note, the immediate on-call schedule (~next two months) will be preserved by manual overrides after this code gets merged.
The TF plan has been run manually with no issues.


davidkelliott (approved):
Looks good 👏

PR #4946, opened 2023-09-06
add new data platform accounts to ecr repo permissions

LavMatt:
Have added in the newly created accounts for data platform to ecr repo permissions

Also removed 2 repos that we will not be using moving forward


ep-93 (approved):
As long as what you are removing isn't used by anyone else, all approved.

PR #4951, opened 2023-09-06
Changed release to v6.1.0 for backup changes

SteveLinden:
Changing the code to point to the v7.0.0 release on modernisation-platform-terraform-baselines.


davidkelliott (changes requested):
If this is a breaking change we need better notes in the release to explain what needs to change in the code here, if not a breaking change please change the release name to an non breaking change version number

davidkelliott (approved):

PR #4966, opened 2023-09-07
Bumping the version of trusted advisor lambda function

ewastempel:
This includes a bump of the node.js that the lambda is using to v.18 and the migration of the AWS SDK from v2 to v3.

Tested locally and in the console.


markgov (approved):
LGTM

PR #5002, opened 2023-09-14
add outbound udp rules for fixngo domain joins

robertsweetman:
- add allowed udp ports out to fixngo ad domains
- test, preproduction and production
- succeeded in development environment already


markgov (approved):
LGTM

PR #5008, opened 2023-09-15
adding extra permissions

markgov:
User is still having issues so adding in extra permissions


ep-93 (approved):
Approved, but discusssed with Mark about alternative roles.

Issue #3859, opened 2023-05-12
Refactor VPC Endpoint code for core-shared-services


dms1981:
## User Story

As a modernisation platform customer
I expect to have VPC endpoints consistently used in `core-shared-services`
So that I have a predictable user experience

## User Type(s)

* Modernisation Platform customer using image-builder
* Modernisation Platform engineer

## Value

After looking at the code in the `environments/core-shared-services` directory it appears that terraform is only configured to create VPC endpoint resources in the `non_live_data` VPC.

Using VPC endpoints for AWS services is a more performant, and more cost effective solution than routing traffic via our Transit Gateway out to the internet.

## Definition of done

- [ ] terraform refactored
- [ ] VPC endpoints consistent in `live_data` and `non_live_data`
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

PR #5100, opened 2023-09-26
Refactor core-shared-services VPC endpoints

dms1981:
This PR changes the code used to create VPC endpoints in the `core-shared-services` VPCs.
At present VPC endpoints are only created in the `non_live_data` VPC.
This PR will destroy and then recreate those VPC endpoints in both the `live_data` and `non_live_data` VPCs.
It will also amend the security groups used to secure the endpoints so that they are created in both VPCs and attached to both sets of endpoints.

This PR will resolve https://github.com/ministryofjustice/modernisation-platform/issues/3859


markgov (approved):
LGTM

Issue #4974, opened 2023-09-08
Add manual static routes for VPN to Azure


seanprivett:
## User Story

The VPN to Azure goes down whenever Microsoft do regular security patches. MP team understands why this is - there is an issue with BGP. Suggestion from David Sibley is "add manual static routes that matched the advertised networks and pointed them via the PTTP TGW"

## Value

This becomes more critical and valuable as more services go into production from Azure to MP. The value is in providing a stable VPN.
## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

SimonPPledger:
we cannot do this directly, as Azure has too many IPs - this would impact other traffic. so we might need to add some other static routes

dms1981:
As I see this, the path to introducing routes propagated via BGP into the inspection route table looks like this:
* Add manual static routes matching those being advertised to ensure a consistent flow of traffic to Azure via the MOJ TGW (https://github.com/ministryofjustice/modernisation-platform/pull/5111)
* Activate propagation from the NOMS VPN connections to the inspection route table
* Remove NOMS static routes from the inspection route table and confirm that propagated routes are installed to replace them

dms1981:
https://github.com/ministryofjustice/modernisation-platform/pull/5120
^^ This PR configures the propagation as discussed in my last comment

dms1981:
https://github.com/ministryofjustice/modernisation-platform/pull/5122
^^ This PR removes the now-unneccesary static routes

PR #5111, opened 2023-09-29
Add Azure static routes to inspection route table

dms1981:
As part of #4974 this PR adds static routes being advertised over the two `NOMS-Transit-Live` and two `NOMS-Transit-Live-DR` VPN connections to the TGW `inspection` route table (`external_inspection_out`). These routes will have a next-hop of the the TGW peering attachment with the MOJ TGW.

This will allow us to propagate routes into the `inspection` route table from the `NOMS-Transit-Live` and `NOMS-Transit-Live-DR` VPNs without inadvertently sending traffic via the VPNS.


ASTRobinson (approved):

SteveLinden (approved):
Assuming the IP addresses are correct LGTM

PR #5121, opened 2023-10-02
Developer to migration role.

ep-93:
For the migration, the team will need permissions to create AMIs from snapshots, and it would be good to know if the migration role has this permission and that it would work.


markgov (approved):
LGTM

PR #5148, opened 2023-10-04
🛂 Add Data Platform Apps and Tools OIDC provider

jacobwoffenden:
Data Platform's Open Metadata instance running on EKS in Data Platform's Apps and Tools account needs to assume roles in our data accounts (data-platform-development) to collect metadata from Glue and S3

We will use IRSA to achieve this in the cluster

An explicit deny prevented me from doing this in the environments repository

Signed-off-by: Jacob Woffenden <jacob.woffenden@digital.justice.gov.uk>


jacobwoffenden:
This is no longer required.

I've removed the OIDC provider from data-platform-development account.

Issue #5001, opened 2023-09-13
Give members access to firewall logs and core networking


davidkelliott:
## User Story

As an MP user
I want to inspect firewall logs and core networking VPCs and firewall rules
So that I can troubleshoot network connectivity.

Need to check there is not one already

## User Type(s)

MP users

## Value

Enable users to solve their own problems

### Proposal
Create new role (or expand and rename the read-dns-role) to allow users to view firewall, firewall logs and VPC flow logs. 
Is there an issue with renaming read-dns-role?

## Definition of done

- [x] role created or amended
- [x] readme has been updated
- [x] user docs have been updated
- [x] another team member has reviewed
- [x] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
Created role, confirmed can assume it and view specified CloudWatch Log Groups, Streams, and events for firewalls, as well as firewall policies and rule groups

dms1981:
There turned out to be a few complications with the replacement of `read-dns` with `read-logs`.
https://github.com/ministryofjustice/modernisation-platform/pull/5152 >> this PR replaced the `read-dns` role.
https://github.com/ministryofjustice/modernisation-platform/pull/5157 >> this PR amended the IAM permissions for the new ReadFirewallRecords policy.

dms1981:
We found that there were other references outside of the Modernisation Platform repository, particularly in Modernisation Platform Environments, that made use of the `read-dns` role and also needed refactoring. As an interim measure I raised this PR to recreate the role while we find out why `read-logs` cannot be assumed.
https://github.com/ministryofjustice/modernisation-platform/pull/5158

dms1981:
`member-delegation-read-only` role can now read VPC flow logs.
`read-log-records` role can read VPC flow logs in `core-network-services` as well as firewall configuration

dms1981:
TODO: remove permission to assume `read-dns-records`
TODO: remove `read-dns-records` role

dms1981:
Removed the trust relationship and role - all done!

PR #5152, opened 2023-10-05
Allow members to read firewall logs

dms1981:
In response to #5001 , this PR renames the `read-dns-records` role to `read-log-records` and attaches a new policy called `ReadFirewallRecords` which will allow this role to view CloudWatch log groups related to our AWS NetworkFirewalls, as well as Network Firewall policies and rule groups through the `network-firewall:Describe*` and `network-firewall:List*` permissions.


ep-93 (dismissed):

davidkelliott (commented):
I think there is text referring to read-dns-records in the user guidance as well as the image - https://github.com/ministryofjustice/modernisation-platform/blob/main/source/user-guide/member-read-only-core-accounts.html.md.erb#L27

markgov (approved):
LGTM

PR #5157, opened 2023-10-05
Fix/5001 permissions

dms1981:
Updates resource statements to allow restricted access to `fw-*` CloudWatch Log Groups, Log Streams, and events.

Updates resource statement to allow restricted access to AWS Network Firewall policies and rule groups.


markgov (approved):
LGTM

PR #5159, opened 2023-10-06
New upgrade account for ccms-ebs.

SteveLinden:
New account for the ccms upgrade work. It's called ccms-ebs-upgrade to distinguish it from the other accounts. Nuke is set to exclude as it was for the original ccms-ebs account.


environments/ccms-ebs-upgrade.json
```diff
@@ -0,0 +1,24 @@
...
+    }
+],
+"tags": {
+  "application": "ccms-ebs-upgrade",
+  "business-unit": "LAA",
+  "infrastructure-support": "laa-ccms-platform-tea-aaaaeti4e5egi736k6a4k46qtm@moj.org.slack.com",
+  "owner": "donovan.artz@digital.justice.gov.uk"
+},
+"github-oidc-team-repositories": [""],
+"go-live-date": "2023-05-26"
```
dms1981, line 22: Can we check that this is relevant? I know that's the CCMS-EBS go live date, but I don't think this is directly part of CCMS, more that it's an area for them to do R&D on upgrade activities

SteveLinden: It's not relevant so I will remove it

dms1981: Thank you!

dms1981 (approved):

PR #5164, opened 2023-10-06
give instance management role s3 Put access

drobinson-moj:
As per agreement with Dave Elliot + ProbationWebOps to allow teams with instance management role to more easily add to S3 buckets.


davidkelliott (approved):
Thanks for this :)

PR #5203, opened 2023-10-10
Add some new permissions to the oidc role

georgepstaylor:
Adding these permissions so that we can restore an EFS AWS Backup recovery point to a temporary EFS and then use that to datasync overwrite the contents of a target EFS with the backup.


ep-93 (approved):
No description provided.

Please add a little more detail next time!

georgepstaylor:
> No description provided.
> 
> Please add a little more detail next time!

Have added a description

PR #5216, opened 2023-10-12
create pagerduty integration for csr alarms

robertsweetman:
- new slack channel for CSR alarms
- no distinction between prod and non-prod as all CSR environments are either production or pre-production


dms1981 (dismissed):

markgov (approved):
LGTM

PR #5235, opened 2023-10-13
🛂 Add MWAA User access for Data Platform Apps and Tools

jacobwoffenden:
Resolves https://github.com/ministryofjustice/data-platform/issues/791

Enables access to [`modernisation-platform-mwaa-user`](https://github.com/ministryofjustice/aws-root-account/blob/main/management-account/terraform/sso-admin-permission-sets.tf#L181-L187)

### DIFF
--- a/environments/data-platform-apps-and-tools.json
+++ b/environments/data-platform-apps-and-tools.json
@@ -13,6 +13,11 @@
           "github_slug": "data-tech-archs",
           "level": "sandbox",
           "nuke": "exclude"
+        },
+        {
+          "github_slug": "data-platform-apps-and-tools-airflow-users",
+          "level": "mwaa-user",
+          "nuke": "exclude"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


davidkelliott:
So normally we would have the policy created in the Modernisation Platform repo, I’ve just had a look at the policy in the root account - 
```
data "aws_iam_policy_document" "modernisation_platform_data_mwaa_user" {
  statement {
    actions = [
      "airflow:CreateWebLoginToken"
    ]
    resources = ["arn:aws:airflow:*:*:role/*/User"]
  }
}
```
This is just one statement + read only is attached, can this not be added to one of the existing data engineering roles?  It doesn’t feel like we need a new access level just for this.

davidkelliott (approved):
Approved following a discussion with data platform team https://mojdt.slack.com/archives/C01A7QK5VM1/p1697620199412489

Airflow is that it's made available to analysts who require it for their work, but is not a system everyone will (or should) be interacting with. Given it drives a large amount of ETL in the org we want to minimise the risk of users who shouldn't really be using it being able to make changes that would affect our ETL pipelines that Airflow will need to drive.

If in future we find they need to add a lot more roles we should re-evaluate how platforms on MP access is given.

PR #5236, opened 2023-10-16
Add datasync permissions for oidc role

georgepstaylor:
Adds:
* datasync:List* 
* datasync:TagResource

To the oidc role, so that datasync tasks, locations can be listed and tags can be added to them.


markgov (approved):
LGTM and deployed to sprinkler no issue

PR #5267, opened 2023-10-20
Refrencing CircleCI Secret in Mod Plat.

ep-93:
This PR gets circle ci working with oidc for digital prison reporting. 

Its been applied in development already, and has a corresponding PR that has already been merged to create the secret. - https://github.com/ministryofjustice/modernisation-platform/pull/5312

This secret has been populated and is now referenced in this PR safely.

It's been tested with the following branch here -  https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-reporting-orb/1007/workflows/47c398e3-6a25-4881-947d-dd0bd3f18506/jobs/8056

The permissions are currently enough, and Hari is happy with them, but they can be edited at any point. Currently I'm trying to reduce how open they are.

If more teams request this, we will have to look at a slicker way of doing it!


terraform/environments/digital-prison-reporting/iam.tf
```diff
@@ -0,0 +1,59 @@
...
+        "ec2:RunInstances",
+        "ec2:DescribeInstances",
+        "ec2:TerminateInstances",
+        "ec2:CreateTags",
+        "s3:PutObject",
+        "s3:GetObject",
+        "s3:ListBucket",
+        "iam:PassRole",
+        "elasticbeanstalk:*",
+        "rds:*",
```
davidkelliott, line 33: Is this the same as the CICD user policy?  These permissions seem quite open?

ep-93: They're similar, and actually as this is just for one team currently they could be less, it's hard to get exactly what the team require though, will chase them again.

SteveLinden (approved):
Assuming it's OK elsewhere

PR #5283, opened 2023-10-23
Actions runner controller deployment

jhpyke:
This PR creates the basis of an ARC deployment in the apps-and-tools cluster.


dms1981:
We'd like to pause on these changes to consider this spike: https://github.com/ministryofjustice/modernisation-platform/issues/5286

jacobwoffenden:
This change is no longer required as we deploy self-hosted runners to Cloud Platform, but we're still interested in the spike because we'd ideally like to shift 99% of this infrastructure to MPE

PR #5305, opened 2023-10-27
Add Notify service domain to firewall FQDN

jemnery:
It seems reasonable to have this widely allowed, but there may be implications I'm not aware of.


dms1981:
Having worked with @jemnery on this last week, the Notify service only offers a HTTPS endpoint, so with that in mind I'm going to close this PR. We can revisit it if we need to conduct any further investigation.

PR #5309, opened 2023-10-27
✨ Add docs for `Tags and Releases` in GitHub

connormaglynn:
## 👀 Purpose
To help Modernisation Platform Engineers ensure they are following common practices when tagging and releasing software

## ♻️ What's changed
Added documentation to our `Ways of Working` within the `Git and GitHub` section

## 📝 Notes
- Mainly adding this because I recently tagged and released some software and wasn't 100% sure about the etiquette
- Still unsure if we should create a Draft version first and get this reviewed? 🤔 I left this out because it's not something I believe in. The code itself has already been reviewed and approved at the release stage. Release notes can always be changed after the fact if any issues are spotted 👀 
- I've attempted to keep the advice fairly generic, but the advice has not been agreed upon as a team, so happy for opinions and comments! 💬


dms1981 (approved):
LGTM - we also reference semantic versioning [in our user guide](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/reviewing-dependabot-prs.html#notes-on-reviewing) so the consistency here is great for users.

PR #5312, opened 2023-10-27
Circle CI Secrets.

ep-93:
Copied CP on circle ci secret management.

### DIFF
--- a/terraform/modernisation-platform-account/secrets.tf
+++ b/terraform/modernisation-platform-account/secrets.tf
@@ -87,3 +87,25 @@ resource "aws_ssm_parameter" "modernisation_platform_account_id" {
 
   tags = local.tags
 }
+
+# CircleCI Organisation ID
+resource "aws_secretsmanager_secret" "circleci" {
+  name        = "cloud-platform-circleci"
+  description = "CircleCI organisation ID for ministryofjustice, used for OIDC IAM policies"
+}
+
+resource "aws_secretsmanager_secret_version" "circleci" {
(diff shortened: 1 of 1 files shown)


terraform/modernisation-platform-account/secrets.tf
```diff
@@ -87,3 +87,25 @@ resource "aws_ssm_parameter" "modernisation_platform_account_id" {
...
+# CircleCI Organisation ID
+resource "aws_secretsmanager_secret" "circleci" {
+  name        = "cloud-platform-circleci"
+  description = "CircleCI organisation ID for ministryofjustice, used for OIDC IAM policies"
+}
+
+resource "aws_secretsmanager_secret_version" "circleci" {
+  secret_id = aws_secretsmanager_secret.circleci.id
+  secret_string = jsonencode({
+    organisation_id = "CHANGE_ME_IN_THE_CONSOLE" # change this value manually in the console once the secret is created
```
dms1981, line 100: Is this value derivable through a data call?

ep-93: So I had it hard code before and was calling it from there which worked great, but having it in code is not allowed, had to scrub it as its classed as sensitive. 

This is how cloud platform do it, I'm open for better options though!

dms1981: If there's no better option than, say, a blank string, then it's going to have to suffice 🤷🫠

dms1981 (approved):

PR #5317, opened 2023-10-30
Restrict `modernisation-account-limited-read-member-access`

dms1981:
Because roles are implicitly allowed access to S3 buckets in an account, this PR denies this for the limited read role to prevent unexpected disclosure of objects in S3.


ewastempel (approved):
LGTM

PR #5362, opened 2023-11-02
Removing cicd user from delegate access.

ep-93:
Removed as part of migrating to OIDC, goodbye long lived creds.

Tested on sprinkler.

If any errors occur on the plan / apply it means that someone has re generated keys. I shall remove them and re run the terraform and it will pass.

Although its removing the user, a lot of code will be left for a week in case we need to revert. Changing as little as possible here seems sensible.


ep-93:
`Cannot delete entity, must delete access keys first.`

Just need to fix that first.

SteveLinden (approved):
If they are happy!

PR #5366, opened 2023-11-02
Add Regal for linting Rego

anderseknert:
Hi there! 👋 This PR introduces [Regal](https://github.com/styrainc/regal) for linting the Rego in this project. A few of the linter violations reported have been fixed as part of the PR:

* [non-raw-regex-pattern](https://docs.styra.com/regal/rules/idiomatic/non-raw-regex-pattern)
* [use-assignment-operator](https://docs.styra.com/regal/rules/style/use-assignment-operator)
* [use-some-for-output-vars](https://docs.styra.com/regal/rules/idiomatic/use-some-for-output-vars)

Rules related mostly to style have been disabled in the configuration file also included here, and may be enabled later if desired.

The Conftest version used here was ancient (from 2020!), which prevented using some modern constructs like the `in` keyword. I've had that updated in the CI pipeline, but if there was any reason to use that version, do let me know.

I've also added Regal to the beforementioned workflow to have it run as part of CI.

### DIFF
--- a/.github/workflows/opa-policies.yml
+++ b/.github/workflows/opa-policies.yml
@@ -18,10 +18,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Setup Regal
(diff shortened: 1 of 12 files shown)


ASTRobinson (approved):
LGTM

PR #5384, opened 2023-11-06
Ensure OIDC policy allows required SSM Session actions for Ansible automation

andrewmooreio:
We have multiple GitHub Workflows located [here](https://github.com/ministryofjustice/hmpps-delius-operational-automation) which utilise Ansible to perform certain operational automation tasks. Ansible accesses EC2 instances via SSH over SSM, currently the OIDC role does not permit establishing or managing SSM instance sessions. This PR adds the required permissions.


SteveLinden (commented):
Not sure about the start and terminate session options. Are these really needed? Terminate can remove all settings so we would prefer if these were not needed.

SteveLinden (approved):

PR #5401, opened 2023-11-07
add data platform push-to-catalogue ecr

LavMatt:

Need ecr to hold versions of daap-push-to-catalogue lambda container images

This creates the repository


markgov (approved):
LGTM

PR #5402, opened 2023-11-07
Add ssm:TerminateSession action to OIDC role policy

andrewmooreio:
A follow-up to the previous PR to add various permissions to allow SSH over SSM for our GitHub Workflows:
https://github.com/ministryofjustice/modernisation-platform/pull/5384

The session can be established, however it cannot close cleanly without the `ssm:TerminateSession` action permission. Adding it back in.


markgov (approved):
LGTM

PR #5413, opened 2023-11-08
Removed cicd refrences

ep-93:

Removing reference to the old user that no longer exists.

I think this might fix the apply issue we have in core-shared-services currently.

## How has this been tested?

Plan looks good, applied it locally and it worked, unlike the apply on main right now.

I did it, which is jumping the gun, but main does not currently work.

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #5447, opened 2023-11-10
Correcting incorrect general-incident.html that was in use on these d…

SteveLinden:
…ocuments

Incorrect reference in the two documents.

Replaces the incorrect value with the one we want.

## How has this been tested?

We got a 400 error and now it uses one that will be opened (tested that)

- [x ] All checks have passed
- [ x] I have made corresponding changes to the documentation
- [ x] Plan and discussed how it should be deployed to PROD (If needed)

{Please write here}

### DIFF
--- a/source/runbooks/disaster-recovery.html.md.erb
+++ b/source/runbooks/disaster-recovery.html.md.erb
@@ -58,4 +58,4 @@ Please contact us over at [#ask-modernisation-platform](https://mojdt.slack.com/
 
 ## How will we deal with an incident?
 
-See here our [general incident runbook](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/manage-an-incident.html)
+See here our [general incident runbook](manage-an-incident.html)
--- a/source/runbooks/disaster-recovery.html.md.erb
+++ b/source/runbooks/disaster-recovery.html.md.erb
@@ -58,4 +58,4 @@ Please contact us over at [#ask-modernisation-platform](https://mojdt.slack.com/
 
 ## How will we deal with an incident?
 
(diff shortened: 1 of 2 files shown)


davidkelliott (changes requested):
You don't need to use the full url, you can just use the relative url

source/runbooks/disaster-recovery.html.md.erb
```diff
@@ -58,4 +58,4 @@ Please contact us over at [#ask-modernisation-platform](https://mojdt.slack.com/
 
 ## How will we deal with an incident?
 
-See here our [general incident runbook](general-incident.html)
+See here our [general incident runbook](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/manage-an-incident.html)
```
davidkelliott, line 61:
```suggestion
See here our [general incident runbook](manage-an-incident.html)
```

source/runbooks/dos-attack.html.md.erb
```diff
@@ -15,7 +15,7 @@ DDoS alarms should be configured for all production public facing interfaces, wh
 
 ## Create an incident and follow the incident guidance
 
-Follow the [general incident guidance](general-incident.html) to record the incident and gather information.
+Follow the [general incident guidance](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/manage-an-incident.html) to record the incident and gather information.
```
davidkelliott, line 18:
```suggestion
Follow the [general incident guidance](manage-an-incident.html) to record the incident and gather information.
```

davidkelliott (approved):
Great thanks!

ASTRobinson (approved):

PR #5547, opened 2023-11-22
Add required permissions for managing Step Functions

harichintala1:
reference, https://aws.permissions.cloud/iam/states

We are unable to start, add tags, add actions and other activities with respect to step functions

This will give permissions for managing under sandbox Role

## How has this been tested?

This permission will allow us to test this feature

No

- [x ] All checks have passed

[{Permission Reference Document}](https://aws.permissions.cloud/iam/states)


markgov (changes requested):
this may break can you make an amendment so that it groups all the states statements so something like states:*

markgov (approved):
LGTM

Issue #5548, opened 2023-11-22
Alerting for Secrets Rotation


richgreen-moj:
### User Story

As a Mod Platform Engineer...
I want to be alerted when our secrets need to be rotated as per the [documentation](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/rotating-secrets.html#how-to-rotate-secrets)...
So that I can ensure that secrets are rotated regularly mitigating the risk of long-lived credentials.

### Value / Purpose

This will ensure that our secrets are being rotated as per the agreed schedule by proactively reminding the MP team via an alert channel.
This story was conceived as a result of https://github.com/ministryofjustice/modernisation-platform/issues/5325 

### Additional Information

Thoughts so far are that this could take the form of a script/reminder that checks the age of the secrets and posts to a Slack channel with a reminder for MP team to update them. This could start to alert a week or so before the expiry to allow time to gather new credentials as in some cases this requires contacting another team.

### Definition of Done

Example - [ ] Documentation has been written / updated
- [ ] README has been updated
- [ ] User docs have been updated
- [x] Another team member has reviewed
- [x] Tests are green

dms1981:
Discussed in standup - raising a GitHub issue is probably better than a Slack message that could be easily lost

richgreen-moj:
Here is my WIP workflow: https://github.com/ministryofjustice/modernisation-platform/blob/feature/5548-secrets-rotation-alerting/.github/workflows/secrets-rotation-reminder.yml

Most recent run https://github.com/ministryofjustice/modernisation-platform/actions/runs/7033390229/job/19139110635 highlights secrets that are over the threshold - I've commented out raising a new issue at the moment as I don't want to spam the repo with test issues but have been testing it locally.

Wondering how important it would be to raise new issues following our new story form template although not sure that's currently possible via gh cli https://github.com/cli/cli/issues/5865 - I might just look to replicate the style of our template for the new issue.

richgreen-moj:
Raised https://github.com/ministryofjustice/modernisation-platform/pull/5631 and incorporated comments from the team. I've now merged in to main - I will verify that it kicks off on schedule today (09:30) and is behaving as expected.

richgreen-moj:
I had made an omission on permissions required to raise gh issues so had to raise https://github.com/ministryofjustice/modernisation-platform/pull/5641 to fix it. This also changed the daily schedule to 11:30 just for convenience that I could watch it running in. 

[...]

PR #5631, opened 2023-11-30
Feature/5548 secrets rotation alerting

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/5548

I've written a GH Action that queries the age of the secrets held in the Mod Platform account. Currently it is scheduled to run at **9:30 am daily** (this could be altered as necessary). It will check to see if the secrets are over 180 days old and also whether an existing issue is open to rotate the secret. If necessary it will raise a new issue to rotate the secret.

It will not raise an issue for the secrets we have deemed unnecessary to rotate in the [docs](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/rotating-secrets.html).

## How has this been tested?

I've run the [workflow](https://github.com/ministryofjustice/modernisation-platform/actions/workflows/secrets-rotation-reminder.yml) on my branch but have commented out the issue creation steps so as not to spam our issues list.

I tested out raising issues in my own repo to check the logic and formatting, [here](https://github.com/richgreen-moj/test-github-actions/issues/24) is an example of the kind of issue that will be created.

No

- [ ] All checks have passed

### DIFF
--- a/.github/workflows/secrets-rotation-reminder.yml
+++ b/.github/workflows/secrets-rotation-reminder.yml
@@ -1,9 +1,16 @@
 ---
 name: Secrets Rotation Reminder
 on:
-    workflow_dispatch:
-    schedule: 
-        - cron: '30 9 * * *' # run at 9:30 daily
+  push:
+    branches:
+      - "feature/5548-secrets-rotation-alerting"
+  pull_request:
+    branches:
+      - "main" 
+    paths:
+      - ".github/workflows/secrets-rotation-reminder.yaml"
+  # schedule: 
+  #     - cron: '30 9 * * *' # run at 9:30 daily
 
 env:
     AWS_REGION: "eu-west-2"
--- a/.github/workflows/secrets-rotation-reminder.yml
+++ b/.github/workflows/secrets-rotation-reminder.yml
@@ -10,7 +10,7 @@ on:
     paths:
       - ".github/workflows/secrets-rotation-reminder.yaml"
   # schedule: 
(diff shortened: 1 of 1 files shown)


SteveLinden (approved):

.github/workflows/secrets-rotation-reminder.yml
```diff
@@ -0,0 +1,90 @@
...
+                gh issue create --title "Rotate $secret Credential" --label security --body "The [secrets-rotation-reminder workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) has identified that the $secret credential requires rotation as it is close to or exceeding the threshold of 180 days. 
+
+                Consult [this documentation](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/rotating-secrets.html#how-to-rotate-secrets) which describes the process for rotation."
+                else 
+                echo "The $secret secret has not been identified for rotation (age: $((age / (24 * 60 * 60))) days)"
+                fi
+
+              done
+
+            env:
```
ewastempel, line 87: this is a bit pedantic of me, but I noticed you are defining env vars globally and then in the job itself... you could argue that all of the vars are global or all of them are local to the job...there is only one job, so it probably does not matter much, but what hurts my eyes is env vars are defined in two places :) 
as I said... a bit pedantic of me :D

richgreen-moj: That's OK - this is a symptom of me having borrowed from other jobs and examples! I will tidy that up.

ewastempel (approved):
well done!

Issue #3451, opened 2023-03-16
Automation to provide info for new onboarders


davidkelliott:
## User Story

When a new environment issue has been closed, I want the person raising the issue to be tagged and provided with the following information:

> Welcome to the Modernisation Platform! Your new accounts have now been created.  Please see the user guidance for details on how to build and access infrastructure in the Modernisation Platform - https://user-guide.modernisation-platform.service.justice.gov.uk/#getting-started
>
> If you need any help or assistance please contact us in the [#ask-modernisation-platform](https://mojdt.slack.com/archives/C01A7QK5VM1)

## Value

Provide essential information to users in an automated fashion
## Definition of done

- [x] automation added
- [x] another team member has reviewed
- [x] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

SimonPPledger:
where does this get raised? Which github repo?

SteveLinden:
Discussed with Dave Elliott and this will be set up so it runs once the issue is closed.

SteveLinden:
Sample yaml code created. Testing will require a temporary call that we can close to ensure it works

SteveLinden:
Failed lots of times with a useless message. Appears to be part of the code where it can't find details. Working on it

SteveLinden:
The code now sends a note to the call (see #5640 for details). No email is sent but we are limited in what we can get from github to progress this

SteveLinden:
Result is approved. If needed we can manually email the details to the customer

PR #5682, opened 2023-12-05
Put env under the appropriate location

SteveLinden:

#3451 

To cure outstanding issue with github-token

## How has this been tested?
Tested with act but it doesn't show the error

Does not affect anything else

- [x ] I have performed a self-review of my own code
- [ x] All checks have passed

Will test by moving an item into closed state. It's shown as test to stop anyone else using it


davidkelliott:
It is more secure to set the env var for the token at the appropriate level, eg step or job.  If you set it at the top level you are giving the whole workflow access to the token.  See more info here - https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

davidkelliott (approved):

PR #5787, opened 2023-12-12
Add networking for copilot

sukeshreddyg:

Adding RAM Share for Copilot


markgov (approved):
LGTM

PR #5790, opened 2023-12-12
Update access from developer to sandbox

sukeshreddyg:

Update bedrock access from developer to sandbox for copilot


markgov (approved):
LGTM

Issue #5713, opened 2023-12-06
Timeout accessing MoJ SMTP relay service from Nomis in Modernisation Platform


SimonPPledger:
### Expected Behavior

As reported by Dom Robinson. 
He needs the Nomis accounts in Mod Platform to be able to communicate with  MoJ SMTP Relay 10.180.104.100  and 10.180.105.100 (some more details here - [MoJO-SMTP-Relay](https://github.com/ministryofjustice/staff-infrastructure-smtp-relay-server#readme) .  Here is a link to the conversation in slack (https://mojdt.slack.com/archives/C01A7QK5VM1/p1668610127017869).  We **think** it needs port 25. We thought it had been sorted out, but it is still not working.

We need to find out where it is failing - and potentially raise a request with Tech Services

### Actual Behavior

it is being timed out
10.26.8.134):
nc -vzw 5 10.180.104.100  25
nc: connect to 10.180.104.100 port 25 (tcp) timed out: Operation now in progress

also timing out  from prod 10.27.12.178

dms1981:
I don't think this is specifically a bug, but more a story to allow traffic from MP ranges to the MOJ SMTP relays.
Our definition of done would look like the following:
- [ ] NACLs checked to allow tcp/25 to 10.0.0.0/8
- [ ] MP `external-inspection` firewall configured to allow tcp/25 to `10.180.104.100/32` & `10.180.105.100/32`

SteveLinden:
Service now call raised
Submitted : 12-12-2023 09:09:34
Request Number : REQ1316682
Estimated Delivery : 17-12-2023

SteveLinden:
Initial call (above) cancelled as Simon required more details. Added additional ones to this
Submitted : 12-12-2023 15:28:17
Request Number : REQ1317707
Estimated Delivery : 17-12-2023

SteveLinden:
Added the firewall rules using https://github.com/ministryofjustice/modernisation-platform/pull/5797 so this call can be closed. The above request is unlikely to take place until well after the new year due to the change shutdown over Christmas (starting 15th December)

PR #5797, opened 2023-12-13
New firewall rules to allow access  to Nomis

SteveLinden:

See #5713 

Adds the firewalls for when the ports are opened

## How has this been tested?

Not yet testable

- [ x] All checks have passed

### DIFF
--- a/terraform/environments/core-network-services/cidr-ranges.tf
+++ b/terraform/environments/core-network-services/cidr-ranges.tf
@@ -44,6 +44,8 @@ locals {
     nomisapi-t3-root-vnet          = "10.47.0.0/26"
     nomisapi-preprod-root-vnet     = "10.47.0.64/26"
     nomisapi-prod-root-vnet        = "10.47.0.128/26"
+    nomis-SMTP-Relay1              = "10.180.104.100"
+    nomis-SMTP-Relay2              = "10.180.105.100"
 
     # hmpps aws cidr ranges
     delius-eng-dev  = "10.161.98.0/25"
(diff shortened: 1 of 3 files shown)


dms1981 (changes requested):
Just one more comment from me

terraform/environments/core-network-services/cidr-ranges.tf
```diff
@@ -44,6 +44,8 @@ locals {
     nomisapi-t3-root-vnet          = "10.47.0.0/26"
     nomisapi-preprod-root-vnet     = "10.47.0.64/26"
     nomisapi-prod-root-vnet        = "10.47.0.128/26"
+    moj-smtp-relay1              = "10.180.104.100"
```
dms1981, line 47: Missing the `/32` at the end here

terraform/environments/core-network-services/cidr-ranges.tf
```diff
@@ -44,6 +44,8 @@ locals {
...
+    moj-smtp-relay2              = "10.180.105.100"
```
dms1981, line 48: Also missing the `/32` at the end here.

SteveLinden: Amended and repushed

terraform/environments/core-network-services/cidr-ranges.tf
SteveLinden: Amended and repushed

dms1981 (approved):

PR #5880, opened 2023-12-27
Account creation for data platform compute

sukeshreddyg:

Create new environment for data platform compute

Creates a development and production environments to get access to sagemaker for data platform compute


mikereiddigital (approved):
lg2m

Issue #5859, opened 2023-12-21
Bug - Error: Failed to persist state to backend on modernisation-platform 


SteveLinden:
### Expected Behavior

The state should save without issues. 

### Actual Behavior

We get the above error as seen in https://github.com/ministryofjustice/modernisation-platform/actions/runs/7285740796/job/19855252914#step:7:113 as an example 

### Steps to Reproduce the Problem

Run a full release that amends everything, e.g. adding a role access change. The number that happen is not consistent but it has been happening more recently.

### Version

Example is the run for PR #5840 

### Modules

modernisation-platform

ewastempel:
The error to be fixed:
```
Error: Failed to save state

Error saving state: failed to upload state: operation error S3: PutObject,
failed to rewind transport stream for retry, request stream is not seekable

Error: Failed to persist state to backend

The error shown above has prevented Terraform from writing the updated state
to the configured backend. To allow for recovery, the state has been written
to the file "errored.tfstate" in the current working directory.

Running "terraform apply" again at this point will create a forked state,
making it harder to recover.

To retry writing this state, use the following command:
    terraform state push errored.tfstate
```

dms1981:
https://mojdt.slack.com/archives/C015UBQ78MR/p1702984257007459 << You can see a short Slack conversation with our AWS TAM here where we were given some guidance / linked to the S3 performance design considerations whitepaper.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

ewastempel:
Thanks David, but according to the doc, we are not hitting the limit in a single workflow run. FYI, the implemented solution is not looking into the limits, but is to push the errored state.

ewastempel:
The state push fix for the state persistence failure is now rolled out to the scheduled baseline workflow with temporarily suppression of slack alerts for when the state push is successful.
There will be separate issues to track the fix rollout to other workflows.
Also, once https://github.com/hashicorp/terraform/issues/34528 is fixed, the alerting should be re-enabled.

ewastempel:
To roll out the fix to other repos/workflows see this issue: https://github.com/ministryofjustice/modernisation-platform/issues/6038

PR #5911, opened 2024-01-04
Persisting the state when tf errors out

ewastempel:

This is to fix https://github.com/ministryofjustice/modernisation-platform/issues/5859

`terraform state push errored.tfstate ` is run, when the pipeline fails with the error below and writes a `errored.tfstate` file:

```
Error: Failed to save state

Error saving state: failed to upload state: operation error S3: PutObject,
failed to rewind transport stream for retry, request stream is not seekable

Error: Failed to persist state to backend

The error shown above has prevented Terraform from writing the updated state
to the configured backend. To allow for recovery, the state has been written
to the file "errored.tfstate" in the current working directory.

Running "terraform apply" again at this point will create a forked state,
making it harder to recover.

To retry writing this state, use the following command:
    terraform state push errored.tfstate
```

## How has this been tested?

It was run locally to test the correctness of code when the file exists and when the file does not exist.
The terraform action itself can only be tested live, when a pipeline fails for that error and generates `errored.tfstate` file.
The correctness of the terraform command was also tested locally (with a fake empty `errored.tfstate` file).

The change is done in a bash script of the `terraform apply` script and it will be enrolled to all pipelines in the `modernisation-platform` that run `scripts\terraform-apply.sh` script.

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/scripts/terraform-apply.sh
+++ b/scripts/terraform-apply.sh
@@ -24,3 +24,7 @@ if [ ! -z "$2" ]; then
 else
   terraform -chdir="$1" apply -input=false -no-color -auto-approve | ./scripts/redact-output.sh  
 fi
+
+if [ -f "$1/errored.tfstate" ]; then
+  terraform -chdir="$1" state push errored.tfstate | ./scripts/redact-output.sh
+fi


ep-93 (dismissed):

davidkelliott:
I'm not sure if this would work currently as this script has `set -e` so it will exit immediately on any non zero exit code, so if the apply fails with a none zero code it would exit and not get to the next step.

davidkelliott (approved):

davidkelliott (approved):
Thanks!

PR #5937, opened 2024-01-09
JML lambda execution role and secrets manager

murad-ali-MoJ:
We are working on a JML (Joiners, Movers, and Leavers) project. We have created a lambda function on the data platform, which queries logs from 'data-platform-apps-and-tools-production' and sends emails to specific users using the 'gov-notify' service. We are using these roles and secrets for that lambda function. If needed more information here is the [ticket](https://github.com/orgs/ministryofjustice/projects/27/views/18?pane=issue&itemId=40142613#:~:text=%E2%9A%99%20Automate%20generating%20and%20sending%20JML%20extracts%20monthly%231719)


terraform/environments/data-platform-apps-and-tools/jml_lambda_execution_roles.tf
```diff
@@ -0,0 +1,50 @@
...
+  description = "Policy to access CloudWatch Logs"
+
+  policy = jsonencode({
+    Version = "2012-10-17",
+    Statement = [
+      {
+        Action = [
+          "logs:DescribeLogStreams",
+          "logs:DescribeLogGroups",
+          "logs:GetLogEvents"
```
jhpyke, line 38: I'd expect to see `cloudwatch:GenerateQuery` as a required permission here - otherwise the JML extract will not be able to query the correct logstreams.

terraform/environments/data-platform-apps-and-tools/jml_secrets_manager_roles.tf
```diff
@@ -0,0 +1,44 @@
+# IAM role with a trust policy that allows the Lambda service to assume this role.
+resource "aws_iam_role" "secrets_manager_execution_role" {
```
jhpyke, line 2: I'm not sure I understand why this has been made a separate policy? The lambda is not set up to do multiple role assumption in its current form. If there's a good reason (which I can believe there may be) we'll need to see adjustments to the lambda code to allow it to pick up different roles for the different actions.

jhpyke (changes requested):
Needs at least one permission adding, and would be worth understanding the need for a separate role for `secretsManager` actions.

terraform/environments/data-platform-apps-and-tools/jml_lambda_execution_roles.tf
murad-ali-MoJ: I added cloudwatch:GenerateQuery

sukeshreddyg (approved):

dms1981:
Hi @murad-ali-MoJ - the `aws_secrets` code is fine here if you need the secrets to be durable, but the `aws_iam` resources can be provisioned through the [modernisation-platform-environments](https://github.com/ministryofjustice/modernisation-platform-environments) repository with an approval from the Modernisation Platform team.

Can you move your IAM code across to your environment in Modernisation Platform Environments, and think about if you need the secrets to be durable - if so, they can be moved too.

PR #5938, opened 2024-01-09
New Code scanning tool

markgov:

As a Modernisation Platform Engineer
I want to migrate from tfsec to trivy
So that we continue to benefit from up-to-date secure code analysis

trivy is now being favoured over tfsec, they are still using the same base rules so this isn't urgent but we should still change at some point.
https://github.com/aquasecurity/tfsec/blob/master/docs/guides/trivy.md

This PR adds in code for running trivy in the code scanning workflow

## How has this been tested?

i have tested this code on a different repo and works as intended

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/terraform-static-analysis.yml
+++ b/.github/workflows/terraform-static-analysis.yml
@@ -25,7 +25,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Run Analysis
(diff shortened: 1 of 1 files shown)


Khatraf (dismissed):
lgtm

PR #5942, opened 2024-01-10
added secret for JML

murad-ali-MoJ:

{Please write here}

{Please write here}

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/data-platform-apps-and-tools/secrets.tf
+++ b/terraform/environments/data-platform-apps-and-tools/secrets.tf
@@ -7,25 +7,12 @@ data "aws_ssm_parameter" "modernisation_platform_account_id" {
 resource "aws_secretsmanager_secret" "govuk_notify_api_key" {
   count = terraform.workspace == "data-platform-apps-and-tools-production" ? 1 : 0
 
-  name = "gov-uk-notify/production/api-key"
-}
-
-# Add the Gov.UK Notify API key to the secret
-resource "aws_secretsmanager_secret_version" "gov_uk_notify_api_key_version" {
-  count       = terraform.workspace == "data-platform-apps-and-tools-production" ? 1 : 0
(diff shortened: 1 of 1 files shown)


terraform/environments/data-platform-apps-and-tools/secrets.tf
```diff
@@ -3,6 +3,31 @@ data "aws_ssm_parameter" "modernisation_platform_account_id" {
...
 }
 
+# Create a new secret in AWS SecretsManager for Gov.UK Notify API key
+resource "aws_secretsmanager_secret" "govuk_notify_api_key" {
+  count = terraform.workspace == "data-platform-apps-and-tools-production" ? 1 : 0
+
+  name = "gov-uk-notify/production/api-key"
+}
+
+# Add the Gov.UK Notify API key to the secret
```
jhpyke, line 13: I think we would want to cut the below block where we attempt to populate with a value, otherwise the actual value we wish to set would be overwritten every time someone merges to this codebase

terraform/environments/data-platform-apps-and-tools/secrets.tf
```diff
@@ -3,6 +3,31 @@ data "aws_ssm_parameter" "modernisation_platform_account_id" {
...
+  secret_string = "API_KEY_PLACEHOLDER"
+}
+
+# Email secret for Lambda function
+resource "aws_secretsmanager_secret" "email_secret" {
+  name = "jml/email"
+}
+
+# Add the email value to the secret
+resource "aws_secretsmanager_secret_version" "email_secret_version" {
```
jhpyke, line 26: Same as above

jhpyke (changes requested):
We just want to create some blank secrets here, and then populate the underlying values via the console.

jhpyke (approved):
Looks good to me, but up to MP to judge at this point

Issue #5908, opened 2024-01-04
Seperating github CODEOWNERS and AWS access


georgepstaylor:
### User Story

As a team responsible for infrastructure in an environment 
I would like to be able to exclude codeowner capabilities from a github team listed as having access in the `environment.json` file
So that certain teams can have high levels of access via the console/cli and raise PRs to make code changes, but are unable to merge without approval of the "environment owner"

### Value / Purpose

Allows an "environment owner" to decide which teams are able to merge code into the environment with/without an approval from the "environment owner".

Useful in situations where we have multiple teams working within an environment and it is helpful for them to have AWS access (eg App teams, DBA teams etc), but ideally PRs should be reviewed by the infrastructure team

### Useful Contacts

Probation WebOps

### Proposal / Unknowns

Could possibly add a codeowner true/false flag to a teams access definition 
```json
"access": [
        {
          "github_slug": "hmpps-migration",
          "level": "sandbox",
          "codeowner": "false"
        },
```

### Definition of Done

Intended as a discussion for now

SteveLinden:
Sent a slack message to confirm if this is just for one team or a more global thing. Suspect it's one team based on the information provided.

dms1981:
This is done through the `provision-member-directories.sh` script.

georgepstaylor:
Thanks @dms1981. Thoughts on PR #5951 as a potential solution?

SteveLinden:
Above has been applied along with the change to the json file for delius-core

SteveLinden:
Ran release #5957 to make the changes

SteveLinden:
The release above generated a new PR on the environments file. Once this was ported it has restricted the access to approve PRs. 
Once we have confirmed it works correctly it may be worth notifying others who may have this issue.

SteveLinden:
This is working now. I need to update the documentation and let the use community know this is available

dms1981:
Documentation has been updated

SteveLinden:
Updated the documentation and sent a slack message to let our community know about the facility. We may get people requesting this.

PR #5957, opened 2024-01-11
Amended the delius-code json to add codeowner false to the dba group

SteveLinden:

As per the request info on #5908 
Additionally another change has been made to the provision-members-directories.sh file too

Should prevent those with codeowners:false from approving PRs

## How has this been tested?

Difficult to test so will be done by approving this and hoping it runs the above sh

- [x ] I have performed a self-review of my own code
- [x ] All checks have passed


mikereiddigital (approved):
lg2m

PR #6053, opened 2024-01-24
Test of new trivy scan

markgov:

New functionality has been added to the trivy scanning now and we need to test that it is working as expected also i have removed the ignore_falures: true as this was ignoring all errors and producing a false positive

Add back in trivy functionality and also adds in the new trivy ignore file that will be use to ignore trivy errors that we want to ignore

## How has this been tested?

This has been tested locally on my laptop with a local install of trivy against the modernisation platform repo

- [ ] All checks have passed

{Please write here}


SteveLinden (approved):
Assuming this sorts out the ignores

dms1981 (approved):

Issue #5449, opened 2023-11-10
Moving Instance scheduler to event bridge scheduling


markgov:
### User Story

As a Modernisation platform engineer
I need to look at how the instance scheduler lambda is executed 
So that it runs at the correct time no matter what time of the year it is.

### Value / Purpose

Currently the instance scheduler lambda only runs in UTC time zone and there is no way of changing this with the current event bridge rule, but aws has implemented a new service called event bridge scheduler which has much more functionality. 

### Additional Information

https://docs.aws.amazon.com/eventbridge/latest/userguide/using-eventbridge-scheduler.html
terraform documentation
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/scheduler_schedule

### Unknowns Potential pitfalls that could cause the story to expand beyond its original scope. Ideally this section will remain blank.
This is new code and has not been tested

### Definition of Done

Example 
- [ ] Documentation has been written / updated
- [x] New code added and tested
- [ ] README has been updated
- [x] Another team member has reviewed
- [x] Tests are green

sukeshreddyg:
Successfully migrated **instance-scheduler-lambda-function** from CloudWatch Events to EventBridge Scheduler

#6063

PR #6054, opened 2024-01-24
IAM role and policy for lambda invoke

sukeshreddyg:

IAM role and policy to invoke lambda functions #5449 

Event bridge scheduler to trigger lambda functions

### DIFF
--- a/terraform/environments/core-shared-services/iam.tf
+++ b/terraform/environments/core-shared-services/iam.tf
@@ -295,7 +295,7 @@ resource "aws_iam_role" "lambda_invoke_role_policy" {
   })
 }
 
-resource "aws_iam_role_policy_attachment" "example_attachment" {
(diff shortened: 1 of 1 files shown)


terraform/environments/core-shared-services/iam.tf
```diff
@@ -262,3 +262,42 @@ module "ssm-cross-account-access-admin" {
...
+        "Principal" : {
+          "AWS" : "arn:aws:lambda:eu-west-2:${local.environment_management.account_ids["core-shared-services-production"]}:function:*"
+        },
+        "Action" : "sts:AssumeRole"
+      }
+    ]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "example_attachment" {
```
ASTRobinson, line 298: should 'example_attachment' be replaced with something more descriptive?

ASTRobinson (approved):

PR #6063, opened 2024-01-25
Moving Instance scheduler to event bridge scheduling

sukeshreddyg:

Moving Instance scheduler to event bridge scheduling #5449

Replaces Cloudwatch Events with EventBridge Scheduling, providing enhanced scheduling capabilities for InstanceScheduler lambda function.


ewastempel (approved):
That looks good @sukeshreddyg  🙌

PR #6065, opened 2024-01-25
adding in preprod and prod rules from planet FM

haitchison:

Including a FW rule that allows traffic from DOM1 domain controllers (10.172.68.146/29) on port 9389 to private subnets, specifically for Planet FM preprod and prod. 

New Preprod and Prod firewall rule, with domain-controller cidr range

## How has this been tested?

Firewall logs indicate that this traffic is currently blocked, will re-test by running connection tests from prod / preprod planet servers.

It will add an additional allow rule to the MP firewall

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/preproduction_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/preproduction_rules.json
@@ -456,8 +456,8 @@
   },
   "dom1_dcs_to_planetfm_preproduction_9389_tcp": {
     "action": "PASS",
-    "source_ip": "${dom1-domain-controllers}",
-    "destination_ip": "${hmpps-preproduction-general-private-subnets}",
+    "source_ip": "${hmpps-preproduction-general-private-subnets}",
+    "destination_ip": "${dom1-domain-controllers}",
     "destination_port": "9389",
     "protocol": "TCP"
   }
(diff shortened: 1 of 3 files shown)


markgov (approved):
LGTM

terraform/environments/core-network-services/firewall-rules/preproduction_rules.json
dms1981, line 1: Hi @haitchison! I think you're actually after this:
```
    "source_ip": "${hmpps-preproduction}",
    "destination_ip": "${dom1-domain-controllers}",
```
As I understand the flow, the source will be your instances in the `hmpps-preproduction` VPC. Also the local that generates the platform CIDR ranges produces them as `$business_unit-$environment` so I've trimmed the VPC name down a bit in my example.

terraform/environments/core-network-services/firewall-rules/production_rules.json
dms1981, line 1: As above - I think you're actually after this:
```
    "source_ip": "${hmpps-preproduction}",
    "destination_ip": "${dom1-domain-controllers}",
```

haitchison: Updated! :)

dms1981 (approved):

PR #6080, opened 2024-01-30
Allow all AWS Backup actions whilst assuming the sandbox role

georgepstaylor:

Currently unable to manage AWS Backups in a sandbox account

We manage custom backup plans for certain resources and in the sandbox account where we have been testing the creation of backup vaults we are unable to remove them with terraform because the vaults are not empty. We are also unable to remove the recovery points in the console or via the cli because the current IAM policy doesn't allow any actions other than starting a restore job. In other environment types, we'd be happy to raise an ask request, but in sandbox it feels appropriate that we could do this ourselves.

Adds wildcard iam perms to allow all actions

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


ep-93 (approved):
Approved!

Issue #4999, opened 2023-09-13
investigate CKV_AWS_149 chekov warning suppression - in the affected workflows


ewastempel:
## User Story

Revisit fixing CKV_AWS_149 chekov warning that is currently suppressed.
See the workflows that were affected:
terraform-static-analysis.yml : https://github.com/ministryofjustice/modernisation-platform/actions/runs/5923065434/job/16058039084

```
Check: CKV_AWS_149: "Ensure that Secrets Manager secret is encrypted using KMS CMK"
	FAILED for resource: aws_secretsmanager_secret.rds_db_credentials
	File: /secrets.tf:19-22
	Guide: https://docs.paloaltonetworks.com/content/techdocs/en_US/prisma/prisma-cloud/prisma-cloud-code-security-policy-reference/aws-policies/aws-general-policies/ensure-that-secrets-manager-secret-is-encrypted-using-kms.html

		19 | resource "aws_secretsmanager_secret" "rds_db_credentials" {
		20 |   name                    = "tactical-products-db-secrets"
		21 |   recovery_window_in_days = 0
		22 | }
```

code-scanning.yml: https://github.com/ministryofjustice/modernisation-platform/actions/runs/5921047939/job/16053081949
Additionally add ticket to update all pipelines to use slack notifications when a pipeline fails.

if sensible then fix these.

Is it possible to use a custom KMS keys in this instance?

## Value

Improving security posture.

## Definition of done

- [x] another team member has reviewed
- [x] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

dms1981:
Applied KMS key to relevant resources, removed Checkov and Tfsec exclusions.

PR #6093, opened 2024-01-31
Use KMS key to encrypt MP account secrets, removed tfsec and checkov exclusions

dms1981:

https://github.com/ministryofjustice/modernisation-platform/issues/4999

Adds a KMS key, and uses this KMS key to encrypt the secrets in the MP account

## How has this been tested?

Using the same policy as used in other MP applications of KMS, this KMS key will be usable freely by principals in the MP account, and to decrypt resources by MP OU members

If the KMS policy is too restrictive, there may be an interruption to secrets access while it is relaxes

- [x] All checks have passed


ewastempel (approved):
looks good

PR #6116, opened 2024-02-01
Update accessing-ec2s.html.md.erb: filter to running instances

georgepstaylor:

Hit a strange edge case tonight where after manually editing the ASG for the linux_bastion to bring an instance back online after 9pm, even though the old instance was instance state Terminated, the aws cli was still returning it (and it was visible in the console). 

The current proxy command grabs the first instance only

Adding a second filter to find instances with status code 16 means only "running" instances will be returned

## How has this been tested?

Tested the output of the command manually and have updated my bastion configuration already.

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

Issue #3053, opened 2023-02-02
Move things out of delegated access that don't need higher credentials


davidkelliott:
## User Story

Move everything that doesn't need the root account access from delegate access to member bootstrap

Everything apart from this - 
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/iam.tf#L12
Things relating to this and also this:
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/policies.tf

Everything else should be able to be moved.
## Definition of done

- [ ] readme has been updated
- [ ] user docs have been updated
- [ ] another team member has reviewed
- [ ] tests are green
- [ ] UR test OR added to continual research plan

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

SimonPPledger:
what can be taken out - need this information to consider it 'refined'

davidkelliott:
Everything apart from this - 
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/iam.tf#L12
Things relating to this and also this:
https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/delegate-access/policies.tf

Everything else should be able to be moved.

ep-93:
Needs to be tested, most stuff can be deleted and recreated, some things might need to import..

sukeshreddyg:
1. Migrated following resources from **delegate access** to **member bootstrap**

	* **Collaborators roles** 
 	* **GitHub-oidc role**
 	* **AWSRTSSupport role**
 	* **iam account alias resource**
 	* **SSM Parameters**
 	
2. Migrated **All policies** from delegate access to single-sign-on

#6144 
#6156 
#6244 
#6254

PR #6144, opened 2024-02-06
Move SSM parameters from delegate to member bootstrap access

sukeshreddyg:

#3053 

This PR moves SSM parameters from delegate to member bootstrap access by deleting the SSM parameters from delegate access and recreating them in the member bootstrap access. 

## How has this been tested?

deployed to the sprinkler


davidkelliott (approved):

davidkelliott (commented):
I think this is needed as we only want to create these resources in member accounts

ewastempel (approved):
LGTM

Issue #6048, opened 2024-01-23
Failure to run terraform plan/apply for RAM share due to terraform version


sukeshreddyg:
### Expected Behavior

**Terraform plan/apply works with the 1.5 versions but it doesn't work when using terraform versions greater than 1.6** 

![Screenshot 2024-01-23 at 16 59 27](https://github.com/ministryofjustice/modernisation-platform/assets/152981374/7e56d0c7-6c48-42c7-b269-0de57a4bf346)

Please see this action that has been successfully executed

https://github.com/ministryofjustice/modernisation-platform-environments/actions/runs/7624736725

### Actual Behavior

Example error when using terraform version 1.7 and also getting the same error when using version 1.6

![Screenshot 2024-01-23 at 16 54 21](https://github.com/ministryofjustice/modernisation-platform/assets/152981374/8026196c-389c-4229-8113-7b938413aecd)

### Steps to Reproduce the Problem

Please see this action that is failing as an example

https://github.com/ministryofjustice/modernisation-platform-environments/actions/runs/7624736725/attempts/1

dms1981:
Looking at this one in conjunction with https://github.com/ministryofjustice/modernisation-platform/issues/6112

dms1981:
A quick way to solve this could be to run two `terraform apply` steps, looking something like this:
```
./scripts/terraform-apply.sh $basedir/$application --target=data.aws_subnets.associated
./scripts/terraform-apply.sh $basedir/$application
```
Alternatively, I could statically define all the attributes I want to use in `data.aws_subnet.host` like so:
```
data "aws_subnet" "host" {
  provider = aws.share-host
  for_each = {
    "subnet-0" = data.aws_subnets.associated.ids[0]
    ...
    "subnet-8" = data.aws_subnets.associated.ids[8]
  }
  id       = each.value

  tags = {
    Name = "${var.vpc_name}-${var.subnet_set}*"
  }
}
```

sukeshreddyg:
A similar issue that was fixed previously is happening again.https://github.com/ministryofjustice/modernisation-platform/issues/6048

PR #6151, opened 2024-02-07
Resolve failure to retag RAM shared subnets

dms1981:

https://github.com/ministryofjustice/modernisation-platform/issues/6048

When we run the `./member-account-ram-association.sh` script, we encounter a problem where we want to apply tags to the RAM shared objects corresponding to VPC subnets.

To add the tags we need a full list of subnet IDs, but this isn't available when we run a Terraform plan. To get this list of subnet IDs we either need to use a statically defined map, or we need to run Terraform in a targeted fashion to get those subnet IDs.

This PR amends the `./member-account-ram-association.sh` script so that it first runs a targeted apply, then a full apply.

## How has this been tested?

Tested locally

No, but there could be unexpected behaviour in a situation where `networking.auto.tfvars` is unpopulated.

- [x] All checks have passed

I'm still not convinced this is the right way to go, but without reworking how we retag the shared subnets - for example, passing the subnet information into the module and refactoring the templates - this is the lightest touch solution.

### DIFF
--- a/scripts/member-account-ram-association.sh
+++ b/scripts/member-account-ram-association.sh
@@ -22,7 +22,9 @@ setup_ram_share_association() {
     
     if [[ $select_workspace ]]; then
 
-      # Run terraform apply
+      # Run terraform apply to get subnet info
+      ./scripts/terraform-apply.sh $basedir/$application -target=module.ram-ec2-retagging[0].data.aws_subnets.associated
+      # Run the full terraform apply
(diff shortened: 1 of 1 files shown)


SteveLinden (approved):

davidkelliott (approved):

dms1981:
> Did you test the script to see if this approach works? Thanks

I did, yes! I had to run the script locally, but it was successful against both of the new `cortex-xsiam` accounts.

Issue #6109, opened 2024-02-01
Documentation: Revoking User Access


ewastempel:
### User Story

As an MP engineer
I need to understand how to block or limit user access in a swiftly manner
So that I can mitigate against malicious actions.
This is just to produce documentation as to what we would have to do for different scenarios.

### Value / Purpose

When a cyber attack occurs it can be difficult to know straight away what actions need to be taken.
This ticket is to prepare the team for such event by producing a documentation on what needs to be done to block user access (individual, team, automation/repo, etc.)

### Additional Information

This should include a list of code areas/ resources that would need amending and it should explain what each change means in practice. It should also break it down into specific scenarios.

### Proposal / Unknowns

This ticket should produce documentation(s) in the User Guide and/or README files.

### Definition of Done

- [ ] Team runbook updated including scenarios and potential actions
- [ ] docs succesfully uploaded to github
- [ ] Another team member has reviewed

richgreen-moj:
I have a draft PR https://github.com/ministryofjustice/modernisation-platform/pull/6181 for a new "revoking user access" runbook I am currently writing.

PR #6190, opened 2024-02-12
Initial version of documentation for Revoking User Access #6109

SteveLinden:

Initial list of items to change for #6109 

Ways of stopping access

### DIFF
--- a/source/runbooks/disabling-network-connections-if-needed.html.md.erb
+++ b/source/runbooks/disabling-network-connections-if-needed.html.md.erb
@@ -1,0 +1,50 @@
+---
+owner_slack: "#modernisation-platform"
+title: Creating VPCs in AWS
+last_reviewed_on: 2024-02-12
+review_in: 6 months
+---
+
+# <%= current_page.data.title %>
+
+## Overview
+There may be occassions where there is a network attack on modernisation platform environments that could require a way of stopping access. This document lists one solution to the issue.
+
+## Establish if there is an attack or unwanted access
+Establish if there is unwanted access or an attack taking place by checking the details on cloudwatch in both core-network-services and core-logging. It should be obvious from the contents of the logs.
+
+On core-network-services examine the contents, for production services, of  
+    * fw-live-data-inline-inspection-logs-ulpiqfaq
+    * live_data-vpc-flow-logs-qoa52v4z
+    * external-inspection-flow-logs
+or the following for non-production services
+    * non_live_data-vpc-flow-logs-fro8eqat
+    * fw-non-live-data-inline-inspection-logs-ks3lnpsk
+
+On core-logging examine the contents, for production services, of 
+    * live_data-vpc-flow-logs
+or the following for non-production services
+    * non_live_data-vpc-flow-logs
+
+## If an issue has been established
+You need to then logon to the environment for which there is an issue
+    1. Connect to the service involved.
+    2. INITIALLY detatch the environment from the VPC - _NOTE THIS IS LIKELY TO CAUSE OTHER_ issues
+    3. Check the log again following the steps above.
+
+## Detatching the VPC didn't work
+In most cases removing the VPC from the environment will stop connections but it is a drastic step. If this didn't work follow these steps
+
+To stop connections you will need to access one of the core-vpc environments through core-vpc-production, core-vpc-development, core-vpc-test, core-vpc-preproduction, core-vpc-sandbox.
+
+Connect to the appropriate environment (so for production go to core-vpc-production) and
+    1. Select the Subnet name under "VPC"
+    2. Pick the actions button and block the offending IP addresses through amending the route table. _THIS WILL IMPACT ALL ENVIRONMENTS CONNECTED TO  THIS SUBNET SO TAKE CARE._
+Do this for all subnets :
+    * general-data
+    * general-private
+    * general-public
+    * protected
+    * transit-gateway
+The subnet names will be proceeded by the service name (e.g. hmpps-development-) and followed by 1 of -eu-west-2a, -eu-west-2b, -eu-west-2c. So a name may be hmpps-development-general-private-eu-west-2b.
+


source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,106 @@
...
+title: Disabling network access if required
+last_reviewed_on: 2024-02-14
+review_in: 6 months or when is resolved.
+---
+
+# <%= current_page.data.title %>
+
+## Overview
+
+There may be occassions where there is a network attack on modernisation platform environments that could require a way of stopping access. This document lists teo possible solutions to the issue.
```
davidkelliott, line 12:
```suggestion
There may be occasions where there is a network attack on Modernisation Platform environments that could require a way of stopping access. This document lists two possible solutions to the issue.
```

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,106 @@
...
+
+[NOTE]
+
+> All the steps below assume you can access the system through the AWS console and do not require changes to the terraform code.
+
+## Establish if there is an attack or unwanted access
+
+Establish if there is unwanted access or an attack taking place by checking the details on cloudwatch in both core-network-services and core-logging. It should be obvious from the contents of the logs.
```
davidkelliott, line 20: I'm not sure it is obvious, how would people know what an attack looks like?  Is that something we want to go into here or should this doc be about how to shut down network access rather than how to identify an attack?

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,106 @@
...
+[core-network-services sample](../images/core-network-services.png)
+
+[core-logging sample](../images/core-logging.png)
+
+## If an issue has been established
+
+You need to then logon to the environment for which there is an issue
+
+1. Connect to the service involved.
+2. INITIALLY detatch the environment from the VPC.
```
davidkelliott, line 51: Which environment from which VPC?

davidkelliott (commented):
A good start, this looks to be focused on how to stop internally routed traffic, how would we prevent traffic from attacking an external service?

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,106 @@
...
+
+# <%= current_page.data.title %>
+
+## Overview
+
+There may be occassions where there is a network attack on modernisation platform environments that could require a way of stopping access. This document lists teo possible solutions to the issue.
+
+[NOTE]
+
+> All the steps below assume you can access the system through the AWS console and do not require changes to the terraform code.
```
ewastempel, line 16: I think we would want to explore all ways of blocking access, so not just through the console, but also through code (which in some cases would be quicker than through the console)

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,106 @@
...
+## If an issue has been established
+
+You need to then logon to the environment for which there is an issue
+
+1. Connect to the service involved.
+2. INITIALLY detatch the environment from the VPC.
+
+[CAUTION]
+
+> **NOTE THIS IS DRASTIC AND LIKELY TO CAUSE OTHER issues.**
```
ewastempel, line 55: what other issues? the impact/risk needs documenting in more details

source/runbooks/disabling-network-connections-if-needed.html.md.erb
SteveLinden: More about how to stop network access than detailing what constitutes an attack. I couldn't actually find any examples which showed an attack (or unwanted access for that matter) so it's difficult to show examples here.

source/runbooks/disabling-network-connections-if-needed.html.md.erb
SteveLinden: It does mention that above the line. I will include an example to make it clearer.

source/runbooks/disabling-network-connections-if-needed.html.md.erb
SteveLinden: I did notice on place where that may be quicker and made a note about it. I have not, however, included the details and location on what would need to change. I will look into this to make sure it can be included.

source/runbooks/disabling-network-connections-if-needed.html.md.erb
SteveLinden: Made a note. It will stop all applications using the VPC from accessing the platform. This is because we have VPCs that just attach to the like of hmpss so everything on hmpps will become inaccessible.

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,99 @@
...
+- live_data-vpc-flow-logs
+  or the following for non-production services
+- non_live_data-vpc-flow-logs
+
+## If an issue has been established
+
+You need to then logon to the environment for which there is an issue. For example if it is on NOMIS Combined Reporting Preproduction (nomis-combined-reporting-production) then connect to the core-vpc-preproduction. In this case it would be the hmpps VPC.
+
+1. Connect to the service involved.
+2. INITIALLY detatch the environment from the VPC using Actions and Delete VPC. It might not be possible to delete this as there will be networked interfaces attached which need to be noted and deleted. In this case amending the code may be more appropriate. In the above example there are 166 preproduction network interfaces to be removed.
```
ewastempel, line 44: Deleting any VPC is not desired.

SteveLinden: Maybe put that as the last - if everything else fails - option?

ep-93 (changes requested):

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,150 @@
...
+
+There are 3 options on this
+
+- live_data
+- non_live_data
+- external
+
+Pick the area you want to access - select the Trnasit route table ID :
+
+1. if you know the item you want select the tick box - otherwise **check**
```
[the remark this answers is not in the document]
SteveLinden: Capitalisation was based on the actual statements in AWS rather than me randomly adding them. Maybe I should put quotes around these bits?

source/runbooks/disabling-network-connections-if-needed.html.md.erb
```diff
@@ -0,0 +1,150 @@
...
+title: Disabling network access if required
+last_reviewed_on: 2024-02-14
+review_in: 6 months or when is resolved.
+---
+
+# <%= current_page.data.title %>
+
+# Overview
+
+There may be occasions where there is a network attack on Modernisation Platform environments that could require a way of stopping access. This document lists the possible solutions to the issue.
```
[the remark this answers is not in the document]
SteveLinden: Changed this

source/runbooks/disabling-network-connections-if-needed.html.md.erb
ewastempel: this is still not addressed

PR #6194, opened 2024-02-13
removed invalid `path` declaration

dms1981:
Fixes https://github.com/ministryofjustice/modernisation-platform/actions/runs/7882878946

Paths are supported in trivyignore, but are done per vulnerability at present.


connormaglynn (approved):
🚀 Looks Good To Me!

PR #6196, opened 2024-02-13
Revert "Migrate policies and collaborator roles from delegate access"

sukeshreddyg:
Reverts ministryofjustice/modernisation-platform#6191


sukeshreddyg:
There is a circular dependency between jobs. I need to manually stop the workflow once delegate access is successful because other jobs are deleting the resources created during delegate access. This is due to the revert operation.

markgov (approved):
LGTM

PR #6200, opened 2024-02-13
Rewriting powrbi policy to narrow it down to specific permissions.

julialawrence:

https://github.com/ministryofjustice/data-platform/issues/1916 in Analytical Platform backlog

PowerBI user needs athena and glue access so amending the policy to include them without resorting to  blanket permissions

## How has this been tested?
We will be testing this out with users. Aside from that, a self-review

Should not do. This is a standalone policy only applicable to AP accounts

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #6220, opened 2024-02-14
Updated developer policy

sukeshreddyg:

Updated developer policy to use resource based access for deleting specific suppressed email address.

{Please write here}

Updated developer policy


dms1981 (approved):

dms1981:
From discussion [here](https://mojdt.slack.com/archives/C013RM6MFFW/p1707922691955279) we think this is a reasonable bespoke administrative action that may not be suitable for applying through CI/CD

Issue #5970, opened 2024-01-12
Solution for migrating HMPPS domain from azure


drobinson-moj:
### User Story

As a DSO tech lead
I need to create some EC2s instances directly in the Modernisation Platform Live and Non-Live data VPCs
So that I can migrate domain controllers from Azure FixNGo without running into environment isolation problems.

### Value / Purpose

Allows us to decommission the Azure FixNGo platform in the longer term.  In the short term can also avoids domain controller traffic traversing the AWS - Azure VPN in theory improving performance of domain reliant services (Remote Desktop, Planet FM and CSR/MyDetail)

### Useful Contacts

Dominic Robinson; David Sibley; David Elliot

### Additional Information

Background can be found in https://dsdmoj.atlassian.net/browse/DSOS-2415

Following discussion with Dom Robinson/David Sibley/David Elliot, we agreed tactical solution to solve migration of Azure FixNGo domains while both PlanetFM and CSR/MyDetail applications exist, hence creation of this ticket.

The solution falls outside of the normal architecture since the AZURE.noms.root domain spans both development and test environments, and the HMPPS.noms.root domain spans both pre-production and production.  Creating Domain Controllers (DCs) in development/test/preproduction/production won't work as they need to sync with each other. 

We need the ability to create resources directly in the Live / Non-Live VPCs to avoid the environment isolation issue.  The ask is for a pipeline and associated terraform infra to allow us to do this .  Core-shared-services potentially.  The actual terraform for creating the instances will be managed by DSO team so outside of the scope for this ticket.

### Proposal / Unknowns

Proposal: EC2 instances are built in Live and Non-Live data.

### Definition of Done

- [ ] Solution has been communicated with DSO team [Dominic Robinson]
- [ ] DSO team have verified they can run terraform using the given solution

SimonPPledger:
Check to see whether this is via an AMI or a new build

drobinson-moj:
Re AMI - plan to use our `hmpps_windows_server_2022` existing AMI image and configure the DC tools using powershell (hopefully automating as much as we can)

dms1981:
Talked this one through with @drobinson-moj via Slack. The short version is that code for new instances would be placed [here](https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/core-shared-services), and picked up and run by [this action](https://github.com/ministryofjustice/modernisation-platform/blob/main/.github/workflows/core-shared-services-deployment.yml).

[...]

PR #6224, opened 2024-02-15
DSOS-2615: initial terraform for fixngo domain controllers

drobinson-moj:

In summary, tactical solution for migrating Domain Controllers from Azure FixNGo estate.
Initial commit will standup roles and security group. EC2s will be commented in later

Skeleton terraform to allow creation of EC2s in live and non-live data as per agreed tactical solution.

## How has this been tested?

Tested as far as possible in environments repo in a member account

No

- [x] All checks have passed

Mod Platform engineers aware of solution: David Sibley + Dave Elliot

### DIFF
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -709,8 +709,7 @@ resource "aws_instance" "ad_fixngo" {
 
   root_block_device {
     encrypted = true
-    ##CSS kms_key_id  = module.kms["hmpps"].kms_key_id [OUTPUT DOESN'T EXIST]
-    kms_key_id  = module.environment.kms_keys["general"].arn
+    kms_key_id  = module.kms["hmpps"].key_ids["ebs"]
     volume_size = 127
     volume_type = "gp3"
 
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -145,8 +145,8 @@ locals {
               "kms:RevokeGrant"
             ]
             resources = [
-              module.kms["hmpps"].key_arns["ebs"].arn,
-              module.kms["hmpps"].key_arns["general"].arn,
+              module.kms["hmpps"].key_arns["ebs"],
+              module.kms["hmpps"].key_arns["general"],
             ]
           },
           {
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -1,0 +1,787 @@
+# azure.noms.root and azure.hmpp.root AD DS infra migrated from Azure FixNGo
+# For more details why this is here, please see: 
+# - https://github.com/ministryofjustice/modernisation-platform/issues/5970
+# - https://dsdmoj.atlassian.net/wiki/x/3oCKGAE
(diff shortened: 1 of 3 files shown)


terraform/environments/core-shared-services/ad-fixngo.tf
```diff
@@ -0,0 +1,786 @@
...
+        ingress = {
+          all-from-self = {
+            port     = 0
+            protocol = -1
+            self     = true
+          }
+          dns-udp-53 = {
+            port        = 53
+            protocol    = "UDP"
+            cidr_blocks = ["10.0.0.0/8"]
```
ewastempel, line 211: can the CIDR ranges to be narrowed down?

terraform/environments/core-shared-services/ad-fixngo.tf
```diff
@@ -0,0 +1,786 @@
...
+            from_port   = 49152
+            to_port     = 65535
+            protocol    = "TCP"
+            cidr_blocks = ["10.0.0.0/8"]
+          }
+        }
+        egress = {
+          all = {
+            port        = 0
+            protocol    = -1
```
ewastempel, line 316: this is a bit open

terraform/environments/core-shared-services/ad-fixngo.tf
```diff
@@ -0,0 +1,786 @@
...
+  for_each = local.ad_fixngo.aws_instances
+
+  ami                    = data.aws_ami.ad_fixngo[each.value.ami_name].id
+  availability_zone      = each.value.availability_zone
+  ebs_optimized          = true
+  iam_instance_profile   = aws_iam_instance_profile.ad_fixngo[each.key].name
+  instance_type          = each.value.instance_type
+  private_ip             = each.value.private_ip
+  subnet_id              = each.value.subnet_id
+  user_data              = base64encode(file("./files/ad-fixngo-ec2-user-data.yaml"))
```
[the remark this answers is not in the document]
drobinson-moj: Not really because the user-data script will join the server to the domain.  It's actually much less hassle overall to use base image here.  The user-data just calls a powershell script in modernisation-platform-configuration-management repo in similar way to ansible on linux instances.

terraform/environments/core-shared-services/ad-fixngo.tf
drobinson-moj: Rules here are mirroring the domain controller NSGs in Azure.  The Domain Controller ports allow 10/8 - so any end user device connected onto our network can talk to the DC, and also the DOM1 DCs can reach as well for the forest trust.  Outbound we allow everything to keep things simple - in fact, this is what we have for all our EC2s in all our applications.  Assumption is there is another layer of control, e.g. firewall, gating the outbound access to internet.

ewastempel: I think the firewall rules are a bit open actually, but let me check.

terraform/environments/core-shared-services/ad-fixngo.tf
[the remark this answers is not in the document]
ewastempel: Yep, fair enough,

terraform/environments/core-shared-services/ad-fixngo.tf
ewastempel: Just checked and they are not as open as these SG :) I think there is enough balance there though.

ewastempel (approved):
LGTM

PR #6233, opened 2024-02-15
Equip PowerBi Firewall FQDN Rule

ASTRobinson:
As part of the current Equip upgrade work including a new power bi server, this PR adds a fqdn firewall rule for `.servicebus.windows.net` as per requirements detailed here https://learn.microsoft.com/en-us/data-integration/gateway/service-gateway-communication#ports

I have also sorted the list alphabetically 

TLDR: add servicebus.windows.net to fqdn rules and sort alphabetically.


markgov (approved):
LGTM

Issue #5947, opened 2024-01-10
Create tests / warning around high risk changes


davidkelliott:
### User Story

As an MP engineer
I want to be alerted to or prevented from making changes that may break the platform
So that I don't break the platform.

### Value / Purpose

There are certain changes we know should not be made, eg changing a json file name for an application, an others which are high risk, eg making any changes to the application json file.

By preventing or warning on the PR for these changes we reduce the risk of them accidentally getting past PR review and going through.

### Proposal A proposal that is something testable, don't worry whether it works or not, it's a place for ideas.
Add to the OPA tests changes that we don't want to be made.
Add a check which will warn if changes are high risk by posting a comment on the PR, along with any additional information which may help a reviewer make a decision.
Going forward we should then add additional tests and warning every time we have an incident which could have been prevented.

### Definition of Done

- [x] OPA tests updated
- [x] New warning tests which post to the PR created
- [x] Documentation has been written / updated
- [x] README has been updated
- [x] User docs have been updated
- [x] Another team member has reviewed
- [x] Tests are green

Accounts 

- [ ] Core-Shared-Services
- [ ] Core-Network-Services 
- [ ] Core-Logging 
- [ ] Core-Security
- [ ] Core-VPC-* 
- [ ] Modernisation-Platform
- [ ] PagerDuty
- [ ] Github

dms1981:
We do have defensive tests in place at present using terratest - https://github.com/ministryofjustice/modernisation-platform/blob/main/.github/workflows/core-logging-deployment.yml#L71

Is this story meant to supplant these old tests and perform them through OPA?

richgreen-moj:
So far I've raised this PR in draft https://github.com/ministryofjustice/modernisation-platform/pull/6281

The plan is to do the following

1. Turn on status checks for the opa policy testing workflow so that PRs cannot be merged without the tests passing.
2. Add a workflow/script that adds comments to PRs warning of the risks associated with making changes to various areas oof the code. 

I've also scheduled a meeting with the wider team to discuss any further test use cases we can implement.

richgreen-moj:
https://github.com/ministryofjustice/modernisation-platform/pull/6281 is now ready for review. It's an MVP for warning reviewers around high risk changes and can be improved over time.

richgreen-moj:
I've raised https://github.com/ministryofjustice/modernisation-platform/issues/6388 which could take the work in this ticket a step further

PR #6281, opened 2024-02-21
Add OPA status check and PR Warnings

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/5947
As an MP engineer
I want to be alerted to or prevented from making changes that may break the platform
So that I don't break the platform.

1. I'm turning on status checks for the `opa-policies.yml` workflow  so that PRs cannot be merged without the tests passing.
(_This will ensure all the current policies written must be adhered to including a check which would highlight if you rename an application.json file, and more..._)

1. I've added a new step to the `opa-policies.yml` workflow which runs a script `pr-warning-comments.sh` which checks for all the files/directories being changed in the PR and then adds an associated warning comment to the PR to give the reviewer some context around the potential impact that the changes could make. 
These are listed in `policies/pr-messages.json` where the `file` denotes a regex for the file/directory and `message` contains the relevant warning message.

I propose that we try this out as an MVP and if useful we can continue to add further warning messages or tests as we go. It could definitely be improved upon but basic functionality is there.

## How has this been tested?

I've tested this by making minor adjustments to the files (adding blank lines) so that I could trigger the logic in the workflow and see the comments on this PR. [Here is an example](https://github.com/ministryofjustice/modernisation-platform/pull/6281#issuecomment-1973724203) of it triggering for all of the current use cases.

- [X] All checks have passed

Ive raised another issue https://github.com/ministryofjustice/modernisation-platform/issues/6388 to look in to some more targeted warnings which inspect the terraform plan (a bit like the TF plan evaluator in the MP Environments repo) where potentially changes to certain files give a more targeted warning or require an extra reviewer etc.


davidkelliott (approved):
This is really great work thanks Rich 👏

PR #6300, opened 2024-02-22
Bootstrap change using the baseline module changes for aws backup alerts.

ep-93:

Will apply the monitoring to all accounts which are not `member-unrestricted`

Ties into the changes made in the baselines module here - https://github.com/ministryofjustice/modernisation-platform-terraform-baselines/pull/393/files

## How has this been tested?

Applied in cooker, and also tried to in bichard, with no changes to apply, filter works correctly.

- [ x] All checks have passed
- [ x] Plan and discussed how it should be deployed to PROD (If needed)


mikereiddigital (dismissed):
lg2m.

terraform/environments/sprinkler/notifications.tf
```diff
@@ -0,0 +1,57 @@
+# Data source to get the ARN of an existing SNS topic
+data "aws_sns_topic" "existing_topic" {
+  name = "backup_failure_topic"
+}
+
+# Create an email subscription to the existing SNS topic
```
ewastempel, line 6: You can set it up at first to test it, but I think it could be spammy to have it there permanently (but see what others think, I did not set up an email notification for the instance scheduler and all other cloudwatch alerts are also not set up with an email notification)

terraform/environments/sprinkler/notifications.tf
```diff
@@ -0,0 +1,57 @@
+# Data source to get the ARN of an existing SNS topic
+data "aws_sns_topic" "existing_topic" {
```
ewastempel, line 2: I wonder if it makes sense to define the sns topic here, rather than in the module (since it's not doing anything there anyway)

terraform/environments/sprinkler/notifications.tf
ep-93: Good point, I have it in there just to test the creation but going forward I could easily remove it!

ASTRobinson (dismissed):

dms1981 (approved):

PR #6230, opened 2024-02-15
DSOS-2615: enable ad-azure-dc-a EC2 in core-shared-services-production

drobinson-moj:
## A reference to the issue / Description of it

Follow on to https://github.com/ministryofjustice/modernisation-platform/pull/6224

## How does this PR fix the problem?

Enables first Domain Controller

## How has this been tested?

Please describe the tests that you ran and provide instructions to reproduce.

Tested as far as possible in environments repo in a member account

## Deployment Plan / Instructions

Will this deployment impact the platform and / or services on it?

No

## Checklist (check `x` in `[ ]` of list items)

- [x] I have performed a self-review of my own code
- [x] All checks have passed
- [ ] I have made corresponding changes to the documentation
- [ ] Plan and discussed how it should be deployed to PROD (If needed)

## Additional comments (if any)

Mod Platform engineers aware of solution: Ewa Stempel, David Sibley, Dave Elliot

PR #6359, opened 2024-02-29
DSOS-2635: add ability to login via FleetManager

drobinson-moj:

and https://github.com/ministryofjustice/modernisation-platform/pull/6230
In summary, tactical solution for migrating Domain Controllers from Azure FixNGo estate.

Problem: the EC2 instances are not accessible via developers

Adds a key-pair for Administrator access to servers.  The key-pair private key is held in hmpps-domain-services account along with other domain specific secrets.

Adds a new role which can be assumed by developers (in same ways as member-shared-services) with minimum permissions required to access EC2 via Fleet Manager GUI.  Although any mod platform user can assume this role, the user can't login unless they also have access to hmpps-domain-services where the creds are stored.

The PR also removes the first DC and creates a new one with the given key-pair

## How has this been tested?

As far as possible in hmpps-domain-services account, e.g. we have correct permission set and role can be assumed.

No - service is not live yet.

- [x] All checks have passed

Dave Elliot / Connor / David Sibley might want to cast their eye over the new role.

### DIFF
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -13,6 +13,12 @@ locals {
 
   ad_fixngo = {
 
+    aws_key_pairs = {
+      # See https://dsdmoj.atlassian.net/wiki/x/3oCKGAE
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

Issue #6145, opened 2024-02-06
Create a job to run and check inactive collaborator IAM accounts.


ep-93:
### User Story

As a mod platform engineer
I want to automate a job to check inactive collaborator accounts.
So that we we dont have old accounts in the mod platform account.

### Value / Purpose

Automation

### Additional Information

NA

### Definition of Done

- [ ] Decide on script type
- [ ] User docs have been updated
- [ ] Another team member has reviewed
- [ ] Tests are green

sukeshreddyg:
created a job to notify the team via email about inactive Collaborator IAM accounts with inactivity of 90 days or more has been successfully completed. Additionally, I have raised a follow-up ticket #6390 to implement automated processes for disabling and deleting inactive IAM accounts. This ticket will address the next steps in our IAM account management strategy.

PR #6371, opened 2024-03-01
Add workflow for notifying inactive collaborators

sukeshreddyg:

#6145 

This PR needs for automating the process of notifying inactive collaborator users


richgreen-moj (approved):
LGTM

PR #6380, opened 2024-03-04
Instructions for revoking network access

SteveLinden:
This is a  list of the instructions showing how to restrict network access if we are being attacked. NOTE it is listed with the console changes needed NOT the terraform code changes. The main reason for this are so the code is not amended and can  be re-used to get everything back to the way it was.

Once approved I will amend the readme in the modernisation-platform code.

- [x] All checks have passed

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -114,7 +114,7 @@ This documentation is for anyone interested in the Modernisation Platform and it
 - [Querying VPC flow logs](runbooks/querying-vpc-flow-logs.html)
 - [Removing a team member from the Modernisation Platform](runbooks/removing-a-team-member.html)
 - [Reviewing Dependabot PRs](runbooks/reviewing-dependabot-prs.html)
-- [Revoke Network Access](revoke-network-access.html)
+- [Revoke Network Access](runbooks/revoke-network-access.html)
 - [Revoking User Access](runbooks/revoking-user-access.html)
 - [Security Testing and ITHC](user-guide/security-testing-and-ithc.html)
 - [Terraform](runbooks/terraform.html)
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -114,6 +114,7 @@ This documentation is for anyone interested in the Modernisation Platform and it
 - [Querying VPC flow logs](runbooks/querying-vpc-flow-logs.html)
 - [Removing a team member from the Modernisation Platform](runbooks/removing-a-team-member.html)
 - [Reviewing Dependabot PRs](runbooks/reviewing-dependabot-prs.html)
+- [Revoke Network Access](revoke-network-access.html)
 - [Revoking User Access](runbooks/revoking-user-access.html)
 - [Security Testing and ITHC](user-guide/security-testing-and-ithc.html)
 - [Terraform](runbooks/terraform.html)
--- a/source/runbooks/revoke-network-access.html.md.erb
+++ b/source/runbooks/revoke-network-access.html.md.erb
@@ -21,7 +21,7 @@ All changes are to be made manually on the relevant environment (for example cor
 
 This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
 
-If you know which VPC application area is being attacked you can remove it from the VPC.
+If you know which VPC application area is being attacked you can remove the connection by following these instructions.
 
 1. Go to VPC->Internet gateways
 2. Select the gateway (example is hmpps-preproduction-internet-gateway)
@@ -37,12 +37,14 @@ This change will affect all applications in the VPC that you have selected. No a
 
 This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
 
-The process needs to be done for all 4 regions listed. An example is
+The process needs to be done for all 3 NACLs listed. An example is
 
 - hmpps-production-data-nacl
 - hmpps-production-private-nacl
 - hmpps-production-protected-nacl
 - hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
 
 Follow the process below for each of these
 
@@ -51,18 +53,20 @@ Follow the process below for each of these
 3. Due to the way checks work in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) we need to amend each one in place to Deny where it is currently Allow. **Note** any that are currently set to Deny in case this is manually changed back.
 4. Save changes
 
-Setting up a deny rule on the inline-inspection firewalls, regardless of the processing order, is not possible due to the issue mentioned above. This is why the deny on each area is required.
+Setting up a deny rule on the stateful inline-inspection rules, regardless of the processing order, is not possible due to the issue mentioned above. This is why the deny on each area is required.
 
 ## Stopping access from the MOJ to an application on MP
 
 This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
 
-The process needs to be done for all 4 regions listed. An example is
+The process needs to be done for all 4 NACLs listed. An example is
 
 - hmpps-production-data-nacl
 - hmpps-production-private-nacl
 - hmpps-production-protected-nacl
 - hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those llisted for the other NACLs.
 
 You need to access this through the same route as above:
 
@@ -72,18 +76,20 @@ You need to access this through the same route as above:
 4. Ensure that the access to the system is completely denied by amending the TCP to ports 1024 - 65535 on 0.0.0.0/0 to Deny too.
 5. Save the changes
 
-As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work.
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work. **Note** any that are currently set to Deny in case this is manually changed back.
 
 ## Stopping access to the MOJ for an application on MP
 
 This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
 
-The process needs to be done for all 4 regions listed. An example is
+The process needs to be done for all 4 NACLs listed. An example is
 
 - hmpps-production-data-nacl
 - hmpps-production-private-nacl
 - hmpps-production-protected-nacl
 - hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
 
 You need to access this through the same route as above:
 
@@ -93,4 +99,4 @@ You need to access this through the same route as above:
 4. Ensure that the access to the system is completely denied by amending the TCP to ports 1024 - 65535 on 0.0.0.0/0 to Deny too.
 5. Save the changes
 
-As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work.
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work. **Note** any that are currently set to Deny in case this is manually changed back.
--- a/source/runbooks/revoke-network-access.html.md.erb
+++ b/source/runbooks/revoke-network-access.html.md.erb
@@ -18,16 +18,16 @@ The steps listed deny access on a platform level. If additional changes are requ
 
 A network attack is an exceptional situation so quick amendments are needed. It is anticipated that this can be done quickly through the console.
 
-This document lists the steps needed to stop access to and from the internet and preventing the system from being accessed from or accessing other applications on the MoJ.
+This document lists the steps required to revoke internet access, and isolate the compromised infrastructure from other applications.
 
 All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything. All of the steps listed will be performed on the core-vpc-<environment> location unless otherwise specified.
 
-[CAUTION]
+## CAUTION
 > All networking changes via terraform should be stopped while this work is ongoing. No MP terraform runs should be permitted while there is work going on to prevent unauthorised access to the environments.
 
 > The changes listed in this document will impact all applications connected to a particular NACL. If you were to amend, for example, hmpps-production-data-nacl, everything that connects via that NACL will be impacted. Nothing that connects from HMPPS through this NACL will have access.
 
-> Additionally we are only making the changes manually to prevent the code being updated to the detriment of subsequent repairs. Assuming the issue is resolved we would need to rerun the original terraform code to reset everything to it's previous state. Any changes made to the code could cause us issues in being able to do this. It is likely that when applying the code there will be an warning shown stating that some changes have been made outside terraform. This is not something to worry about as all access will be reset to the previous state.
+> Additionally, we are only making the changes manually to prevent the code being updated to the detriment of subsequent repairs. Assuming the issue is resolved we would need to rerun the original terraform code to reset everything to it's previous state. Any changes made to the code could cause us issues in being able to do this. It is likely that when applying the code there will be an warning shown stating that some changes have been made outside terraform. This is not something to worry about as all access will be reset to the previous state.
 
 ## Stopping access to/from a single application
 
@@ -39,7 +39,7 @@ The most appropriate way is to remove rules from the relevant security groups. Y
 4. Delete each rule listed. For some security groups there may be a long list of items to delete. 
 5. You will need to Save rules to apply your change.
 
-The default behaviour for a security group with no rules is to Deny all traffic. 
+The default behaviour for a security group with no rules is to `deny` all traffic. 
 
 ## Stopping access from the internet to an application on MP
 
@@ -74,8 +74,8 @@ Follow the process below for each of these.
 
 1. Go to VPC->Security->Network ACLs.
 2. Click on the checkbox for the ACL you want to change then select **outbound rules**.
-3. Add a new Deny for **0.0.0.0/0**, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above based on our current code.    
-3a. For `hmpps-production-protected-nacl` add rule number 1 with a Deny on **0.0.0.0/0**.
+3. Add a new `deny` for **0.0.0.0/0**, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above based on our current code.    
+3a. For `hmpps-production-protected-nacl` add rule number 1 with a `deny` on **0.0.0.0/0**.
 4. Save changes.
 
 ## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
@@ -94,7 +94,7 @@ You need to access this through the same route as below.
 
 1. Go to VPC->Security->Network ACLs.
 2. Click on the checkbox for the one you want to change then select **inbound rules**.
-3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
+3. Amend all the rules that are >= 4000 and < 5000 so they are in a `deny` state.
 4. Save the changes.
 
 **NOTE** the rules between 4000 and 5000 is based on the code being correctly set up to put the rules in this range. Amend the instructions if this changes.
@@ -115,7 +115,7 @@ You need to access this through the same route as below:
 
 1. Go to VPC->Security->Network ACLs
 2. Click on the checkbox for the one you want to change then select **outbound rules**
-3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
+3. Amend all the rules that are >= 4000 and < 5000 so they are in a `deny` state.
 4. Save the changes
 
 **NOTE** the rules between 4000 and 5000 is based on the code being correctly set up to put the rules in this range. Amend the instructions if this changes.
@@ -149,7 +149,7 @@ The process needs to be done for the 3 NACLs listed below. Examples for producti
 
 1. Connect to `core-shared-services`.
 2. Click on the checkbox for the one you want to change then select **inbound rules**.
-3. Amend all the rules that are >= 500 and < 600 so they are in a Deny state.
+3. Amend all the rules that are >= 500 and < 600 so they are in a `deny` state.
 4. Save the changes.
 
 **NOTE** `core-network-services` VPCs host network firewall endpoints. Amending NACLs here will affect all traffic moving through the VPCs, either to the internet or to the rest of the MOJ.
@@ -168,7 +168,7 @@ The process needs to be done for the 3 NACLs listed below. Examples for producti
 
 1. Connect to `core-shared-services`.
 2. Click on the checkbox for the one you want to change then select **outbound rules**.
-3. Add a new Deny for **0.0.0.0/0**, rule number 211, which will prevent access to the items listed after this point. As rule 210 allows access inside the VPC, all rules above allow traffic to leave the VPC.
+3. Add a new `deny` for **0.0.0.0/0**, rule number 211, which will prevent access to the items listed after this point. As rule 210 allows access inside the VPC, all rules above allow traffic to leave the VPC.
 4. Save the changes.
 
 **NOTE** `core-network-services` VPCs host network firewall endpoints. Amending NACLs here will affect all traffic moving through the VPCs, either to the internet or to the rest of the MOJ.
--- a/source/runbooks/revoke-network-access.html.md.erb
+++ b/source/runbooks/revoke-network-access.html.md.erb
@@ -20,7 +20,7 @@ A network attack is an exceptional situation so quick amendments are needed. It 
 
 This document lists the steps needed to stop access to and from the internet and preventing the system from being accessed from or accessing other applications on the MoJ.
 
-All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything. All of the steps listed will be performed on the core-vpc-(environment) location unless otherwise specified.
+All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything. All of the steps listed will be performed on the core-vpc-<environment> location unless otherwise specified.
 
 [CAUTION]
 > All networking changes via terraform should be stopped while this work is ongoing. No MP terraform runs should be permitted while there is work going on to prevent unauthorised access to the environments.
@@ -43,7 +43,7 @@ The default behaviour for a security group with no rules is to Deny all traffic.
 
 ## Stopping access from the internet to an application on MP
 
-This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+This work is carried out from the core-vpc-<environment> environment e.g. core-vpc-production
 
 If you know which business unit the application under attack is in  you can disconnect the VPC from the internet by following these instructions.
 
@@ -61,7 +61,7 @@ This change will affect all applications in the VPC that you have selected. No a
 
 Due to the way Network Firewall applies rules in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL. 
 
-This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production.
+This work is carried out from the core-vpc-<environment> environment e.g. core-vpc-production.
 
 The process needs to be done for all 3 NACLs listed. An example is
 
@@ -69,18 +69,18 @@ The process needs to be done for all 3 NACLs listed. An example is
 - hmpps-production-private-nacl
 - hmpps-production-public-nacl
 
-Follow the process below for each of these
+Follow the process below for each of these.
 
 1. Go to VPC->Security->Network ACLs.
 2. Click on the checkbox for the ACL you want to change then select **outbound rules**.
-3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above.
+3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above based on our current code.
 4. Save changes.
 
 ## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
 
 As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL.  
 
-This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+This work is carried out from the core-vpc-<environment> environment e.g. core-vpc-production
 
 The process needs to be done for all 3 NACLs listed. An example is
 
@@ -88,18 +88,18 @@ The process needs to be done for all 3 NACLs listed. An example is
 - hmpps-production-private-nacl
 - hmpps-production-public-nacl
 
-In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
-
-You need to access this through the same route as above:
+You need to access this through the same route as below.
 
 1. Go to VPC->Security->Network ACLs.
 2. Click on the checkbox for the one you want to change then select **inbound rules**.
 3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
 4. Save the changes.
 
+**NOTE** the rules between 4000 and 5000 is based on the code being correctly set up to put the rules in this range. Amend the instructions if this changes.
+
 ## Stopping access to the MOJ for an application via the Modernisation Platform Transit Gateway
 
-As previously mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work. **Note** any that are currently set to Deny in case this is manually changed back.
+As previously mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL. 
 
 This work is carried out from the core-vpc-<environment> account, e.g. core-vpc-production
 
@@ -109,12 +109,14 @@ The process needs to be done for all 3 NACLs listed. An example is
 - hmpps-production-private-nacl
 - hmpps-production-public-nacl
 
-You need to access this through the same route as above:
+You need to access this through the same route as below:
 
 1. Go to VPC->Security->Network ACLs
 2. Click on the checkbox for the one you want to change then select **outbound rules**
 3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
 4. Save the changes
+
+**NOTE** the rules between 4000 and 5000 is based on the code being correctly set up to put the rules in this range. Amend the instructions if this changes.
 
 ## Blocking access for something moving internally on MP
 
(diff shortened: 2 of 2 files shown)


source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+1. Go to VPC->Internet gateways
+2. Select the gateway (example is hmpps-preproduction-internet-gateway)
+3. Note the Internet gateway ID and the VPC ID
+4. Click on Actions, Detach from VPC
+5. If you are sure this is correct click on the Detach internet gateway button
+
+The selected internet gateway will be detached from the VPC.
+
+This change will affect all applications in the VPC that you have selected. No applications in the VPC will be accessible from the internet.
```
ewastempel, line 34: At this point it is worth mentioning, that the issue could be isolated to one application only and if other applications cannot afford an outage, then another approach should be taken (perhaps looking at NACLs on the MP side and elaborating on it and/or LB/WAF/SG on the application side, requiring to work closely with the application team)

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+# <%= current_page.data.title %>
+
+# Summary
+
+The steps listed deny access on a platform level. If additional changes are required at an application level then these must be undertaken by the application owning team.
+
+A network attack is an exceptional situation so quick amendments are needed. It is anticiapted that this can be done quickley through the console.
+
+This document lists the steps needed to stop access to and from the internet and prevening the system from brineg accessed from or accessing other applications on the MoJ.
```
ewastempel, line 16:
```suggestion
This document lists the steps needed to stop access to and from the internet and prevening the system from being accessed from or accessing other applications on the MoJ.
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code and be run and this will reset everything.
```
ewastempel, line 18:
```suggestion
All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything.
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
```
ewastempel, line 19: At this point it would be good to mention (perhaps in a warning NOTE format) all the risks that the manual changes carry, see the [slack thread ](https://mojdt.slack.com/archives/C013RM6MFFW/p1709552392385159)for more information

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+## Stopping access from the internet to an application on MP
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+
+If you know which VPC application area is being attacked you can remove it from the VPC.
```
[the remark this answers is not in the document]
SteveLinden: Probably just an oversight. I will adjust it

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Yes. I missed this

source/runbooks/revoke-network-access.html.md.erb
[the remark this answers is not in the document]
SteveLinden: Is this acceptable?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+The selected internet gateway will be detached from the VPC.
+
+This change will affect all applications in the VPC that you have selected. No applications in the VPC will be accessible from the internet.
+
+## Stopping access to the internet for an application on MP
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+
+The process needs to be done for all 4 regions listed. An example is
```
ewastempel, line 40: The following are not regions but NACLs for each subnet type set and there are only 3 regions per NACLs).

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+- hmpps-production-data-nacl
+- hmpps-production-private-nacl
+- hmpps-production-protected-nacl
+- hmpps-production-public-nacl
+
+Follow the process below for each of these
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the one you want to change then select **outbound rules**.
+3. Due to the way checks work in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) we need to amend each one in place to Deny where it is currently Allow. **Note** any that are currently set to Deny in case this is manually changed back.
```
ewastempel, line 51: This is not true and the and the linked documentation is irrelevant here, as it is related to the firewall changes and not NACLs.
See [this documentation ](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html) on the evaluation order of the rules and update the documentation to reflect it.
Think about a smarter way to implement it given the rule evaluation behaviour, updating every single rule in each NACL sounds like a lot of manual work.

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+4. Save changes
+
+Setting up a deny rule on the inline-inspection firewalls, regardless of the processing order, is not possible due to the issue mentioned above. This is why the deny on each area is required.
```
ewastempel, line 54: You're mixing NACLs with firewall again.

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+## Stopping access from the MOJ to an application on MP
```
ewastempel, line 56: `from the MOJ` sounds a bit broad

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
```
ewastempel, line 58: Is that right? Would it not be better to block the firewall?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the one you want to change then select **inbound rules**.
+3. To prevent access change the one for the current system to Deny. For example in production you would amend 10.27.8.0/21 to deny for hmpps-production.
+4. Ensure that the access to the system is completely denied by amending the TCP to ports 1024 - 65535 on 0.0.0.0/0 to Deny too.
+5. Save the changes
+
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work.
+
+## Stopping access to the MOJ for an application on MP
```
ewastempel, line 77: The same comment as above, controlling the access through NACLs requires more work in this scenario, but the same can be achieved through the firewall rules or TGW changes (or possibly route table changes)

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Mistakenly added to the previous thread
https://github.com/ministryofjustice/modernisation-platform/pull/6380#discussion_r1511242225

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,96 @@
...
+- hmpps-production-public-nacl
+
+You need to access this through the same route as above:
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the one you want to change then select **outbound rules**
+3. To prevent access change the one for the current system to Deny. For example in production you would amend 10.27.8.0/21 to deny for hmpps-production.
+4. Ensure that the access to the system is completely denied by amending the TCP to ports 1024 - 65535 on 0.0.0.0/0 to Deny too.
+5. Save the changes
+
```
ewastempel, line 95: The documentation so far addresses (although in some cases it is not correct)

-  Blocking access for something entering via the internet
-  Blocking access for something leaving via the internet
-  Blocking access for something leaving via an internal route (eg, TGW)
-  Blocking access for something entering via an internal route (again, eg TGW)

 
 but it is still missing:

-   Blocking access for something moving internally inside the Modernisation Platform

  and ignores looking at the problem from the other angle:
   - isolating a single service/app or an environment (or even worth considering isolating a business unit as oppose to as single service/app)
   -  blocking ingress/egress in multiple or all services/environments

source/runbooks/revoke-network-access.html.md.erb
ewastempel: this is still not addressed

source/runbooks/revoke-network-access.html.md.erb
ewastempel: this is still not addressed

source/runbooks/revoke-network-access.html.md.erb
ewastempel: Your warning is regarding the changes impact (which is already covered in this documentation), but I am looking for a warning about the risks that manual changes (vs code changes) carry, since we are doing it through click-ops and not through code.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: No more than 10 rules to amend in each. Sounds like a lot but changing from allow to deny isn't too much effort

source/runbooks/revoke-network-access.html.md.erb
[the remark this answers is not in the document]
ewastempel: This caution is not needed as it's already covered (see the other thread).
This sentence just needs rewriting so that it makes sense.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Corrected

source/runbooks/revoke-network-access.html.md.erb
ewastempel: See my previous comment, as what you are saying about the NACLs rule evaluation is incorrect.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Added a section about reducing access to a single application

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: This has been amended

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: This has been corrected

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Amended

source/runbooks/revoke-network-access.html.md.erb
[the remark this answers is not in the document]
SteveLinden: This has been corrected

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Can you clarify this please?

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: We have reviewed this again and have amended the documentation to add a new line to the rules to add a global deny. I have listed the current section as an option.

source/runbooks/revoke-network-access.html.md.erb
dms1981: Blocking at the firewall would be a lot more work than it would look like at the outset: the short version is that the `permits` all get applied before the `denies` are looked at, so you'd need to prune what you permit first before then setting the denies, or consider changing to use `alert` instead of `permit` for traffic passing through a network firewall (alert allows traffic, but logs it. permit allows traffic and doesn't log it).

SteveLinden: We discussed this and decided it was better to do it through this route rather than letting something bypassing here and hitting the firewall. It would also make this "quick fix" temporary measure a lot more complex.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: As above

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I may have covered this with my other changes. They will be posted later.

dms1981: Regarding isolating a single business unit, that ought to be covered by the guidance on detaching an internet gateway, and putting a deny rule at the top of the ingress/egress NACL rules. In theory you could also do this by removing the TGW route table association in `core-network-services` but that would only cover part of the routing (enough to break TCP sessions, but not to stop UDP traffic coming in).

source/runbooks/revoke-network-access.html.md.erb
ewastempel: What I was after and what was already discussed in the [slack thread](https://mojdt.slack.com/archives/C013RM6MFFW/p1709559726905539?thread_ts=1709552392.385159&cid=C013RM6MFFW) is something along the lines of: 
These manual changes mean that:
- there is no history of changes in code and therefore if there are additional changes done that are not part of current IaC (e.g. introducing a new route table), these will need noting
- there is a danger of unintentionally undoing these changes with a pipeline run, hence for the time when the manual changes are in place, no pipeline relevant to the networking bits that were updated can be run

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+This change will affect all applications in the VPC that you have selected. No applications in the VPC will be accessible from the internet. 
+
+## Stopping access to the internet for applications on MP
+
+Due to the way Network Firewall applies rules in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL. 
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production.
+
+The process needs to be done for all 3 NACLs listed. An example is
```
ewastempel, line 64: There are 5 NACLs sets per each business unit's environment and there are 3 subnets associated with each NACL set. This needs rewriting to be accurate.

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+- hmpps-production-data-nacl
+- hmpps-production-private-nacl
+- hmpps-production-public-nacl
+
+Follow the process below for each of these
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the ACL you want to change then select **outbound rules**.
+3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to add the items listed after this point. All internet rules are 5000 and above.
```
ewastempel, line 74:
```suggestion
3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above.
```

source/runbooks/revoke-network-access.html.md.erb
ewastempel: you can probably skip the transit gateway subnets, but need all: data, private, protected and public

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+4. Save changes
+
+## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work. **Note** any that are currently set to Deny in case this is manually changed back.
```
ewastempel, line 81: This sentence makes no sense.

source/runbooks/revoke-network-access.html.md.erb
ewastempel: Fair enough @dms1981 this would be challenging (hence originally suggested changes in code), but Steve's suggestion was to amend every Allow rule to Deny, which would not only mean that we would block all the traffic (which in case of blocking MOJ internal traffic only, was not desired), but it also meant a large number of manual changes (e.g. 10 NACL rules in each subnet type, such as data, private, public, protected then multiplied by 3 regions, never mind making changes for multiple business units or multiple environments).

it's the other way around Steve, incoming traffic from MOJ, coming through the MP TGW will be first filtered through the firewall in core-network-services VPC then routed to the core-vpc-<env>

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+- hmpps-production-private-nacl
+- hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
+
+You need to access this through the same route as above:
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the one you want to change then select **inbound rules**.
+3. Amend all the rules on 4000, 4100 and 4200 so they are in a Deny state.
```
ewastempel, line 97: This should say rules >= 4000 and < 5000 rather than list specific rules, as they can vary between each NACL and they can be expanded in the future. You could also mention at the beginning that we dedicate these rules to private CIDR ranges (hence only looking to amend the rules within this range).

source/runbooks/revoke-network-access.html.md.erb
dms1981: I don't `protected` subnets are going to be relevant in this case as they never hold customer infrastructure and should only host VPC endpoints.

source/runbooks/revoke-network-access.html.md.erb
dms1981: So for the avoidance of doubt @SteveLinden , this could be met by adding a [blockquote](https://www.markdownguide.org/basic-syntax/#blockquotes-1) in near the top of the document that says those things.

source/runbooks/revoke-network-access.html.md.erb
dms1981: All good points @ewastempel - I'll work with @SteveLinden to firm up his understanding here.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Additional comment added

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Understood. I have amended the code to add a new rule, prior to the ones we want to deny, which will alleviate the need to amend all the rows.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I have added additional code to list what to do to remove the transit gateway. This will prevent an environment routing around through others.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: This has been reduced to the 3 (data, private, public) as per David's suggestion.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Moved and corrected.

source/runbooks/revoke-network-access.html.md.erb
ewastempel: true, although is there anything stopping users from hosting infrastructure there?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+# Summary
+
+The steps listed deny access on a platform level. If additional changes are required at an application level then these must be undertaken by the application owning team. 
+
+A network attack is an exceptional situation so quick amendments are needed. It is anticipated that this can be done quickly through the console.
+
+This document lists the steps needed to stop access to and from the internet and preventing the system from being accessed from or accessing other applications on the MoJ.
+
+All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything. All of the steps listed will be performed on the core-vpc-(environment) location unless otherwise specified.
```
ewastempel, line 23:
```suggestion
All changes are to be made manually on the relevant environment (for example core-vpc-production) and are not made through the code. If a way of reverting back to baseline is required the terraform code can be run and this will reset everything. All of the steps listed will be performed on the core-vpc-<environment> location unless otherwise specified.
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above.
+4. Save changes.
+
+## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
+
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL.  
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+
+Blocking the firewall may no be an effective solution. Blocking is done here to prevent unauthorised access.
```
dms1981, line 85: Is this necessary since you've already mentioned it at the start of this section?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+[CAUTION]
+> All changes via terraform should be stopped while this work is ongoing. No MP terraform runs should be permitted while there is work going on to prevent unauthorised access to the environments.
+
+> The changes listed in this document will impact all applications connected to a particular NACL. If you were to amend, for example, hmpps-production-data-nacl, everything that connects via that NACL will be impacted. Nothing that connects from HMPPS through this NACL will have access.
+
+> Additionally we are only making the changes manually to prevent the code being updated to the detriment of subsequent repairs. Assuming the issue is resolved we would need to rerun the original terraform code to reset everything to it's previous state. Any changes made to the code could cause us issues in being able to do this. It is likely that when applying the code there will be an warning shown stating that some changes have been made outside terraform. This is not something to worry about as all access will be reset to the previous state.
+
+## Stopping access to a single application
```
ewastempel, line 32:
```suggestion
## Stopping access from/to a single application
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+The selected internet gateway will be detached from the VPC.
+
+This change will affect all applications in the VPC that you have selected. No applications in the VPC will be accessible from the internet. 
+
+## Stopping access to the internet for an application on MP
+
+Due to the way Network Firewall applies rules in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL. 
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production.
```
ewastempel, line 64:
```suggestion
This work is carried out from the core-vpc-<environment>, e.g. core-vpc-production.
```

source/runbooks/revoke-network-access.html.md.erb
ewastempel: also am I OK to assume that nothing can be deployed in the transit subnet if the TGW attachment is in place? or should we worry about that subnet too?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+4. Save changes.
+
+## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
+
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL.  
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
+
+Blocking the firewall may no be an effective solution. Blocking is done here to prevent unauthorised access. 
+
```
ewastempel, line 86:
```suggestion
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+- hmpps-production-private-nacl
+- hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
+
+You need to access this through the same route as above:
+
+1. Go to VPC->Security->Network ACLs.
+2. Click on the checkbox for the one you want to change then select **inbound rules**.
+3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
```
ewastempel, line 99: Mention it somewhere why we are amending only >= 4000 and <5000 rules, so that it is clear that we rely on our good practices (such as using >= 4000 and < 5000 rules for private address ranges only) being followed

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+4. Save the changes.
+
+## Stopping access to the MOJ for an application via the Modernisation Platform Transit Gateway
+
+As previously mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ) mean adding a step for Deny may not work. **Note** any that are currently set to Deny in case this is manually changed back.
```
ewastempel, line 104: This sentence does not read right again.

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
```
ewastempel, line 106:
```suggestion
This work is carried out from the core-vpc-<environment> e.g. core-vpc-production
```

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+The process needs to be done for all 3 NACLs listed. An example is
+
+- hmpps-production-data-nacl
+- hmpps-production-private-nacl
+- hmpps-production-public-nacl
+
+You need to access this through the same route as above:
```
ewastempel, line 114: or below?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+1. Go to VPC->Security->Network ACLs
+2. Click on the checkbox for the one you want to change then select **outbound rules**
+3. Amend all the rules that are >= 4000 and < 5000 so they are in a Deny state.
+4. Save the changes
+
+## Blocking for something moving internally on MP
```
ewastempel, line 121:
```suggestion
## Blocking access for something moving internally on MP
```

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I tried to use <> initially and they didn't show up in the review of the document. I can add them again but we may just see core-vpc  without the environments word

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: As mentioned above with the <> is does not resolve correctly when opened.

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+2. Select the appropriate security group.
+3. Select both Edit inbound rules and Edit outbound rules.
+4. Delete each rule listed. For some security groups there may be a long list of items to delete. 
+5. You will need to Save rules to apply your change.
+
+The default behaviour for a security group with no rules is to Deny all traffic. 
+
+## Stopping access from the internet to an application on MP
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
```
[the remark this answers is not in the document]
SteveLinden: As above

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+1. Go to VPC->Security->Network ACLs.
+2. Click on the checkbox for the ACL you want to change then select **outbound rules**.
+3. Add a new Deny rule for 0.0.0.0/0, rule number 4999, which will prevent access to the items listed after this point. All internet rules are 5000 and above.
+4. Save changes.
+
+## Stopping access from the MOJ to an application via the Modernisation Platform Transit Gateway
+
+As mentioned above the way checks are undertaken in AWS (see [here](https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-rule-evaluation-order.html) ), this section only covers changes to the VPC NACL.  
+
+This work is carried out from the core-vpc-(environment) environment e.g. core-vpc-production
```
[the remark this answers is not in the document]
SteveLinden: As above

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Tidied again. For some reason a section I thought I had removed re-appeared

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Changed

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: The TGW subnets are not shared across the whole system. We should not have issues. A bad actor would also need access to the code and be able to amend and deploy it.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I can't see this line in my code now

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I had amended it but the old version is still showing. It has been amended again

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,134 @@
...
+
+The process needs to be done for all 3 NACLs listed. An example is
+
+- hmpps-production-data-nacl
+- hmpps-production-private-nacl
+- hmpps-production-public-nacl
+
+In this case the protected one needs to be amended as, although TCP references are in place, the IP addresses are the same as those listed for the other NACLs.
+
+You need to access this through the same route as above:
```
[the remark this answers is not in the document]
SteveLinden: Removed the initial comment about accessing the protected route.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Amended. I think I was pointing out it was as done in the previous version

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: Applied but still showing here?

source/runbooks/revoke-network-access.html.md.erb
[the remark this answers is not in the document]
SteveLinden: See above

source/runbooks/revoke-network-access.html.md.erb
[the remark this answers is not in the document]
SteveLinden: See above

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: It looks like it was related to something I had installed as an extension. This has been removed and the code appears OK with the <> now.
I have manually amended it in place rather than through the commit here.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: See above

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: I've added a footnote (according to the writing on github code) so this may be in place.

source/runbooks/revoke-network-access.html.md.erb
ewastempel: After having a discussion in slack, this is now to extend the list of NACLs, for `core-*` accounts

source/runbooks/revoke-network-access.html.md.erb
ewastempel: I can see it as plain text with bold NOTE underneath the instructions, but that will do.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: To include the protected network this has been included in the list so there are now 4 shown. 
The transit gateway is shown in 4 other locations so these will be listed too.

source/runbooks/revoke-network-access.html.md.erb
SteveLinden: The footnote didn't seem to appear in my testing so I added the note to compensate.

source/index.html.md.erb
```diff
@@ -114,6 +114,7 @@ This documentation is for anyone interested in the Modernisation Platform and it
 - [Querying VPC flow logs](runbooks/querying-vpc-flow-logs.html)
 - [Removing a team member from the Modernisation Platform](runbooks/removing-a-team-member.html)
 - [Reviewing Dependabot PRs](runbooks/reviewing-dependabot-prs.html)
+- [Revoke Network Access](revoke-network-access.html)
```
ep-93, line 117: [Revoke Network Access](runbooks/revoke-network-access.html)

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+If the work is required outside normal hours then the teams are unlikely to respond and you can continue with the items listed below. Ensure that the teams are aware before starting this if it happens inside working hours.
+
+# Summary
+
+The steps listed deny access on a platform level. If additional changes are required at an application level then these must be undertaken by the application owning team. 
+
+A network attack is an exceptional situation so quick amendments are needed. It is anticipated that this can be done quickly through the console.
+
+This document lists the steps needed to stop access to and from the internet and preventing the system from being accessed from or accessing other applications on the MoJ.
```
ep-93, line 21: This document lists the steps required to revoke internet access, and isolate the compromised infrastructure from other applications.

Maybe?

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+[CAUTION]
+> All networking changes via terraform should be stopped while this work is ongoing. No MP terraform runs should be permitted while there is work going on to prevent unauthorised access to the environments.
+
+> The changes listed in this document will impact all applications connected to a particular NACL. If you were to amend, for example, hmpps-production-data-nacl, everything that connects via that NACL will be impacted. Nothing that connects from HMPPS through this NACL will have access.
+
+> Additionally we are only making the changes manually to prevent the code being updated to the detriment of subsequent repairs. Assuming the issue is resolved we would need to rerun the original terraform code to reset everything to it's previous state. Any changes made to the code could cause us issues in being able to do this. It is likely that when applying the code there will be an warning shown stating that some changes have been made outside terraform. This is not something to worry about as all access will be reset to the previous state.
```
ep-93, line 30: a warning

source/runbooks/revoke-network-access.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+The process needs to be done for the 3 NACLs listed below. Examples for production are shown here:
+
+- `live_data-data`
+- `live_data-private`
+- `live_data-public`
+
+1. Connect to `core-shared-services`.
+2. Click on the checkbox for the one you want to change then select **outbound rules**.
+3. Add a new Deny for **0.0.0.0/0**, rule number 211, which will prevent access to the items listed after this point. As rule 210 allows access inside the VPC, all rules above allow traffic to leave the VPC.
```
ep-93, line 171: Should deny be capitalised?

ep-93 (changes requested):

ep-93 (approved):

ep-93 (approved):

PR #6403, opened 2024-03-05
Remove forced suffix for ECR REPOs

georgepstaylor:

Remove the force suffix `-ecr-repo` from the repo name.

ECR does call it the repo name, but it is technically the image name.

Removes the forced suffix - adds a ignore on the name so that existing repos are unffected.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modules/app-ecr-repo/main.tf
+++ b/terraform/modules/app-ecr-repo/main.tf
@@ -1,7 +1,7 @@
 #tfsec:ignore:AWS078 tfsec:ignore:aws-ecr-repository-customer-key
 resource "aws_ecr_repository" "ecr_repo" {
   #checkov:skip=CKV_AWS_51:Skip Tag Mutable requirement we want tags to be overwritten   
-  name                 = "${var.app_name}-ecr-repo"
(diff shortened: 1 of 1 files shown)


dms1981:
For context...
> Not really a problem per se, but..
If I've got an image called delius-foobar and I build it and push it to both ECR and another registry, say GHCR.
I can push it to GHCR as [ghcr.io/ministryofjustice/](http://ghcr.io/ministryofjustice/)<repo_name>/delius-foobar:latest
but to ECR, I have to push it as  <accountid>.[dkr.ecr.eu-west-2.amazonaws.com/delius-foobar-ecr-repo:latest](http://dkr.ecr.eu-west-2.amazonaws.com/delius-foobar-ecr-repo:latest)
So the image name has to include -ecr-repo which seems strange as it's the image name, not the repo name. But in ECR, it's kind of both? Personally I feel like it'd make more sense if the image/repo name didn't include any suffix at all, and given the registryUri already has .dkr.ecr it seems redundant anyway.

dms1981 (approved):

Issue #6369, opened 2024-03-01
Remove `electronic-monitoring-data-preproduction` account


mikereiddigital:
### User Story

As a modernisation platform engineer I need to remove the above member environment. Documentation in the team handbook covers this - https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/deleting-an-environment.html#deleting-an-environment-aws-account.

This was requested by Matthew Price on 1/03/24 at 11am and confirmation should be sought before commencing with the deletion.

### Value / Purpose

Removing unwanted environments reduces cost & maintenance overheads.

### Useful Contacts

Matthew Price - matthew.price2@digital.justice.gov.uk

### Definition of Done

Example - 
- [ ] Confirm that requester is in the team 
- [ ] Confirm with user that the environment deletion can go ahead.
- [ ] Undertake deletion as per team documentation.
- [ ] Confirm with user that environment was deleted.
- [ ] Update documentation

PR #6421, opened 2024-03-07
Removing electronic-monitoring-data-preproduction Account

sukeshreddyg:

#6369 

Removes electronic-monitoring-data-preproduction account


markgov (approved):
LGTM

PR #6472, opened 2024-03-13
try setting secure baseline provider to adaptive mode

richgreen-moj:

As documented here: https://mojdt.slack.com/archives/C013RM6MFFW/p1710322520089769

Tells provider to use adaptive mode which will hopefully make it retry making API calls rather than erroring once a set limit has been reached.

## How has this been tested?

It's not - I'm not sure of another way to test it locally as it's the bootstrap code. My plan would be to monitor how well this performs when running the worflow.

- [X] All checks have passed


SteveLinden (approved):
LGTM

Issue #6163, opened 2024-02-08
Protective Monitoring: Integration 


SimonPPledger:
### User Story

As a product manager 
I need to ensure that we have an appropriate level of protective monitoring
So that we have as secure a service as possible.

This ticket is to work with the protective monitoring team to capture the following levels of detail - for PRODUCTION only.
We are looking to provide the following info:

VPC flow logs 
Networking live data
Network Firewall logs
Route 53

First step will me a conversation with Leo 

### Value / Purpose

This is to help security highlight issues

### Useful Contacts

leonardo.marini@justice.gov.uk

### Additional Information

For the integration work:
1) Route 53 - this should just require a configuration change in route 53

2) We then need to get the following  2 logs into a singel group in the tool (see below for previously used terraform):

- VPC flow logs 
- Networking live data

3) And these logs into a separate group

- Network Firewall logs

This is the terraform code used by the LAN team. 
[module](https://github.com/ministryofjustice/network-access-control-infrastructure/tree/main/modules/kinesis_firehose_xsiam)

### Definition of Done

-  [ ] Route 53 config config changes made
-  [ ] Route 53 info ingested by monitoring tool
- [ ] Log files made available to the the monitoring tool
- [ ] Log files ingested by monitoring tool
- [ ] User docs have been updated
- [ ] Another team member has reviewed

davidkelliott:
Route53 logs are also in cloudwatch, so we might want to do one of the other cloudwatch integrations first, then when we start to do route53 we will have the knowledge of the cloudwatch integration already so it may help decide which method is better.

dms1981:
Hi @SimonPPledger. Just so I'm clear, this would cover us _exporting_ log information into a remote tool?

We can do that for the VPC flow logs with a little bit of coordination with the requestor.

For Route53 I'd need to know exactly what they're after. There are a few different kinds of Route 53 logs - resolve log queries and public dns queries are the sort of thing I'd assume the requestor would want.

Likewise I'm confident we can give them the Network Firewall logs they're after, but it's worth understanding what kind of granularity they want from the logs. They may, for example, want logs for traffic that we deny (but do not log at present).

* https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-firehose.html

mikereiddigital:
I've been through and identified the following resources that are relevant:


```
Flow Logs: (core-vpc-production) (excluding default)

[...]

PR #6485, opened 2024-03-14
Adds a new local module (firehose) that builds the firehose delivery stream & associated resources.

mikereiddigital:

This PR adds a new local module for the creation of firehose delivery stream resources  and its use by the three network firewall inspection logs plus the flow logs for the VPCs. Also adds data items for secrets and their associated locals.

Note that there have been small amendments to the firewall-logging & vpc-inspection modules with additional variables required by firehose.

These changes cover the 2nd part of the log data sharing requirement that is set out in ticket 6163 - https://github.com/ministryofjustice/modernisation-platform/issues/6163

## How has this been tested?

As there is no sandbox or test environment for these changes, the testing has been done based on the plan output.

No resources are being updated or deleted and so there should be no impact on existing services.

- [x] All checks have passed

{Please write here}


terraform/environments/core-network-services/secrets.tf
```diff
@@ -21,6 +21,99 @@ data "aws_secretsmanager_secret_version" "pagerduty_integration_keys" {
   secret_id = data.aws_secretsmanager_secret.pagerduty_integration_keys.id
 }
 
+
+# Data for Firehose Endpoint URL & Key that are held in secrets manager.
+
+# 1. Secrets for Firewall-related log data.
```
dms1981, line 27: Could these secrets be refactored to use a for_each declaration? You might need to think about how to supply it with values, but it would make for cleaner code.

terraform/modules/firewall-logging/variables.tf
```diff
@@ -8,7 +8,22 @@ variable "fw_arn" {
   type        = string
 }
 
+variable "fw_name" {
+  description = "Name of firewall to identify whether live or non-line"
```
richgreen-moj, line 12: Not a big issue should this say "non-live" rather than "line"?

dms1981 (commented):
As a starter, I don't think I would go with the approach of nesting the firehose endpoint module into the existing firewall-logging modules. I would call it separately and supply it with the relevant `log-group` as an output from the firewall or inspection vpc module.
I think that this nested approach can make understanding the code harder; where you call a module, supply values into that module, and then inside that module pass those values on again into the firehose module. Given that the interface with firehose is a separate thing that we're doing, I would prefer to see it as an independent module.

dms1981 (approved):

PR #6508, opened 2024-03-18
DSOS-2635: add firewall rules for FixNGo DCs

drobinson-moj:

The new DCs need connectivity to DOM1 DCs and Mojo devices also need to reach.

Adds firewall rules.  Note this is in live_data VPC so updated terraform accordingly.
Hardcoded to the specific IPs of the 2 DCs.

## How has this been tested?

I can see current connectivity is triggering alert on firewall inspection logs so this should fix

No

- [x] All checks have passed

{Please write here}


dms1981 (approved):
LGTM! Thanks for the update to the local that creates sets.

PR #6511, opened 2024-03-18
Add acm:DeleteCertificate permission to developer role

ASTRobinson:
as per Slack request https://mojdt.slack.com/archives/C01A7QK5VM1/p1710772027207819 user would like temp permissions to delete certificates.

(also sorted the acm permissions alphabetically)


davidkelliott:
This will add it to all users and add a risk that someone could delete a cert in prod

PR #6530, opened 2024-03-19
Allow IP & port sets for nwfw

dms1981:

Some customers require multiple destination ports for single source/destination pairings.

Adds `dynamic {}` blocks for `IP_SETS` and `PORT_SETS` to module responsible for creating AWS Network Firewall policy & rule groups.

Users can reference these sets in rules like so:
```
- destination      = "1.1.1.1/32"
- destination_port = "$NAME-OF-PORT-SET"
- direction        = "FORWARD"
- protocol         = "TCP"
- source           = "ANY"
- source_port      = "ANY"
```

## How has this been tested?

Tested manually, tested with an apply from this branch.

{Please write here}

- [x] All checks have passed

Raised in reference to https://github.com/ministryofjustice/modernisation-platform/pull/6508


markgov (approved):
LGTM

PR #6553, opened 2024-03-20
remove and add collaborator

Lee-250:

We've recently had Kaleb Sofer leave our team and Madhu Kadiri has joined. I have therefore replaced the names in the collaborators.json

### DIFF
--- a/collaborators.json
+++ b/collaborators.json
@@ -547,15 +547,15 @@
       ]
     },
     {
-      "username": "kaleb.sofer",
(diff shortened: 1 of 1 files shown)


dms1981 (approved):

dms1981:
Hi @Lee-250 - you'll need us to do a bit of setup for the new account. Can you get in touch with us via #ask-modernisation-platform on Slack once this is merged in?

PR #6580, opened 2024-03-21
Amended length of name for the firehose resource_prefix

Khatraf:

#6163


mikereiddigital (approved):
lg2m

Issue #6486, opened 2024-03-14
Bug: Error listing tags for Config Rule in Secure-baselines


richgreen-moj:
### Expected Behavior

The `Terraform: Scheduled Baseline` workflow hasn't run completely successfully for about a week. In particular the `secure-baselines` jobs have not all been succeeding due to an error when trying to list tags for config rule sin a terraform plan.

### Actual Behavior

Example of error:
``Error: listing tags for Config Config Rule (arn:aws:config:eu-west-2:767123802783:config-rule/config-rule-mzxknb): operation error Config Service: ListTagsForResource, failed to get rate limit token, retry quota exceeded, 3 available, 5 requested``

See [Slack thread ](https://mojdt.slack.com/archives/C013RM6MFFW/p1710322520089769)for more info.

richgreen-moj:
See this TF AWS provider bug which seems to be related https://github.com/hashicorp/terraform-provider-aws/issues/34669

richgreen-moj:
Also see this... https://github.com/hashicorp/terraform-provider-aws/issues/36024

richgreen-moj:
I have raised https://github.com/ministryofjustice/modernisation-platform/pull/6494 to pin the TF provider to v5.38.0 as we didn't have this sort of failure occurring previously. This worked successfully under this run ... https://github.com/ministryofjustice/modernisation-platform/actions/runs/8341764086
See v5.38.0 in TF init....
https://github.com/ministryofjustice/modernisation-platform/actions/runs/8341764086/job/22828878548#step:6:144

richgreen-moj:
I've run a job on main branch (so latest TF provider v5.41.0 e.g. https://github.com/ministryofjustice/modernisation-platform/actions/runs/8342287044/job/22830537939#step:6:145) and it's failed again 

https://github.com/ministryofjustice/modernisation-platform/actions/runs/8342287044 96 out of 175 secure-baseline jobs failed.

richgreen-moj:
PR #6494 has been merged to main which has pinned the provider for the secure-baseline code to `v5.38.0`

This has added some stability to the job; it runs through on a single attempt. 

I'm considering whether or not that's enough for now, I think that something must have been introduced by the provider which has been acknowledged here https://github.com/hashicorp/terraform-provider-aws/issues/36024 and will hopefully be addressed in future.

[...]

PR #6598, opened 2024-03-25
Removed pinned provider version & increased max_retries for Secure Baselines

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/6486

I've reverted the code that was pinning the version of the AWS TF provider back to `v5.38.0`

The jobs are now picking up the latest version `v5.42.0` which has had an update that stops client side rate limiting which stops you getting the original error we encountered e.g. 
`failed to get rate limit token, retry quota exceeded, 3 available, 5 requested`

But we were still encountering this error e.g. 
`ListTagsForResource, exceeded maximum number of attempts, 25, https response error StatusCode: 400, RequestID: c17f1153-756e-4968-bd57-e38a51792366, api error ThrottlingException: Rate exceeded`

There is an optional [max_retries](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#max_retries) setting that can be applied to the provider so you can raise it from the default of `25`.
In this PR I have set it to `100` which has been tested successfully on two subsequent runs of the scheduled baseline job.

## How has this been tested?

It worked ... https://github.com/ministryofjustice/modernisation-platform/actions/runs/8422101284

twice... https://github.com/ministryofjustice/modernisation-platform/actions/runs/8422873826

- [X] All checks have passed

This might just be a setting we keep our eye on over time as with all the scheduled baseline code which is growing as the platform grows.

### DIFF
--- a/terraform/environments/bootstrap/secure-baselines/providers.tf
+++ b/terraform/environments/bootstrap/secure-baselines/providers.tf
@@ -14,8 +14,9 @@ provider "aws" {
 
 # Region specific providers for the workspace. Required for bootstrapping resources in enabled regions.
 provider "aws" {
-  alias  = "workspace-eu-central-1"
-  region = "eu-central-1"
+  alias       = "workspace-eu-central-1"
+  region      = "eu-central-1"
+  max_retries = 100
 
   assume_role {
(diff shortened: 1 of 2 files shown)


mikereiddigital (approved):
lg2m

PR #6610, opened 2024-03-26
Adds firehose stream resources for R53 resolver logs plus other minor changes.

mikereiddigital:

This covers the last set of firehose stream transfers to SecOps - namely the resolver logs for the member account shared VPCs that is managed in core-vpc. It also covers a number of other changes including the use of a more descriptive common_attribute value plus other simplifications.

See above.

## How has this been tested?

Local plan with the core-vpc workspaces

{Please write here}

No resources are being deleted - mainly additions and some changes for the new common_attribute var.

- [ ] All checks have passed

{Please write here}


mikereiddigital:
Note - there is an issue whereby the core-vpc changes are not being identified but are with a local plan. We are investigating why this is.

SteveLinden (approved):

Issue #6390, opened 2024-03-05
Follow-Up Ticket for Inactive Collaborator IAM Account Management


sukeshreddyg:
### User Story

As a mod platform engineer
 I want to create follow-up tickets to implement automated processes for disabling and deleting inactive user accounts So that we don’t have old accounts in the mod platform account.

### Value / Purpose

To improve security by automating the management of inactive IAM accounts. By implementing these processes, we can efficiently disable and delete inactive accounts, reducing the risk of unauthorised access and ensuring that our IAM environment remains secure. 

### Definition of Done

- [ ] Create a job to disable IAM user accounts that have been inactive for more than 120 days.
- [ ] Create a job to raise a PR and tag the user to the PR for deleting accounts inactive for more than 7 months
- [ ] User docs have been updated
- [ ] Another team member has reviewed
- [ ] Tests are green

PR #6743, opened 2024-04-10
Implement Feature to Disable Console Login and Access Keys for `Inactive Collaborator Users`

sukeshreddyg:

This PR addresses the implementation of a feature to disable console logins and access keys for `inactive collaborator users` after 120 days of inactivity. #6390 
 
This PR introduces a new job that identifies collaborator users with inactivity of 120 days or more and disables their console login and access keys. 

## How has this been tested?

Tested this new functionality in the sprinkler environment to validate its effectiveness and reliability under various scenarios.

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/check-inactive-collaborators.yml
+++ b/.github/workflows/check-inactive-collaborators.yml
@@ -6,7 +6,6 @@ on:
 env:
     AWS_REGION: "eu-west-2"
     ENVIRONMENT_MANAGEMENT: ${{ secrets.MODERNISATION_PLATFORM_ENVIRONMENTS }}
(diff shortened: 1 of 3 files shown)


dms1981 (commented):
A few comments.

* Using an environment variable for `{SCRIPTS_PATH}` feels a bit redundant as we're not likely to change where we store scripts and use simple hardcoding for all other actions. I'd personally ditch this and stay consistent.
* There's a small typo for `thresold` - it should be `threshold`.
* Personally I'd move the inline bash script for disabling console logins to its own script rather than in the action. That would again stay consistent with the existing approach.

ewastempel (approved):
Very nice!

PR #6748, opened 2024-04-11
Implement User Deletion Workflow for Inactive Collaborators

sukeshreddyg:

This PR implements the workflow for scheduling user deletion for inactive collaborators whose inactivity is more than or equal to 180 days. #6390 

This PR adds the necessary logic and steps to identify inactive collaborators and raise a GitHub issue for their deletion. By automating this process, we ensure that inactive collaborators are promptly identified and managed, enhancing the security and efficiency of AWS accounts.

### DIFF
--- a/.github/workflows/check-inactive-collaborators.yml
+++ b/.github/workflows/check-inactive-collaborators.yml
@@ -4,8 +4,7 @@ on:
   workflow_dispatch: 
 
 env:
-    GH_TOKEN: ${{ secrets.TERRAFORM_GITHUB_TOKEN }}
-    SECRET: ${{ secrets.GITHUB_TOKEN }}
+    GH_TOKEN: ${{ github.token }}
     AWS_REGION: "eu-west-2"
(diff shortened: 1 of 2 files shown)


mikereiddigital (approved):
lg2m

mikereiddigital (dismissed):
lg2m

.github/workflows/check-inactive-collaborators.yml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_dispatch: 
 
 env:
+    GH_TOKEN: ${{ secrets.github_ci_user_pat }}
```
dms1981, line 7: Do you specifically need to use the GitHub CI User PAT, or can you use automatic token authentication? (https://docs.github.com/en/actions/security-guides/automatic-token-authentication)

sukeshreddyg: I tried both `github.token` and `secrets.GITHUB_TOKEN`, but neither of them worked.

PR #6834, opened 2024-04-19
📝  fix markdown lint error's

ASTRobinson:
spotted a few lint errors on the Megalinter run while looking into another issue... 

no content changes, only linting suggestions/errors and formation improvements.

https://github.com/ministryofjustice/modernisation-platform/actions/runs/8748550257/job/24008767485#step:6:86


ewastempel (approved):
Nice!

PR #6840, opened 2024-04-22
Fix new account logic in New Environment Workflow

richgreen-moj:

This PR fixes a couple of issues with the `get new account(s)` job in the `new-environment.yml` workflow.

I noticed when deleting an account which includes updates to the `environment/*.json` files _and_ `environments-networks/*.json` files that the `git diff` command was also picking up these changes as it's filters included `'*.json'`

This job is only meant to pick out a list of updated `environment/*.json` files so that it can create a list (or array) of application names to loop through in the subsequent jobs e.g. `Run delegate-access`

I also noticed that when multiple newline separated values were being passed to the `accounts` variable that the subsequent loop was reading this as one value rather than an array. Also the logic that checked whether the `accounts` variable was empty did not work when multiple values were passed causing a `binary operator expected` error.

**Example where this occured:**
As a result of PR https://github.com/ministryofjustice/modernisation-platform/pull/6835 which just updated some formatting.

```
accounts=$(echo "ZW52aXJvbm1lbnRzLW5ldHdvcmtzL3BsYXRmb3Jtcy10ZXN0CmRlbGl1cy1pYXBzCg==" | base64 --decode)

echo "ZW52aXJvbm1lbnRzLW5ldHdvcmtzL3BsYXRmb3Jtcy10ZXN0CmRlbGl1cy1pYXBzCg==" | base64 --decode
environments-networks/platforms-test
delius-iaps
```
errored with..
```
/home/runner/work/_temp/e26972ed-8189-4df8-9be6-a88ba6f7c10c.sh: line 2: [: environments-networks/platforms-test: binary operator expected
[+] There were no AWS member accounts to process
```
https://github.com/ministryofjustice/modernisation-platform/actions/runs/8780835226/job/24094340360#step:7:26

https://github.com/ministryofjustice/modernisation-platform/actions/runs/8740977849/job/24015465001#step:6:16

I've removed the reference to `'*.json'`in the `git diff` so only application names are filtered. I've also added a `readarray` command so that the `accounts` value is read as an array and added double square brackets in the if statement to ensure the `if [[ ! -z ${accounts} ]]` logic works correctly.

## How has this been tested?

I created a branch with a test job to verify the logic would work, examples of runs:
https://github.com/ministryofjustice/modernisation-platform/actions/runs/8784754179/job/24103650888

This commit made several changes to various `environments/*.json` files and `environments-networks/*.json` files and has read the accounts array and is only running delegate access for the list of account names only.

```
Run readarray -t accounts <<< "$(echo "ZWxlY3Ryb25pYy1tb25pdG9yaW5nLWRhdGEKZXF1aXAKZXJpYwpleGFtcGxlCg==" | base64 --decode)"
[+] Running delegate-access baseline for account electronic-monitoring-data
[+] Running delegate-access baseline for account equip
[+] Running delegate-access baseline for account eric
[+] Running delegate-access baseline for account example
```

{Please write here}

- [X] All checks have passed

{Please write here}


SteveLinden (commented):
Notice there are a few "[[" and "]]" - are these needed?
e.g.  if [[ ! -z ${accounts} ]]

Notice they are all proceeded by a "readarray" so it might be required.

davidkelliott (approved):
LGTM

ewastempel (approved):

PR #6849, opened 2024-04-23
new bedrock policy

markgov:

There is a missing bedrock policy for the sandbox role which is needed by some users

This PR creates and added that policy to the sso role and i will follow up with a PR for the collaborators later on

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

Issue #6615, opened 2024-03-26
Review the use of -targets in the core-vpc workflows


mikereiddigital:
### User Story

As a platform engineer I need to be able to see the full scope of changes identified when reviewing a PR for core-vpc workspaces. As it stands, some changes may not be covered by the plan and so the reviewer will not see the full picture without reverting to the use of local plans.
This ticket would cover the review of the current scope of targets and consider whether their scope should be increased or whether they should be removed.

### Value / Purpose

Accurate terraform change plans are necessary when reviewing PRs.

### Definition of Done

- [ ] Review list of targets
- [ ] Amend targets where appropriate

mikereiddigital:
Current terraform documentation on the use of -target:

```
This targeting capability is provided for exceptional circumstances, such as recovering from mistakes or working around Terraform limitations. It is not recommended to use -target for routine operations, since this can lead to undetected configuration drift and confusion about how the true state of resources relates to configuration.

Instead of using -target as a means to operate on isolated portions of very large configurations, prefer instead to break large configurations into several smaller configurations that can each be independently applied. [Data sources](https://developer.hashicorp.com/terraform/language/data-sources) can be used to access information about resources created in other configurations, allowing a complex system architecture to be broken down into more manageable parts that can be updated independently.
```

mikereiddigital:
Dave Elliot has confirmed that addition of the -target flags was avoid plan errors. Will test the workflow of core-vpc-development-deployment without the targets to confirm whether they can be removed. If if the error persists I will add targets for the new firehose modules.

mikereiddigital:
Having removed the -target flags from the development plan workflow, the output is the same as test. So applying the same changes to the other workflows.

mikereiddigital:
Final PR - https://github.com/ministryofjustice/modernisation-platform/pull/6873

mikereiddigital:
Moving to done.

PR #6869, opened 2024-04-24
Test to confirm the core-vpc-development-deployment plan runs correctly without the -target flags.

mikereiddigital:
Removes the -target flags from the core-vpc-development-deployment job in order to confirm its correct working. If the plan runs correctly then other -target flags will be removed from the plan for the other accounts. If it fails the change will be reverted. 

https://github.com/ministryofjustice/modernisation-platform/issues/6615

As -targets limits the scope of the plan, any new resources created via a module call will be excluded from the plan. This is to confirm whether removing the targets doesn't result in plan errors.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #6878, opened 2024-04-25
Add AWSVPCLog role and related resources for VPC logging

sukeshreddyg:
This PR addresses the issue caused by the removal of the baseline module, which contained the AWSVPCLog role. To resolve this, I've added the AWSVPCLog role and related resources directly to the core accounts.


mikereiddigital (approved):
lg2m

PR #6910, opened 2024-04-30
feature/6753 - firewall rule to allow traffic between laa LZ account and new test account

Kudzai-moj:

Customer needs traffic to flow between the ccms-ebs-upgrade test account and their LZ test account. In order to do this, i have followed the process below,

https://user-guide.modernisation-platform.service.justice.gov.uk/concepts/networking/network-firewall.html#amending-rules

The PR amends the rule that allows connectivity between the 2 accounts

- [X] All checks have passed

{Please write here}


dms1981 (approved):
Approved, but if we're permitting any traffic should we clean up other rules for `laa-lz-test` to `laa-test`?

Issue #2403, opened 2022-10-13
Align core-vpc NACLs with member-vpc NACLs


dms1981:
## User Story

As a modernisation platform engineer 
I want to align how VPC NACLs are provisioned to both member and core VPCs
So that I can consistently apply network ACL rules across VPCs

## Value

At present, network ACL rules are provisioned through a recently refactored module in the modernisation platform for member VPCs. By refactoring how network ACL rules are provisioned for core VPCs, we can tighten our current security posture, and enable the consistent application of network ACL rules

## Questions / Assumptions / Hypothesis

* Can we refactor the `vpc-nacls` module to complete this with the minimum of effort?
I think this is possible, but could be minorly disruptive. By adjusting the rule _numbers_ we can could do the following:
```
  # 1000 = intra-vpc traffic
  # 2000 = inter-vpc traffic (dynamic rules)
  # 3000 = dynamic address ranges
  # 4000 = deny east-west
  # 5000 = private address ranges
  # 6000 = https internet access
  # 7000 = access from internet
```

* Would we prefer to create a separate module? `core-vpc-nacls`?
My indicative feel here is that this isn't quite right. While it would work, it could just as easily be done by refactoring `vpc-hub`

* Would we prefer to refactor the `vpc-hub` module?

## Definition of done

- [x] ACL changes reviewed and approved
- [x] Core VPC NACLs tightened
- [x] another team member has reviewed

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)
[Refactor NACLs to prevent East-West traffic and allow North-South](https://github.com/ministryofjustice/modernisation-platform/issues/2010)

PR #6939, opened 2024-05-02
Amend NACLs applied to public subnets

dms1981:

#2403 

Observed that rules for all private subnets were also being applied to public subnets. Re-sited block for RDP traffic so that it is only applied to public subnets, and removed rules for non-public subnets from public subnets

## How has this been tested?

Checked against production VPC flow logs to ensure no services except HTTPS were listening in the `1-1023` TCP port range.

Deploy through CI

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-vpc/locals.tf
+++ b/terraform/environments/core-vpc/locals.tf
@@ -9,6 +9,9 @@ locals {
   environment_management     = jsondecode(data.aws_secretsmanager_secret_version.environment_management.secret_string)
   pagerduty_integration_keys = jsondecode(data.aws_secretsmanager_secret_version.pagerduty_integration_keys.secret_string)
 
+  # This carries out a simple extraction to retrieve the trailing environment name (e.g. development)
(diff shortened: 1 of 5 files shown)


ASTRobinson (approved):

dms1981:
Moved this back to draft as I identified a case where traffic was moving directly between the public subnet and a private address range

dms1981:
After looking through the VPC flow logs I could see a single exception. I'd need to confer with the responsible team, but something like this would be the likely way to account for this exception:
```
locals {
  current_environment = substr(terraform.workspace, length(local.application_name), length(terraform.workspace))
}

resource "aws_network_acl_rule" "additional-public-rule" {
  network_acl_id = module.vpc-nacls[format("hmcts%s",local.current_environment)].public.id
  cidr_block     = ["?.?.?.?/??"]
  ...
}
```
I don't 100% love the format expression, but with `current_environment` set up this way it's consistent with other locals used to scrape the environment name from the workspace.

dms1981:
Added some extra handling for `xhibit-portal` requirements, but it's a very first-pass solution at present.

PR #6956, opened 2024-05-08
Align NACLs used in `vpc-hub` module with `vpc-nacls` approach

dms1981:

#2403 

This PR amends how network ACLs are created in the `vpc-hub` module. When the story was originally written this would have used the same `vpc-nacls` module that is currently used by the `modernisation-platform-terraform-member-vpc` code and applied to `core-vpc-*` environments.

This simplifies how and where NACL rules are defined. The content is now located in `locals.tf` under a series of simple maps - `static_acl_rules`, `public_access_acl_rules`, and `transit_gateway_acl_rules`.

As much as possible this is identical to the content in the `vpc-nacls` module to ensure consistency, although there may be future opportunities to trim these rules if we invest time in inspecting VPC flow logs for potential exceptions.

No changes to the `inspection` VPCs - those in `core-network-services` - are carried out in this PR, although they could also be in scope as part of the story. However, they're a little trickier. 

The inline inspection VPCs used as egress VPCs for traffic from the per-business-unit VPCs in `core-vpc-*` need to be very open because of their purpose; inspecting traffic with AWS Network Firewalls and conducting NAT in AWS NAT Gateways on the way out to the internet.

The `external-inspection` VPC similarly needs to be open because it's an inspection point for traffic moving internally between the MOJ TGW and the MP TGW / MP VPCs. Given the wide range of customers we have, NACLs aren't a better choice for securing flows compared to the AWS Network Firewall that inspects traffic in this position.

In a nutshell, this PR does the following:
* Private subnet and Data subnet NACLs allow unrestricted intra-VPC traffic, and unrestricted private address range traffic.
* Public subnet NACL allows unrestricted intra-VPC traffic, and fairly unrestricted traffic out to the internet.
* Transit Gateway subnet NACL allows unrestricted traffic as per the AWS design guidance

It also removes the old code used to create NACLs.

## How has this been tested?

- [x] Conducted a local terraform plan
- [x] Checked existing flow logs over a reasonable timeframe to identify if this will break existing functionality.

Deploy through CI

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modules/vpc-hub/locals.tf
+++ b/terraform/modules/vpc-hub/locals.tf
@@ -1,0 +1,205 @@
+locals {
+  # 1000 = intra-vpc traffic
+  # 2000 = inter-vpc traffic (dynamic rules)
+  # 3000 = deny east-west
+  # 4000 = private address ranges
+  # 5000 = internet access
+  # 6000 = public address ranges (dynamic)
+  # 7000 = access from internet
+
+  static_acl_rules = {
+    allow_vpc_cidr_in = {
+      cidr_block  = var.vpc_cidr
+      egress      = false
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 1000
+      to_port     = null
+    },
+    allow_vpc_cidr_out = {
+      cidr_block  = var.vpc_cidr
+      egress      = true
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 1000
+      to_port     = null
+    },
+    allow_10-0-0-0_in = {
+      cidr_block  = "10.0.0.0/8"
+      egress      = false
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 4000
+      to_port     = null
+    },
+    allow_10-0-0-0_out = {
+      cidr_block  = "10.0.0.0/8"
+      egress      = true
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 4000
+      to_port     = null
+    },
+    allow_172-16-0-0_in = {
+      cidr_block  = "172.16.0.0/12"
+      egress      = false
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 4100
+      to_port     = null
+    },
+    allow_172-16-0-0_out = {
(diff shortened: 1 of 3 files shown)


terraform/modules/vpc-hub/locals.tf
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  public_access_acl_rules = {
+    allow_vpc_cidr_in = {
+      cidr_block  = var.vpc_cidr
+      egress      = false
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 1000
```
mikereiddigital, line 146: Should this rule number be 6000?

terraform/modules/vpc-hub/locals.tf
```diff
@@ -0,0 +1,227 @@
...
+      to_port     = null
+    },
+    allow_vpc_cidr_out = {
+      cidr_block  = var.vpc_cidr
+      egress      = true
+      from_port   = null
+      protocol    = "-1"
+      rule_action = "allow"
+      rule_number = 1000
```
mikereiddigital, line 155: As above.

terraform/modules/vpc-hub/locals.tf
```diff
@@ -0,0 +1,227 @@
...
+      to_port     = null
+    },
+    allow_10-0-0-0_in = {
```
mikereiddigital, line 29: Are the network firewall rules in place that prevent hopping between vpcs?

dms1981: Yes and no. Traffic inside a "domain" like prod/preprod and dev/test hits the MP Transit Gateway and goes directly to the destination VPC.
EG. `laa-test` can send to `platforms-development` without being passed through the `core-network-services["external-inspection"]` VPC.
EG. `laa-development` can send to `core-network-services["non_live_data"]` without being passed through the `core-network-services["external-inspection"]` VPC.

When traffic wants to move between those domains - `laa-development` to `laa-production` or `core-shared-services["live_data"]` then it would move through the MP Network firewall in the `core-network-services["external-inspection"]` VPC.

terraform/modules/vpc-hub/locals.tf
dms1981: As above, this would again cover traffic moving internally in the VPC.

ASTRobinson (approved):

richgreen-moj (approved):

PR #6959, opened 2024-05-09
🇫🇷  😢 Unbootstrapping France.

julialawrence:

https://github.com/ministryofjustice/data-platform/issues/4222
The code enabling Paris region is breaking secure baseline job for Sprinkler because the account isn't enrolled in eu-west-3 security hub. Once that is fixed, the Paris changes can be reapplied in full.

The Paris region has been removed from the list of bootstrapped regions but because some of the resources have been created, the provider region change needs to persist as switching it back to the `eu-west-2` region will stop them being found and destroyed. 

Once sprinkler apply ran once, this should be revertable too.

## How has this been tested?

It hasn't been unfortunately.

An apply on `sprinkler` should be enough to see if this undid the damage. This PR does not need merging.

- [ ] All checks have passed

{Please write here}


ASTRobinson:
sprinkler reverted, closing PR

Issue #6492, opened 2024-03-14
🧹 Clean up unit tests


dms1981:
### User Story

As a Modernisation Platform Engineer
I want to freshen up our unit tests
So that we have more consistent standards in how we write and use unit tests

### Value / Purpose

As part of a [recent spike](https://github.com/ministryofjustice/modernisation-platform/issues/6334) some inconsistency was noted in our unit tests. This story should achieve the following:
* Use of a common name for unit tests. EG `test/unit_test.go`
* Check that tests cover all important functions in module
* Conduct a hygiene pass on code in `test/unit-test/*.tf`
  * EG, ensure some randomness is used in resource names
  * See [here](https://terratest.gruntwork.io/docs/#testing-best-practices) for other thoughts on best use of Terratest

### Definition of Done

- [x] Cleanup pass completed
- [x] Unit tests pass

connormaglynn:
## Some useful commands to find  

You can use the following command to get all repositories with the string "modernisation-platform-terraform" in them 👇 

```bash
$ gh api -X GET orgs/ministryofjustice/teams/modernisation-platform/repos -f 'per_page=100' | jq -r '.[] | select(.archived == false) | .name' | grep modernisation-platform-terraform
modernisation-platform-terraform-iam-superadmins
modernisation-platform-terraform-environments
...
```

Using the output here we can clone all the repositories locally 👇 

```bash
$ MP_REPOS=$(gh api -X GET orgs/ministryofjustice/teams/modernisation-platform/repos -f 'per_page=100' | jq -r '.[] | select(.archived == false) | .name' | grep modernisation-platform-terraform)
$ while IFS= read -r repo ; do git clone "https://github.com/ministryofjustice/${repo}.git"; done <<< "$MP_REPOS"
```

We can then check which repositories have tests locally with 👇 

```bash
$ MP_REPOS=$(gh api -X GET orgs/ministryofjustice/teams/modernisation-platform/repos -f 'per_page=100' | jq -r '.[] | select(.archived == false) | .name' | grep modernisation-platform-terraform)
$ while IFS= read -r repo ; do cd "${repo}" && printf "- https://github.com/ministryofjustice/${repo} Files: " && find test | xargs echo  && cd ../  ; done <<< "$MP_REPOS"
```

I've separated the output of the last command into repos with tests and repos without them

**Repos with tests**
[...]

PR #6994, opened 2024-05-14
Removed references to the S3 replication role

Khatraf:

#6492 

The replication role module has been integrated into the main s3 bucket module so I removed the references to it and updated the source with the new release for the S3 bucket module.


ep-93 (approved):
Plan ran through with Khatra, output looks good, and worst case scenario its only a role.

Approving, Khatra is going to run and apply locally to one instance where this occurs to test, and then will merge in if successful.

Issue #6991, opened 2024-05-13
CloudFormation Stack deployment using Terraform


vc13837:
### User Story

As an Ops Engineer migrating the application CWA
I need to deploy the WAFv2 Web ACL using CloudFormation
So that the migration of CWA can continue without changes to existing security rules

### Value / Purpose

I would like to deploy WAFv2 Web ACL in Modernisation Platform. However with Terraform there is an issue where the rules cannot go more than 3 levels deeps - https://github.com/hashicorp/terraform-provider-aws/issues/15580, thus current work around is to deploy using CloudFormation.

The only resource being deployed via CloudFormation is `AWS::WAFv2::WebACL`, thus other resource creation can be blocked policy-wise to stop other users using CloudFormation as a workaround to deploy any resources they like.

### Useful Contacts

Vincent Cheung (via Slack please)

### Proposal / Unknowns

Assumption:
- Changes to the security rules of the Web ACL is not acceptable by the application team

### Definition of Done

- [ ] laa-aws-infrastructure GitHub team (i.e. LAA Ops) can use Terraform via the infra pipeline to deploy CloudFormation Stack to create WAF v2 Web ACL

davidkelliott:
Proposed solution:

Add permissions for cloudformation to the memberinfrastructureaccess role
Add [aws_cloudformation_stack](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudformation_stack) to the plan evaluator
User to store cloudformation template in environments repo and deploy with the tf resource above

mikereiddigital:
Have reviewed the list of policy actions for cloudformation in https://docs.aws.amazon.com/service-authorization/latest/reference/list_awscloudformation.html. I have excluded those permissions that pertain to operations on stack resources (e.g. CreateChangeSet) and stack-set operations and instead included only those permissions that deal with the stack itself.

mikereiddigital:
Issues with the `cloudformation:ResouceTypes` variable is resulting in the condition not allowing AWS::WAFv2::WebACL types to be created. Have spent time with AWS reviewing the policy. Will be talking to them again today.

mikereiddigital:
This is a summary of the issue. The current proposed change is to add the following:
`
    statement {
      effect = "Allow"
      actions = [
          "cloudformation:ListStacks",
          "cloudformation:DescribeStacks",
          "cloudformation:DeleteStack",
          "cloudformation:GetTemplate"
      ]
[...]

PR #7162, opened 2024-06-04
DRAFT - Allow creation of Wafv2 ACL resources through CloudFormation

mikereiddigital:
**_Note - this is pending advice from AWS on issues arising from this proposed change._**

Amends aws_iam_policy_document "member-access" and adds permissions that allow only the creation of wafv2:acl resources via an aws_cloudformation_stack. Note that the permissions are limited only to stack-level operations and does not include those actions for stacksets or the creation/exec of changesets.

See issue 6991 - https://github.com/ministryofjustice/modernisation-platform/issues/6991

This change has been requested by the LAA Ops team due to limitations with terraform implementing wafv2:acl rules nested beyond 3 levels. The identified solution is to create the acl via an aws_cloudformation_stack resources. The added statement has been written to only allow the creation of resources of type AWS:WAFv2:ACL via cloudformation so as to avoid the potential for this approach to be used more generally.

## How has this been tested?

This change will be tested in sprinkler first where I will attempt to create both a wafv2:acl and other types of resource. If the tests are successful, the acl will be created via a cloudformation stack but not others.

As this is a supplemental change it should not adversely impact existing resources.

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/iam.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/iam.tf
@@ -253,12 +253,8 @@ data "aws_iam_policy_document" "member-access" {
   statement {
     effect = "Deny"
     actions = [
-      "cloudformation:ListStacks",
-      "cloudformation:DescribeStacks",
       "cloudformation:CreateStack",
-      "cloudformation:UpdateStack",
-      "cloudformation:DeleteStack",
-      "cloudformation:GetTemplate"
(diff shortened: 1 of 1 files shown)


mikereiddigital:
Change applied to sprinkler for testing.

mikereiddigital:
New solution offered by AWS so testing that first before any new PR.

Issue #6888, opened 2024-04-29
Create a reusable Terraform plan/apply workflow for MP


richgreen-moj:
### User Story

As an MP Engineer
I want to create a reusable Terraform plan/apply workflow for MP
So that the many Terraform workflows in the MP repo are easier to maintain going forward

### Value / Purpose

Currently we have a number of workflows that run Terraform for the various elements of the platform. Whilst we do have some modular elements e.g. reusable bash scripts for terraform init/plan/apply, if we want to add further steps to the deployment process (for example; extra testing etc) several workflows would require editing rather than just making one change and having it as an optional step on a reusable workflow.

Essentially this story is to look at replicating the [reusable TF workflow](https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/.github/workflows/reusable_terraform_plan_apply.yml) used in the MP Environments repo and make it work for MP

Once we have this in place it would be easier to customise our deployment process adding features like;
- optionally post terraform plans into PRs 
-  adding extra testing steps etc.

### Additional Information

https://github.com/ministryofjustice/modernisation-platform/issues/6388 - This story was dreamed up as a result of this spike where we discussed having the ability to post TF plans into PRs or have extra testing steps that issue warnings if certain resources are being deleted.

### Definition of Done

- [ ] Current MP Terraform workflows have been analysed for common functions
- [ ] Reusable Terraform Plan/Apply workflow has been developed
- [ ] Reusable Terraform Plan/Apply workflow has been tested
- [ ] All Existing Terraform workflows have been updated to use new Terraform Plan/Apply workflow
- [ ] README has been updated
- [ ] User docs have been updated
- [ ] Another team member has reviewed
- [ ] Tests are green

PR #7264, opened 2024-06-18
Reusable workflow for pipelines

sukeshreddyg:

#6888 

The creation of a reusable workflow resolves the problem of having multiple, redundant workflows for similar tasks. One reusable workflow can be called by various pipelines, eliminating the need to duplicate workflow logic across different projects.

## How has this been tested?

Tested reusable workflow functionality with modernisation-platform-account workflow

- [x] All checks have passed


ep-93 (approved):
Approved based on discussions with Sukesh. Its a big change so will talk about deployment and testing.

PR #7280, opened 2024-06-19
Calm down OPA Validation workflow

dms1981:

`"Open Policy Agent: validate JSON structures"` runs on every PR.

This PR restricts the use of this action to only trigger when `*.rego` and `*.json` files present in a PR.

## How has this been tested?

It has not been tested.

No.

- [x] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/opa-policies.yml
+++ b/.github/workflows/opa-policies.yml
@@ -10,6 +10,9 @@ on:
       - '**.rego'
   merge_group:
     types: [checks_requested]
+    paths:
+      - '**.json'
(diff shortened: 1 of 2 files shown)


Khatraf (dismissed):

dms1981:
We're going to talk about this a bit more; at the moment we require it regardless of whether or not we want it, but I don't think that's quite right. We also may need to consider the merge queue functionality and any adjustments to be made there.

ASTRobinson (approved):

PR #7331, opened 2024-06-25
Panda network changes

SteveLinden:

Changes to make the network changes for Panda - may not be needed but an attempt to fix an error

### DIFF
--- a/environments-networks/platforms-development.json
+++ b/environments-networks/platforms-development.json
@@ -8,7 +8,8 @@
           "data-platform-development",
           "data-platform-apps-and-tools-development",
           "observability-platform-development",
-          "operations-engineering-development"
+          "operations-engineering-development",
+          "panda-cyber-appsec-lab-development"
         ]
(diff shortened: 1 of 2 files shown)


SteveLinden:
> **Please check the plan carefully before deploying these changes.**
> 
> ⚠️ Making changes to the `environments-networks/*.json` files will affect the networking of MP AWS accounts, in particular which VPC/Subnet is being shared in to which environment. For more guidance consult the [documentation](https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/creating-accounts-for-end-users.html#networking-stage).

Understood but there are errors showing up because there are no VPCs. My intention is to run through this to clear the issue and then to remove any subnets it creates. I know this is not how it would normally be done but removing the initial VPC is what is causing this issue I believe. In 99% of cases this will not be an issue as we will be running the networking steps.

ASTRobinson (approved):

PR #7341, opened 2024-06-26
Added the isolated-network true code to the Panda settings

SteveLinden:

Included the missed code of "isolated-network": "true",

Panda should be isolated according to the information in there.


mikereiddigital (approved):
lg2m

PR #7374, opened 2024-06-28
ignore markdown files (e.g. README.md) for terraform deployment workflows

tom-webber:

No existing issue. The current github workflows for terraform deployment are triggered on changes within the terraform directories. This includes changes exclusively to `README.md` files, which won't affect the deployment, and shouldn't trigger the workflow.

This PR adds a `- '!**.md'` condition to the `paths` argument for the triggering actions (see [including and excluding paths syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)). This means that changes to markdown files ending in `.md` (i.e. not `html.md.erb` files) will not trigger these deploy workflows.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/bootstrap-sprinkler.yml
+++ b/.github/workflows/bootstrap-sprinkler.yml
@@ -6,8 +6,8 @@ on:
       - main
     paths:
       - 'terraform/environments/bootstrap/**'
-      - '!**.md'
       - 'terraform/modules/iam_baseline/**'
+      - '!**.md'
       - '.github/workflows/bootstrap-sprinkler.yml'
   workflow_dispatch:
 
--- a/.github/workflows/bootstrap-sprinkler.yml
+++ b/.github/workflows/bootstrap-sprinkler.yml
@@ -6,6 +6,7 @@ on:
       - main
     paths:
       - 'terraform/environments/bootstrap/**'
+      - '!**.md'
       - 'terraform/modules/iam_baseline/**'
       - '.github/workflows/bootstrap-sprinkler.yml'
   workflow_dispatch:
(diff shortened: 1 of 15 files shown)


dms1981:
In its current form this feels overdriven to me; setting `paths` to ignore any markdown, rather than only the markdown in relevant paths is efficient from a number-of-characters perspective but gets away from the purpose of the included paths. I think I'd prefer to see `!/terraform/environments/$triggering-directory/**.md` or something similar.

tom-webber:
> In its current form this feels overdriven to me; setting `paths` to ignore any markdown, rather than only the markdown in relevant paths is efficient from a number-of-characters perspective but gets away from the purpose of the included paths. I think I'd prefer to see `!/terraform/environments/$triggering-directory/**.md` or something similar.

An advantage of the existing syntax is that it reads as 'ignore markdown files', irrespective of the directory. Another is that it avoids needlessly bulking out the paths with replication of each path (e.g. for [3 paths here](https://github.com/ministryofjustice/modernisation-platform/pull/7374/commits/34b96d6674c0021abebba593fb458cf33b6a548a#diff-6edcfac4b0f8d6663b4617841bdead6183121454e477af0a964db11c8dac57b4R9-R11), and [9 paths here](https://github.com/ministryofjustice/modernisation-platform/pull/7374/commits/28edf7417c10bb7220b12324beb16d954bb2c252#diff-4f5830b5f03880cb136b9ee89592432b5a46b7a34bf243f5b71345af1c18d6fcR13-R21) <- _this last one was added after your comment, as I'd missed that the `modules/` folder contained readme's_). The specificity of the matching comes from the positive match assertions (directories), and the ignore is filtering from that subset.

dms1981 (approved):

Issue #7430, opened 2024-07-05
New account process incorrectly populates `codeowners` file


dms1981:
### Expected Behavior

From our guidance on [creating new accounts for customers](https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/creating-environments.html#blank-example) the `codeowners` key has a value of `[""]` which should result in the previous behaviour before this field was introduced being followed. EG. For every `github-team` in the application environments, these teams should be represented in the `CODEOWNERS` file.

### Actual Behavior

When we set `codeowners: [""]` as defined in our example, instead of ignoring this field the CODEOWNERS file is populated like so:
```
/terraform/environments/$example-application @ministryofjustice/ @ministryofjustice/modernisation-platform
```

### Steps to Reproduce the Problem

See this application json: https://github.com/ministryofjustice/modernisation-platform/blob/54474cf47a198b0029882e23cc55a69d37e99727/environments/genesys-call-centre-data.json
And this version of the `modernisation-platform-environments` CODEOWNERS file: https://github.com/ministryofjustice/modernisation-platform-environments/blob/aebfcdb8b292819d415988d3b1e239dd375ffd6d/.github/CODEOWNERS

The relevant script should be this: https://github.com/ministryofjustice/modernisation-platform/blob/36e3d9185ae2bfd39c17c94f337e40b7ccfae0cf/scripts/provision-member-directories.sh#L138

SteveLinden:
I've had a quick look at the code and it looks like this may be the issue. If there's no codeowners it looks like it defaults to the modernisation-platform ID.
    
if [ -n "$codeowners" ]; then
        echo "Adding $directory $codeowners@ministryofjustice/modernisation-platform  to codeowners"
        echo "$directory $codeowners@ministryofjustice/modernisation-platform" >> $codeowners_file
      # otherwise, use the github_slugs array
      else
        echo "Adding $directory $github_slugs@ministryofjustice/modernisation-platform to codeowners"
        echo "$directory $github_slugs@ministryofjustice/modernisation-platform" >> $codeowners_file
      fi

Going to seek clarification on the channel.

SteveLinden:
Based on my chatgpt information I can find the missing codeowners using this code
# Read the CODEOWNERS file, filter out comments and empty lines
codeowners_entries=$(grep -v -e '^#' -e '^$' ${codeowners_file})

# Find entries with missing or blank owners
[...]

PR #7468, opened 2024-07-10
Testsh

SteveLinden:

This is an additional piece of code that will be removed as the original is replaced. 
It allows a test of repos to see if I can correct the bug found under #7430 

If this correction identifies those then I will amend the code to correct the issue.


SteveLinden:
Removed the PR and I will try to amend the code to take into account missing codeowners

PR #7496, opened 2024-07-11
Bug/blank codeowners

SteveLinden:

To try to clear the bug on #7430 

This is a push to get the proposed changes approved/confirmation they will work to correct the issue.
The change is within the generate_codeowners.sh section.

Following discussions with team members (Aaron and Rich) it was decided that the solution proposed by Aaron would complete the required change with no issues. The code was amended to make this change.

### DIFF
--- a/scripts/provision-member-directories.sh
+++ b/scripts/provision-member-directories.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bashenvironments-networks/*.json
 
 # To test the script locally, uncomment the following two lines
 # core_repo_dir=.
@@ -145,6 +145,7 @@ echo "Writing codeowners file"
 * @ministryofjustice/modernisation-platform
 EOL
 
+
   for file in $environment_json_dir/*.json; do
     application_name=$(basename "$file" .json)
     directory=/terraform/environments/$application_name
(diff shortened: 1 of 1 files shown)


richgreen-moj:
When I tried this code locally Steve it made more changes to the CODEOWNERS file than expected. In my view the ideal scenario is that we update the script and re-run it so that the only change it would make is to the error on the [/terraform/environments/genesys-call-centre-data](https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/.github/CODEOWNERS#L34) line.

I've found a more succinct way of altering the if statement [here](https://github.com/ministryofjustice/modernisation-platform/blob/907928fbfea35b2ca884f25c0f9c7064f7164912/scripts/provision-member-directories.sh#L157)
`if [ -n "$codeowners" ] && [ "$codeowners" != "@ministryofjustice/ " ]; then`

I found that on this [line](https://github.com/ministryofjustice/modernisation-platform/blob/907928fbfea35b2ca884f25c0f9c7064f7164912/scripts/provision-member-directories.sh#L152) where `$codeowners` is defined, the JQ is appending `@ministryofjustice/` to any values in `.codeowners[]` which means that rather than `$codeowners` being a null value it is actually given a value of `"@ministryofjustice/"` and so the logic was adding that to the file.

I tested it locally and it worked as expected. I can show you that and help you to try running it locally.

scripts/provision-member-directories.sh
```diff
@@ -145,16 +145,17 @@ echo "Writing codeowners file"
...
     directory=/terraform/environments/$application_name
     account_type=$(jq -r '."account-type"' ${environment_json_dir}/${application_name}.json)
-    codeowners=$(jq -r 'try (.codeowners[] | "@ministryofjustice/" + .)' ${environment_json_dir}/${application_name}.json | sort | uniq | tr '\n' ' ')
+    codeowners=$(jq -r 'try (.codeowners[] | select(length > 0) | "@ministryofjustice/" + .)' ${environment_json_dir}/${application_name}.json | sort | uniq | tr '\n' ' ')
     github_slugs=$(jq -r '.environments[].access[].github_slug | "@ministryofjustice/" + .' ${environment_json_dir}/${application_name}.json | sort | uniq | tr '\n' ' ')
 
     if [ "$account_type" = "member" ]; then
       # if codeowners array has been defined in the json file, use that
-      if [ -n "$codeowners" ]; then
+      if [ -n "$codeowners" ] && ["$missing_codeowners" -ne 0 ]; then
```
davidkelliott, line 158: Is this needed?  I can't see where it is defined

SteveLinden: Bum. I thought I'd tidied that up too. It'll be up in a few minutes

SteveLinden: New one posted

richgreen-moj (approved):
Looks good and tested locally with success!

PR #7540, opened 2024-07-19
Adding lakeformation:get and put settings permissions to the data eng role

julialawrence:

The work this PR is a part of is being tracked here: 
https://github.com/ministryofjustice/analytical-platform/issues/4358

The problem:
In order to allow the process of sharing resources we have chosen, LakeFormation needs to be updated to `cross-account-sharing-version` 4. Ref: https://docs.aws.amazon.com/lake-formation/latest/dg/optimize-ram.html 
The ability to update the cross account sharing version isn't supported by terraform at present so this needs to be manually done.

This gives the data engineering role enough permissions to make the update in the console once the Lake Formation has been bootstrapped in terraform. 

## How has this been tested?

The permission have been retrieved from CloudTrail

Testing can only happen post-deployment

This should be a zero-impact change.

- [X] All checks have passed

### DIFF
--- a/terraform/environments/bootstrap/single-sign-on/policies.tf
+++ b/terraform/environments/bootstrap/single-sign-on/policies.tf
@@ -405,6 +405,8 @@ data "aws_iam_policy_document" "data_engineering_additional" {
       "glue:List*",
       "glue:BatchGetJobs",
       "glue:*Trigger",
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

Issue #7607, opened 2024-07-30
Protective Monitoring: Replace Firehose logs with s3


SimonPPledger:
### User Story

We currently provide a number of sets of logs to the XSIAM protective monitoring tool. However they are having issues with the info coming, via firehose, from the following:
VPC flow logs
Network firewall logs
Route 53 resolver logs

This ticket is to provide the same log information into  new S3 buckets.
To then liaise with the security development team to confirm that they can now ingest these logs correctly

### Value / Purpose

This will improve the security of the modernisation platform and enable the SOC to alert us of any issues

### Useful Contacts

leonardo.marini@justice.gov.uk

### Additional Information

This was originally done using firehose - the original ticket is here https://github.com/ministryofjustice/modernisation-platform/issues/6163

### Definition of Done

- [x] Discuss with Mike log files required
- [x] New module created - including new S3 bucket(s) created
- [x] Implement module to replace firehose
- [x] Security confirms info ingested by monitoring tool
- [x] Older firehose module removed
- [x] User docs have been updated
- [x] `source/runbooks/integration-with-protective-monitoring.html.md.erb` updated as required

dms1981:
From discussion with Leonardo I think we actually have almost everything in place. We already collate the logs required in S3 (VPC Flow logs, Route 53 Resolver logs, Firewall logs), so all that ought to be required is the SQS setup, notifications to SQS, and an appropriate user & policy for XSIAM

mikereiddigital:
Hi @dms1981. When I looked into this I could not see where the vpc flow logs output to s3 is set up so my impression was that we would need to add additional buckets & new flow logs to output to them. I was also considering the merit of keeping all of these buckets in core logging - so the same cross-account configuration that cloudtrail uses in its module - and having the accessible to the single IAM user account already set up. 
How this would translate into actionable terraform:
- A local module in core-logging for the s3 buckets, kms, iam resources etc & these resources created first. Assuming one bucket per cloud watch
- A new remote module for the s3 subscription filters, sqs queue etc where the cloudwatch log resources are created (core-vpc, core-logging, core-shared-services etc)
There may be questions around the one IAM account being aligned with security guidance.

Anyhow let me know what you think. Talk to you tomorrow.

dms1981:
I've done some reading and agree with the approach of exporting the logs into our `core-logging` account. I think we'll want a new bucket (with attendant bucket policy & SQS), and some expansion of the existing user to access that bucket.
[...]

PR #7753, opened 2024-08-22
Stream logs to separate buckets

dms1981:

#7607 

Because of how processors work in Cortex XSIAM, separate log categories require separate collection. Because we didn't get dynamic processing working in firehose, a short-term solution is to create independent buckets and Amazon Data Streams to separate out the logs.

This PR creates separate buckets & queues for the three log categories: `generic, vpc flow, route53`. This PR also calls the streaming module separately for each class of log required.

## How has this been tested?

Tested with local Terraform plan

Deploy through CI

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

Issue #7169, opened 2024-06-06
Use Observability platform for Modernisation Platform cost metrics


davidkelliott:
### User Story

As an MP Product manager, service owner and MP user
I want to view platform costs in total, broken down into sensible categories (TBC as part of issue)
So that I can use the data to drive cost savings

### Value / Purpose

Replace a manual process of cost metrics with automated dashboards which are easily accessible to mp PM and service owner to view their cost metrics.

### Additional Information

When doing this work it would be beneficial to do it at an AWS org level, so that we could also monitor total AWS spend, off platform AWS spend, platforms AWS spend, and application AWS spend in future issues.

### Proposal / Unknowns

Not sure where this should be done - maybe in the MOJ Master Account?

### Definition of Done

- [x] Total cost metrics for the modernisation platform on the observability platform
- [x] User docs have been updated
- [ ] Another team member has reviewed
- [x] Tests are green

dms1981:
We think this is probably best handled through integrating the MOJ Master account (organizational root) with observability platform, but we have a lot of unknowns here along the way to a defined outcome.

sukeshreddyg:
One approach is to:

1. Create a new report in the Cost Explorer saved reports.
2. Export that data to store in S3 using data exports by creating a new report.

However, there are a few questions regarding this implementation:

1. We don't have permissions to create an S3 bucket in the root account.
2. It's unclear whether we can create an S3 bucket in a cross account (i.e., in the MP account). If it is possible, then we can easily implement this solution.

ASTRobinson:
I have raised a question on the #ask-obervability-platform channel if this setup & configuration is feasible: https://aws.amazon.com/blogs/mt/visualize-and-gain-insights-into-your-aws-cost-and-usage-with-amazon-managed-grafana/

also I have imported two billing dashboards which can used with the current data sources (our core-accounts) to see costs across the different services in AWS. 

https://g-9d213fbc19.grafana-workspace.eu-west-2.amazonaws.com/d/AWSBilling/aws-billing-1?orgId=1
https://g-9d213fbc19.grafana-workspace.eu-west-2.amazonaws.com/d/AWSBillin/aws-billing-2?orgId=1

ASTRobinson:
[...]

PR #7763, opened 2024-08-23
Create core-billing account

ASTRobinson:
This PR creates a new environment name `core-billing` to help investigate better ways to display and dissect our platform costs as per issue #7169


mikereiddigital (approved):
lg2m

PR #7771, opened 2024-08-23
for_each the policies for buckets so they match each bucket

dms1981:

#7607

Bucket policies need to be scoped only to the bucket they're applied to

## How has this been tested?

Tested with local plan

Deploy through CI

- [x] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

PR #7792, opened 2024-08-29
Initial attempt to build pagerduty access for Sprinkler. Full details later

SteveLinden:
Various changes have been made to the code to ensure that this will create the correct pagerduty code to build it within Sprinkler.
Initially a chunk of code was placed in a routine but, after discussing this with a colleague, it was found the pagerduty work was in place on the environments file so it was not needed in here.

The other changes made can be seen in the latest version of the code. Sprinkler was added to the member-servces-integrations.tf code to ensure it was picked up and also included the line sprinkler                               = pagerduty_service_integration.integrations["sprinkler-development"].integration_key in aws.tf
This line may need to be amended to complete the testing.

The code was run manually against the pagerduty routine (pagerduty within the code) by passing in the correct secrets values. This was successful but the code shown in this PR is invalid. The reason for this will be investigated.

Temporarily setting this to "Ready to Review" to test signing


SteveLinden:
Closing pull request to bing in newer branch

Issue #7689, opened 2024-08-15
Signed commits and github-actions[bot]


dms1981:
### User Story

As a Modernisation Platform Engineer
I want to find a way for `github-actions[bot]` to sign its commits
So that we can continue to require signed commits in our repositories

### Value / Purpose

In a recent PR it was observed that the `github-actions[bot]` pushed a linting change into the branch. This commit was no signed, causing the PR to fail to meet our branch protection rules for signed commits.

Therefore we need to research and, if sensible, implement a way for the `github-actions[bot]` to sign its commits.

### Definition of Done

- [x] Options for signing commits by `github-actions[bot]` identified
- [x] Best option implemented across MP repositories
- [x] Solution tested successfully
- [x] Solution demonstrated to team

ASTRobinson:
Example bot PR: https://github.com/ministryofjustice/modernisation-platform/pull/7709

jacobwoffenden:
FWIW we use this https://github.com/planetscale/ghcommit-action 

https://github.com/ministryofjustice/analytical-platform/blob/fc64a650265ec6cf3ff6f8b62e72784fed09a936/.github/workflows/repository-dependabot-configuration-generator.yml#L39-L48

markgov:
https://github.com/Nautilus-Cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md
found this while looking up some information

mikereiddigital:
There are two workflows in scope:

- A scheduled workflow that runs daily on week days and which sweeps the existing code and generates a PR with any identified changes - ministryofjustice/modernisation-platform/blob/main/.github/workflows/format-code.yml

- A workflow that runs on any PR which calls the action ministryofjustice/github-actions/code-formatter which scans the code of the PR generates required changes and updates the PR with them. It is this workflow that the above description refers to.

mikereiddigital:
The suggested approach is to use planetscale/gscommit-action as set out in @jacobwoffenden's example above. This allows us to avoid maintaining gpg keys as secrets & sharing them between workflows. The first step will be to implement this in the code-formatter action and to test it via a commit ref from one of the formatter workflows in a lesser-used module.

jacobwoffenden:
I forgot to mention actually, when using our robot's token, its allows GitHub Actions to trigger workflows, when using GitHub Actions' token this is not allowed

mikereiddigital:
Completed testing.

[...]

PR #7794, opened 2024-08-29
Issue/7698

mikereiddigital:
Adds support for signed commits using the planetscale/ghcommit-actions github actions plus additional minor changes to support this addition.

#7689

This removes the existing git commands that commit changes and adds the use of planetscale/ghcommit-actions to create a signed commit that will be accepted by the repository with the mandatory signed commits branch protection. This approach avoids the needs to store and maintain GPG keys.

## How has this been tested?

This workflow has been tested (using some added minor formatting issues in a terraform file) as shown here - https://github.com/ministryofjustice/modernisation-platform/actions/runs/10613114434. All the changes work although because it is not running against the main branch the final (unchanged) part where the PR is created fails. 

This change amends a workflow that is run daily each weekday.

- [x] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -1,89 +1,88 @@
 ---
-  name: 'Format Code: ensure code formatting guidelines are met'
+name: 'Format Code: ensure code formatting guidelines are met'
+
+on:
+  workflow_dispatch: null
+  schedule:
+    - cron: 45 4 * * 1-5
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: '${{ github.ref }}-${{ github.workflow }}'
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: MegaLinter
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/format-code.yml
```diff
@@ -58,21 +58,22 @@ jobs:
...
           fi
-
-      - name: Push changes
-        if: env.changes == 'true'
-        run: |
-          git config --global push.autoSetupRemote true
-          git push
+  
+      - name: Generate signed commit for any changes that may exist
+        uses: planetscale/ghcommit-action@c7915d6c18d5ce4eb42b0eff3f10a29fe0766e4c #v1.44
```
ewastempel, line 70: We should be careful when using external actions from 3rd parties. Is it too difficult to achieve the same without using a GH action? Have you done some research on the action provider?

mikereiddigital: The above provider is one that was recommended by @jacobwoffenden in Analytics Platforms and is used by them for the same purpose. Without this, the other options would be to 
- Store & maintain our own GPG keys (which is an administrative burden for all of our repositories),
- Use the GraphQL API which is significantly more work involved. 
- The other option would be to temporarily disable signed commits for the workflow but I'm not in favour of that.

In short there is no simple or straight forward solution that I have been able to determine through this ticket.

ewastempel (approved):
LGTM, but it's introducing a new GH provider, so let's keep an eye on the action behaviour.

PR #7806, opened 2024-08-30
:bug: convert files to single line array rather than b64 encoded multiline output

georgepstaylor:
## A reference to the issue / Description of it

#7804 

## How does this PR fix the problem?

Replaces use of base64 to encode a string with a single-line array output

## How has this been tested?

Tested locally successfully.

## Deployment Plan / Instructions

Deploy through CI

## Checklist (check `x` in `[ ]` of list items)

- [x] I have performed a self-review of my own code
- [x] All checks have passed
- [x] I have made corresponding changes to the documentation
- [x] Plan and discussed how it should be deployed to PROD (If needed)

## Additional comments (if any)

https://mojdt.slack.com/archives/C013RM6MFFW/p1725016281240179?thread_ts=1725008454.283789&cid=C013RM6MFFW

PR #7809, opened 2024-08-30
:passport_control: add to codeowners for delius related projects

georgepstaylor:
Probation WebOps is no more! We are now part of a larger team for migrations - LAA Ops, PWO and DSO all as one. This new team reflects all member of the new team so that PR reviews can be more collaborative.

----
Reintroduces https://github.com/ministryofjustice/modernisation-platform/pull/7801
after fixing multiple env changes at once via
https://github.com/ministryofjustice/modernisation-platform/pull/7806/files

### DIFF
--- a/environments/delius-core.json
+++ b/environments/delius-core.json
@@ -1,6 +1,9 @@
 {
   "account-type": "member",
-  "codeowners": ["hmpps-migration"],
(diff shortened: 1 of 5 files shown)


richgreen-moj (approved):
Let's try again!

PR #7849, opened 2024-09-04
:coffin: Delete removed accounts

ASTRobinson:
This PR tidies up previously removed/deleted accounts from the `terraform/environments` directory, this has been checked against the .json files in the `environments` directory and also the `policies/environments/expected.rego` file.

Accounts Affected:
data-platform-compute
hmpps-intelligence-management
remote-supervision
tariff


richgreen-moj (approved):
Looks good. Status checks fail because of folder deletion but this has to be accepted in this scenario.

PR #7871, opened 2024-09-06
Adding Quicksight Admin policy for the new permission set

julialawrence:

https://github.com/ministryofjustice/analytical-platform/issues/5289

QuickSight can't be fully managed in code. Specifically, adding and removing identity center groups requires manual intervention as terraform forces a destroy when groups are amended.

This creates a policy that will eventually be utilised by a new permission set to carry out admin tasks in QS.

## How has this been tested?
The policy was copied from the AWS docs. Testing will take place after deployment

No special deployment instructions. There will be a follow-up two additional PRs. One that creates the permission set in `aws-root-account` and then amending of the json environment file and the rego tests that will assign the new permissions sets.

- [ ] All checks have passed

{Please write here}


markgov (approved):
LGTM

Issue #7800, opened 2024-08-29
collaborator s3 upload permissions


wullub:
### User Story

As a mod platform customer 
I need to be able to give collaborators ONLY permissions to upload artifacts to s3 - in particular the mod-platform-image-artefact-bucket20230203091453221500000001 bucket inside core-shared-services-production.
So that they can keep giving us software releases for oasys

https://mojdt.slack.com/archives/C01A7QK5VM1/p1724772031490929

### Value / Purpose

This is essential for the continuing collaboration with capita and their software

### Useful Contacts

DSO dev - William Gibbon, capita collaborator - Carl Last, oasys product manager - Howard Smith

### Definition of Done

- [x] s3-operator or s3-uploader role created
- [x] Carl Last (CarlDevCap) ,  Martin Hiorns (martinhiorns-capita) , Morgan Boniface (morgan-boniface) added as collaborators and given the role
- [ ] Confirm a collaborator can upload to mod platform artifact bucket

dms1981:
Creating SSO roles can be seen here: https://user-guide.modernisation-platform.service.justice.gov.uk/runbooks/adding-a-new-sso-user-role.html

However, it's worth having a think about what we're doing here. This might be more of a `modernisation-platform` scope where we add this in our bootstrap code and create a role across the MP member accounts? Have a look at how collaborators presently assume a role in a member account and work out from there what's required to create a role scoped to S3.

PR #7890, opened 2024-09-09
Add Role and Policy for S3 Upload

sukeshreddyg:

This PR introduces a new IAM role and policy that enables the user to list and upload objects to the `mod-platform-image-artefact-bucket20230203091453221500000001` in the `core-shared-services` account. The role provides necessary permissions for S3 operations specific to this bucket. #7800 

The newly created IAM role and policy address the requirement to allow specific users to upload and list objects in the test-bucket of the `core-shared-services` account. This ensures secure and limited access by restricting permissions to the necessary actions (s3:ListBucket, s3:PutObject) and to the specific bucket

## How has this been tested?

I tested the solution by assuming the newly created role (s3-upload-role) using test-user-1

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/bootstrap/single-sign-on/policies.tf
+++ b/terraform/environments/bootstrap/single-sign-on/policies.tf
@@ -1330,3 +1330,26 @@ data "aws_iam_policy_document" "fleet-manager-document" {
     resources = ["*"]
   }
 }
+
+resource "aws_iam_policy" "s3_upload_policy" {
+  provider = aws.workspace
+  name     = "s3_upload_policy"
+  path     = "/"
+  policy   = data.aws_iam_policy_document.s3_upload_policy_document.json
+}
+
+#tfsec:ignore:aws-iam-no-policy-wildcards
+data "aws_iam_policy_document" "s3_upload_policy_document" {
+  statement {
+    sid    = "AllowS3Upload"
+    effect = "Allow"
+    actions = [
+      "s3:PutObject",
+      "s3:ListBucket"
+    ]
+    resources = [
+      ""arn:aws:s3:::mod-platform-image-artefact-bucket20230203091453221500000001",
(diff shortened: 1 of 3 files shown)


terraform/environments/bootstrap/single-sign-on/policies.tf
```diff
@@ -1330,3 +1330,26 @@ data "aws_iam_policy_document" "fleet-manager-document" {
...
+data "aws_iam_policy_document" "s3_upload_policy_document" {
+  statement {
+    sid    = "AllowS3Upload"
+    effect = "Allow"
+    actions = [
+      "s3:PutObject",
+      "s3:ListBucket"
+    ]
+    resources = [
+      "arn:aws:s3:::mod-platform-image-artefact-bucket20230203091453221500000001",
```
dms1981, line 1351: I don't think this should be hard coded; I know it can be a pain to get information on bucket ARNs without resorting to hard coding, but hard coding ARNs is a very brittle approach. Do we even need to specify access to a single bucket?

ewastempel: I agree, not a good practice. My feeling is that the user wants to narrow it down to that specific bucket though (Sukesh is to confirm that), but, rather than referencing the bucket itself in the IAM role policy, perhaps we can narrow it down at the s3 bucket policy (by listing principals to specific roles?)

sukeshreddyg: Discussed with Ewa. Since there are many roles assuming access to the destination bucket, it is a bit challenging to edit the S3 permissions. Replaced hardcoded ARNs with dynamically fetched ARNs.

ewastempel (approved):
LGTM

Issue #5980, opened 2024-01-15
Address changes to PagerDuty integration


dms1981:
### User Story

As a Modernisation Platform Engineer
I need to look at how we use PagerDuty integration keys with SNS / other AWS code
So that we're not caught out by PagerDuty's deprecation of Rulesets and Event Rules at the end of 2024

### Value / Purpose

From [PagerDuty's documentation](https://support.pagerduty.com/docs/amazon-cloudwatch-integration-guide#integrate-with-rulesets), it's recommended that users switch from Rulesets / Event Rules to Event Orchestration.

We can see that the Terraform PagerDuty provider is currently warning us about the deprecated use of the `integration_key` argument which I believe is linked to this [PagerDuty provider change.](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/775).

We should take steps, if needed, to update our PagerDuty integration to make use of Event Orchestration before this argument is removed from the PagerDuty provider / any integration with PagerDuty itself is removed

### Definition of Done

- [x] Necessary changes assessed
- [ ] Changes implemented
- [ ] User docs have been updated
- [x] Another team member has reviewed
- [ ] Tests are green

SteveLinden:
Initial checks indicate that this is used in hundreds of location (pagerduty_service_integration)

dms1981:
Moved to blocked until Steve returns

SteveLinden:
Amended the pagerduty code in modernisation-platform to pick set up sprinkler for testing purposes. Unfortunately this has also created many other items in pagerduty so we may need to consider removing those that are not needed later.

The plan is to test changes against sprinkler to see if the proposed changes will work as required. This will, hopefully, stop other systems being amended/hit with the same changes.

SteveLinden:
It looks like the manual run, to add new items to pagerduty worked, or there were no errors send to pagerduty. 

I have also amended the [PR](https://github.com/ministryofjustice/modernisation-platform/pull/7792) so that it contains the correct location for Sprinkler. This was also run manually so it will point to [#modernisation-platform-low-priority-alarms](https://moj.enterprise.slack.com/archives/C02PFCG8M1R).

Testing will not commence on the changes required to correct the code to point at event orchestration.

SteveLinden:
A different approach was taken
Sprinkler was added in the current vise and this was completed through PR #7909 which was successful

Added orchestration using PR #7912 which, again, was successful.

[...]

PR #7904, opened 2024-09-11
Initial setup of Sprinkler

SteveLinden:

Related to #5980 

This is the initial attempt to set up PagerDuty on Sprinkler so that it's the same as current settings.

### DIFF
--- a/terraform/pagerduty/member-services-integrations.tf
+++ b/terraform/pagerduty/member-services-integrations.tf
@@ -2044,7 +2044,7 @@ resource "pagerduty_event_orchestration_service" "corporate-staff-rostering-prep
 }
 
 resource "pagerduty_service" "sprinkler-development" {
-  name                    = "sptinkler-development"
+  name                    = "sprinkler-development"
   description             = "sprinkler-development"
   auto_resolve_timeout    = 345600
   acknowledgement_timeout = "null"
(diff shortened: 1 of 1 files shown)


terraform/pagerduty/member-services-integrations.tf
```diff
@@ -2042,3 +2042,37 @@ resource "pagerduty_event_orchestration_service" "corporate-staff-rostering-prep
     actions {}
   }
 }
+
+resource "pagerduty_service" "sprinkler-development" {
+  name                    = "sptinkler-development"
```
dms1981, line 2047: Misspelling here?

terraform/pagerduty/member-services-integrations.tf
```diff
@@ -2042,3 +2042,37 @@ resource "pagerduty_event_orchestration_service" "corporate-staff-rostering-prep
...
+  description             = "sprinkler-development"
+  auto_resolve_timeout    = 345600
+  acknowledgement_timeout = "null"
+  escalation_policy       = pagerduty_escalation_policy.member_policy.id
+  alert_creation          = "create_alerts_and_incidents"
+}
+
+resource "pagerduty_service_integration" "sprinkler-integration" {
```
dms1981, line 2055: Do we need to include commented-out sections? Probably not.

SteveLinden: I'll remove the commented out bits.

SteveLinden:
Amendments made based on the comments. Branch re-deployed

ep-93 (approved):
Dont forget to take this out after testing if its no longer required

PR #7912, opened 2024-09-12
Initial attempt to create pagerduty orchestration for sprinkler

SteveLinden:
Again based on the call #5980 

This is an attempt to create orchestration on pager duty. It will, initially, just be on the creation of the code. 

I have had to include the team in there so there may be other versions in the future.

### DIFF
--- a/terraform/pagerduty/member-services-integrations.tf
+++ b/terraform/pagerduty/member-services-integrations.tf
@@ -2052,11 +2052,23 @@ resource "pagerduty_service" "sprinkler-development" {
   alert_creation          = "create_alerts_and_incidents"
 }
 
-resource "pagerduty_service_integration" "sprinkler-integration" {
-  name    = data.pagerduty_vendor.cloudwatch.name
-  service = pagerduty_service.sprinkler-development.id
-  vendor  = data.pagerduty_vendor.cloudwatch.id
+resource "pagerduty_team" "sprinkler-development" {
+  name = "sprinkler-development"
 }
+
+resource "pagerduty_event_orchestration" "monitor-sprinkler-integration" {
+  depends_on = [ pagerduty_team.sprinkler-development  ]
+  name = "My Monitoring Orchestration of Sprinkler"
+  description = "Testing the orchestration"
+  team = pagerduty_team.sprinkler-development.id
+}
+
+# Original definition.
+# resource "pagerduty_service_integration" "sprinkler-integration" {
+#   name    = data.pagerduty_vendor.cloudwatch.name
+#   service = pagerduty_service.sprinkler-development.id
+#   vendor  = data.pagerduty_vendor.cloudwatch.id
+# }
 
 resource "pagerduty_slack_connection" "sprinkler_connection" {
   source_id         = pagerduty_service.sprinkler-development.id


terraform/pagerduty/member-services-integrations.tf
```diff
@@ -2052,12 +2052,24 @@ resource "pagerduty_service" "sprinkler-development" {
...
-  name    = data.pagerduty_vendor.cloudwatch.name
-  service = pagerduty_service.sprinkler-development.id
-  vendor  = data.pagerduty_vendor.cloudwatch.id
+resource "pagerduty_team" "sprinkler-development" {
+  name = "sprinkler-development"
+}
+
+resource "pagerduty_event_orchestration" "monitor-sprinkler-integration" {
+  depends_on = [ pagerduty_team.sprinkler-development  ]
+  name = "My Monitoring Orchestration of Sprinkler"
```
davidkelliott, line 2061: I think this orchestration can be left here for the team to use to test future pagerduty changes.  Could you name it and the description something to reflect that, eg name="sprinkler-development-integration" description="Integrates sprinkler-development account with PagerDuty" thanks

terraform/pagerduty/member-services-integrations.tf
```diff
@@ -2052,12 +2052,24 @@ resource "pagerduty_service" "sprinkler-development" {
...
+  description = "Testing the orchestration"
+  team = pagerduty_team.sprinkler-development.id
 }
 
+# Original definition.
```
davidkelliott, line 2066: We don't need to commit comments, we can see the previous definition in the commit history

terraform/pagerduty/member-services-integrations.tf
```diff
@@ -2052,12 +2052,24 @@ resource "pagerduty_service" "sprinkler-development" {
   alert_creation          = "create_alerts_and_incidents"
 }
 
-resource "pagerduty_service_integration" "sprinkler-integration" {
-  name    = data.pagerduty_vendor.cloudwatch.name
-  service = pagerduty_service.sprinkler-development.id
-  vendor  = data.pagerduty_vendor.cloudwatch.id
+resource "pagerduty_team" "sprinkler-development" {
```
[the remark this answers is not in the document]
SteveLinden: I actually defined it as sprinkler team as per the terraform instructions. I can change it to modernisation-platform if you'd prefer and run another plan. Let me know

terraform/pagerduty/member-services-integrations.tf
SteveLinden: OK. I'll remove that.

terraform/pagerduty/member-services-integrations.tf
SteveLinden: I have amended the description to the one you suggested.

terraform/pagerduty/member-services-integrations.tf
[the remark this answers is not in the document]
davidkelliott: The terraform instructions define it as - `ID of the team that owns the Event Orchestration. If none is specified, only admins have access.`

So we want to make sure the right team (modernisation platform) owns the orchestration.

terraform/pagerduty/member-services-integrations.tf
davidkelliott: You have put the description in the name field and the description is still the same

terraform/pagerduty/member-services-integrations.tf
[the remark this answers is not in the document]
SteveLinden: I amended as requested

terraform/pagerduty/member-services-integrations.tf
SteveLinden: Corrected

davidkelliott (approved):
👍

Issue #6761, opened 2024-04-12
📝 Create runbook for Modernisation Platform account


dms1981:
### User Story

As a Modernisation Platform Engineer
I want to write a runbook
So that I can recreate resources for the Modernisation Platform account if the need arises

### Value / Purpose

[This backlog issue](https://github.com/ministryofjustice/modernisation-platform/issues/1908) identifies key platform components that may need to be recreated.

For the Modernisation Platform account we may need to recreate any or all of the following:
* The AWS account
* Resources inside the account used by customers
  * AWS Secretsmanager secrets & values
  * AWS resources used by Terraform

### Definition of Done

- [x] Runbook written
- [x] Another team member has reviewed

dms1981:
Runbook has been written; it's awaiting a review and a merge.

PR #7990, opened 2024-09-20
Add detail on reopening account to runbook

dms1981:

#6761 

Adds a little extra detail on how to recover a closed account. It's possible that we might need to set a different alias to the account, but if we're in scope to reuse the alias then we're out of scope to recover the account and I can't see why we'd not recover the account.

## How has this been tested?

Checked link correctly refers to AWS guidance

N/A

- [x] All checks have passed

{Please write here}


Kudzai-moj (approved):
LGTM

PR #8044, opened 2024-09-24
Replaced modernisation_platform.id with the correct 

SteveLinden:
modernisation_platform_members.id name

After the previous successful run Dave noticed I had used modernisation-platform.id rather than modernisation-platform-members.id. This PR is to correct that error.

The previous PR was #8040


davidkelliott:
There are some unrelated changes in the plan?

SteveLinden:
> There are some unrelated changes in the plan?

I have added the missing items back to the PR. Hopefully this will fix the issues

dms1981 (approved):
LGTM

PR #8065, opened 2024-09-25
Associate R53 Resolver QLC RAM share across MP OU

dms1981:

#7607

Adds a terraform resource to share the new RAM Shares with accounts in the MP OU. They will still need to accept the share to make use of it.

This is somewhat different to how we carry out shares in the `modules/ram-principal-association` module which share resources on a per-account basis.

## How has this been tested?

Tested with local plan

Deploy through CI

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #8080, opened 2024-09-26
Create data calls and local values to associate production VPCs with …

dms1981:

#7607

This PR does three things:

1. Create data objects for the RAM-shared `aws_route53_resolver_query_log_config` resources created and shared from `core-logging`
2. Create a set of local values which are combined to feed the `aws_route53_resolver_query_log_config_association` so that each VPC is associated with each resolver log query
3. Create the config associations in production only

## How has this been tested?

Tested with local terraform plan

Deploy through CI

- [x] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

PR #8111, opened 2024-09-30
Assumable role for Cortex XSIAM

dms1981:

#7607 

Creates an assumable role that can be used by Palo Alto Cortex XSIAM in line with their documented guidelines. I deliberately haven't attached the `Security Audit` managed policy as this isn't currently being used by the Cortex user, and provides access to the account which we may not prefer to offer.

## How has this been tested?

Tested with CI pipeline. I expect that this will fail because - at present - there is no value for the account ID SSM parameter. This will be added manually once the parameter has been created

Deploy through CI. Potentially re-run after populating SSM parameter.

- [x] All checks have passed

{Please write here}


richgreen-moj (approved):
Looks good , lets give it a go!

PR #8120, opened 2024-09-30
Switch to insecure_value for nonsensitive values

dms1981:

#7607

SSM Parameter Store was chosen specifically to hold parameters that don't need special handling. This PR swaps from `value` to `insecure_value` to address this. The alternative - wrapping outputs as `nonsecure()` would achieve the same thing, so to make the correct handling of these values clear I've chosen to switch the attributes used.

## How has this been tested?

Tested with local plan and through CI pipeline

Deploy through CI

- [x] All checks have passed

{Please write here}


richgreen-moj (approved):
lgtm

PR #8122, opened 2024-09-30
More fixes for ssm params

dms1981:

#7607

SSM parameters can't be overwritten without `overwrite = true`. Once an SSM parameter has been created, the value can't be changed to `insecure_value` without this flag. I worked around it originally by deleting the parameters manually to recreate them, but this will allow us to continue to work programatically.

## How has this been tested?

Tested with local apply targeted at parameters.

Deploy through CI

- [x] All checks have passed

{Please write here}


richgreen-moj (approved):
LGTM

PR #8133, opened 2024-10-01
add check requirements function

richgreen-moj:

Having spoken to @Khatraf who recently ran through the account deletion process - she didn't have jq installed which is required in one of the scripts so it failed part-way through.

I've added a small function to check the requirements for the script. Esssentially it checks that `terraform` and `jq` are installed and cancels the run if they are not.

## How has this been tested?

Tested locally. When I put in dummy package names I got the following result:

```
 ./delete-accounts.sh
Error: terrafom is not installed.
Error: j is not installed.
Please install the missing programs and try again.
```
But it runs through successfully when set to the proper names and the requirements are installed.

- [X] All checks have passed

{Please write here}

### DIFF
--- a/scripts/account-deletion/delete-accounts.sh
+++ b/scripts/account-deletion/delete-accounts.sh
@@ -1,5 +1,26 @@
 #!/bin/bash
 
+# Check required programs are installed
+check_requirements() {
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #8137, opened 2024-10-02
Improve use of `firewall_policy` module

dms1981:

When the firewall policy terraform is executed an empty `rule_variables` block is created regardless of if `ip_sets` or `port_sets` is populated.

Wraps the `rule_variables {}` in a dynamic block, and creates it only when `ip_sets` or `port_sets` is supplied.

## How has this been tested?

Tested with local terraform plan

Non-functional deployment through CI; expect no changes.

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #8152, opened 2024-10-03
Amended name on pagerduty_event_orchestration cdpt_ifs_cloudwatch to …

SteveLinden:
…another name

There was an error in a previous release that was unhappy with the name being used. This will correct that error and ensure the PR passes smoothly.

See #5980


markgov (approved):
LGTM

PR #8154, opened 2024-10-03
Add permissions for users to view console home widgets

davidkelliott:
The AWS console home has widgets which display some data such as health and costs.

This gives permissions for users assuming the mp account read only role to view that data.

This will reduce the number of unauthorised API call alerts that we recieve every time a user assumes that role.

The read only policy is also being attached to the collaborator user group, this means that collaborators also have permissions to view these widgets, plus they no longer need to switch to a read only role in the same account.


markgov (approved):
LGMT

PR #8246, opened 2024-10-11
Adds AVD-GIT-0004 to trivyignore.

mikereiddigital:
Adds AVD-GIT-0004 which adds an ignore for signed commits not being set for modernisation-platform-environments.

The repository modernisation-platform-environments does not have "require signed commits" enabled for the main branch. This adds the required ignore as it is failing the test as evidenced here - https://github.com/ministryofjustice/modernisation-platform/actions/runs/11287877020

See above.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/.trivyignore.yaml
+++ b/.trivyignore.yaml
@@ -39,6 +39,7 @@ vulnerabilities:
 
 misconfigurations:
 - id: AVD-GIT-0001
+- id: AVD-GIT-0004
 - id: AVD-AWS-0031
 - id: AVD-AWS-0039
 - id: AVD-AWS-0057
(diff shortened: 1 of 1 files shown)


.trivyignore.yaml
```diff
@@ -39,6 +39,7 @@ vulnerabilities:
 
 misconfigurations:
 - id: AVD-GIT-0001
+- id: AVD-GIT-0004
```
ewastempel, line 42: are you able to provide a comment here  (wtih # ) why we are suppressing it?

mikereiddigital: Yes I will add that, not least because we will want to remove this rule when MPE has require_signed_commits enabled in the future.

ewastempel (approved):
LGTM

PR #8249, opened 2024-10-11
Add custom key to s3 bucket

ASTRobinson:
As part of ticket https://github.com/ministryofjustice/modernisation-platform/issues/7169 this PR adds the KMS key to the cur-reports-buckets for encryption.


mikereiddigital (approved):
lg2m

PR #8262, opened 2024-10-14
Trivy ignore AVD-AWS-0102 for subnet NACLs

richgreen-moj:

Daily run of the Static Code Analysis job is failing...
https://github.com/ministryofjustice/modernisation-platform/actions/runs/11340947819/job/31538282534#step:4:3094

Added an ignore on a few resources in the `vpc-nacls` module for `avd-aws-0102` as the rules are open to all ports by design.

As suggested, I've also removed the tfsec ignores as that is no longer in use.

## How has this been tested?

Status checks are passing.

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modules/vpc-nacls/main.tf
+++ b/terraform/modules/vpc-nacls/main.tf
@@ -35,7 +35,7 @@ resource "aws_network_acl" "protected" {
   )
 }
 
-#tfsec:ignore:aws-vpc-no-excessive-port-access tfsec:ignore:aws-ec2-no-public-ingress-acl
+#trivy:ignore:AVD-AWS-0102
 resource "aws_network_acl_rule" "data_subnet_static_rules" {
   #checkov:skip=CKV_AWS_352:Verified - these rules are reasonable
   #checkov:skip=CKV_AWS_231:Allow ingress from 0.0.0.0:0 to port 3389 required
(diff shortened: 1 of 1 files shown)


dms1981 (changes requested):
Can you try removing the `ignore` from `core-vpc/vpc.tf`, and moving the `ignore` so that it's before the resource declaration down in the module please?
https://aquasecurity.github.io/trivy/v0.41/docs/configuration/filtering/#by-inline-comments
Since we've migrated from tfsec we can probably remove those tfsec ignores too 😄

dms1981 (dismissed):
LGMT - thanks!

terraform/modules/vpc-nacls/main.tf
```diff
@@ -35,7 +35,7 @@ resource "aws_network_acl" "protected" {
   )
 }
 
-#tfsec:ignore:aws-vpc-no-excessive-port-access tfsec:ignore:aws-ec2-no-public-ingress-acl
+#trivy:ignore:AVD-AWS-0102
```
ewastempel, line 38: it would be useful if we added the information why the warning is being suppressed

richgreen-moj: Comment added to trivy ignores

ewastempel (approved):
LGTM, thanks for adding the comments

PR #8301, opened 2024-10-21
Correct Lambda trigger bucket

ASTRobinson:
This PR corrects the bucket used for the lambda trigger and also adds a data source tag to the Athena bucket for Grafana.


mikereiddigital (approved):
lg2m

PR #8304, opened 2024-10-22
Use Bucket Name not ARN

ASTRobinson:
following on from PR https://github.com/ministryofjustice/modernisation-platform/pull/8301 which failed to deploy this PR updates the resources from the ARN to bucket name


mikereiddigital (approved):
lg2m.

Issue #7943, opened 2024-09-17
Replace Current KMS Keys with Multi Region Keys


Kudzai-moj:
### User Story

Multi region keys and [replica's](https://github.com/orgs/ministryofjustice/projects/17/views/4?pane=issue&itemId=70910081) have been created for the KMS keys below,

- s3-state-bucket-multi-region
- pagerduty-secret-multi-region
- dynamodb-state-lock-multi-region
- environment-management-multi-region

### Value / Purpose

We need to analyse impact of changing the current keys to the new multi region keys. Once we are sure on impact we can then change the keys. Its important to use multi region keys as part of DR because, If one region experiences a failure or outage, you can use the same KMS key in a different region without any manual intervention

### Useful Contacts

Kudzai and Edd

### Additional Information

_No response

### Definition of Done

- [x] Fully understand impact of changing the keys - will be good to speak to Dave Elliot to understand impact
- [x] KMS keys replaced by multi region keys
- [x] Test that old resources encrypted with old keys can still be accessed

dms1981:
Having looked at this issue some of it doesn't look too bad. 

* Replacing the KMS key for the dynamodb table used with a multi-region key looks seamless (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/EncryptionAtRest.html).
* Replacing the KMS key for the secrets also doesn't look too bad (https://docs.aws.amazon.com/secretsmanager/latest/userguide/security-encryption.html).

dms1981:
With the S3 bucket things are a little less clear. If we replace the key we then need to re-encrypt the content of each bucket that uses the same key. Further to this we can see the following in the S3 docs:

> You can use multi-Region AWS KMS keys in Amazon S3. However, Amazon S3 currently treats multi-Region keys as though they were single-Region keys, and does not use the multi-Region features of the key. For more information, see [Using multi-Region keys](https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-overview.html) in the AWS Key Management Service Developer Guide.

From reading through the guidance I don't think this is going to be a factor for us; the major goal for us it to ensure that Terraform can be run seamlessly.

dms1981:
https://github.com/ministryofjustice/modernisation-platform/pull/8309 << Pagerduty change PR.

dms1981:
GitHub secret, PagerDuty secret, DynamoDB table all updated to use new key without issues.

PR #8310, opened 2024-10-22
Update references for `environment_management` KMS to multi-region key

dms1981:

#7943 

Updates terraform references to the `environment_management` KMS key from the single-region key to the multi-region key.

## How has this been tested?

It has not. Assuming CI tests pass, replacing one KMS key with another should be seamless.

Deploy through CI.

Once deployed and tested, old key can be removed and scheduled for deletion.

- [x] All checks have passed

{Please write here}


ewastempel (commented):
LGTM

PR #8316, opened 2024-10-23
♻️  Refactor / tidy up 

ASTRobinson:
This PR renames the lambda crawler script to `moj_cur_crawler_lambda` to fix a naming issue when being called. 
also...
- switch workgroup resource to snake case from kebab
- add custom KMS key to workgroup
- remove hardcoded S3 bucket names


dms1981 (approved):
LGTM

PR #8319, opened 2024-10-23
Replace state bucket and cur bucket KMS keys with multi-region key

dms1981:

#7943

Replaces references to single-region KMS key with multi-region equivalent

## How has this been tested?

Tested through CI checks.

Deploy through CI. Once deployed, existing objects will need to be re-encrypted using new key prior to scheduling old key for deletion.

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #8321, opened 2024-10-23
Revert workgroup KMS

ASTRobinson:
As PR reverts part of https://github.com/ministryofjustice/modernisation-platform/pull/8316/ where the KMS key was added to the Athena workgroup as I am now seeing errors when trying to run queries and want to rule this out as the cause.


richgreen-moj (approved):
lgtm

PR #8343, opened 2024-10-25
Remove non-multi-region KMS keys

dms1981:

#7943

Now that the multi-region KMS keys are in active use this PR removes the old keys and schedules them for deletion

## How has this been tested?

Tested through CI pipeline checks

Deploy through CI

- [x] All checks have passed

{Please write here}


dms1981:
NB. If merging this PR in _does_ cause unexpected outcomes we can revoke the deletion following the guidance here: https://docs.aws.amazon.com/kms/latest/developerguide/deleting-keys-cancelling-key-deletion.html

Once the keys have had their deletion markers removed we can revert this PR and re-import the keys.

ASTRobinson (approved):

PR #8394, opened 2024-10-31
Resolve resource deprecation warning for read-only role

dms1981:

When running Terraform for one of the `core-vpc` workspaces the following warning is output:
```
│ Warning: Argument is deprecated
│ 
│   with aws_iam_role.member_delegation_read_only,
│   on vpc.tf line 320, in resource "aws_iam_role" "member_delegation_read_only":
│  320:   managed_policy_arns = ["arn:aws:iam::aws:policy/ReadOnlyAccess"]
│ 
│ The managed_policy_arns argument is deprecated. Use the aws_iam_role_policy_attachments_exclusive resource instead.
│ 
│ (and one more similar warning elsewhere)
```

This PR implements the `aws_iam_role_policy_attachments_exclusive` resource. Strictly speaking I don't think this exact resource is required, but it will enforce that only the read-only policy will be attached to this role, and any other policies will be detached should they be picked up on a terraform run.

## How has this been tested?

Tested in `core-vpc-sandbox` successfully

Deploy through CI

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-vpc/vpc.tf
+++ b/terraform/environments/core-vpc/vpc.tf
@@ -316,8 +316,7 @@ resource "aws_iam_role_policy" "member-delegation" {
 
 # Read only role for developer sso plans and for viewing via the console
 resource "aws_iam_role" "member_delegation_read_only" {
-  name                = "member-delegation-read-only"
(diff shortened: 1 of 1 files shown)


ewastempel (approved):
LGTM

Issue #7907, opened 2024-09-11
Restrict iam:PassRole for Github OIDC role


Khatraf:
### User Story

As a Modernisation Platform engineer
I want to restrict the use of the iam:PassRole action for the OIDC role
So that the permissions are not overly permissive, and only specific roles can be passed to services

### Value / Purpose

The value of implementing this is to improve security by ensuring that iam:PassRole is only used with specific roles, reducing the risk of privilege escalation or misuse.  By adding the iam:PassedToService condition key or specifying resource ARNs, it aligns with best practices for least privilege access control.

### Additional Information

- https://docs.aws.amazon.com/IAM/latest/UserGuide/access-analyzer-reference-policy-checks.html#access-analyzer-reference-policy-checks-security-warning-pass-role-with-star-in-resource
- https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_passrole.html

### Definition of Done

- [ ] Update the IAM policy and ensure implementation follow security best practices
- [ ] Test and validate the policy changes to ensure they function as intended without impacting existing services.
- [ ] Another team member has reviewed
- [ ] Tests are green

dms1981:
Having investigated this I need to reach out to some of the role users. One strand of users appear to make use of the role for ECR related tasks in build pipelines. A second strands make use of the role for environment automation tasks such as ansible playbooks.

For the first case, `iam:passrole` isn't a factor as the role is used to directly interact with ECR. For the second case I might be able to scope passrole with an `iam:PassedToService` condition so I've reached out to see what services might be in scope.

Kudzai-moj:
I've reviewed this and am happy it's complete.

dms1981:
https://github.com/ministryofjustice/modernisation-platform/pull/8443 << This PR also relates. Found out after we'd added the `PassedToService` condition that ECS Tasks users were negatively impacted, but this PR fixed the problem.

PR #8415, opened 2024-11-05
Add condition to `modernisation-platform-oidc-cicd` `iam:PassRole` permission

dms1981:

#7907 

This PR prevents the `modernisation-platform-oidc-cicd` role from being passed to any AWS service or resource unless it is `ssm.amazonaws.com`.

This will prevent the accidental - or malicious - passing of this role to an unexpected resource such as an EC2 instance or lambda function.

## How has this been tested?

Tested via ClickOps with `nomis-test` and @keirwilliams who confirmed the condition does not negatively affect their automation tasks.

Deploy through CI

- [x] All checks have passed

{Please write here}


ep-93 (approved):
The testing is good, and deploying it to a few environments already is a good idea. Any other teams you can deploy it to the better, before the big roll out.

PR #8474, opened 2024-11-13
🙅 Exclude Analytical Platform Compute accounts from instance scheduler

jacobwoffenden:

We have long running overnight jobs in our test EKS cluster that are being shutdown

https://github.com/ministryofjustice/data-platform-support/issues/946

Excludes APC non-prod accounts

## How has this been tested?

It hasn't

I don't think so

- [ ] All checks have passed

Signed-off-by: Jacob Woffenden <jacob.woffenden@justice.gov.uk>


dms1981 (approved):
Approved in discussion with @jacobwoffenden

Issue #8272, opened 2024-10-16
Can we allow members to add their own Terraform providers?


dms1981:
### User Story

As a Modernisation Platform Customer
I need to add providers to my `terraform { required_providers { ... } }`
So that I can supply providers that my code requires

### Value / Purpose

Modernisation Platform customers have an ongoing need to manage their own terraform code, including specifying their own providers and versions. At present the only location that they can specify the relevant `providers` is in the `platform_versions.tf` file which is created from a standard template.

There is a risk that this file can be overwritten by a script in the Modernisation Platform repository - [replace-platform-files-with-templates.sh](https://github.com/ministryofjustice/modernisation-platform/blob/main/scripts/replace-platform-files-with-templates.sh).

We should find a way to allow customers to specify their own providers as they require.

### Definition of Done

- [x] Assess possible options
  - [x] Should we refactor `platform_versions.tf` to `versions.tf` and update our templates?
  - [x] Should we just remove `platform_versions.tf` from the overwrite script?
  - [x] Would we need to update the CODEOWNERS file to allow customer management without our oversight? 
- [x] Chosen option implemented

dms1981:
I think it's worth pointing out here that customers can define providers separately in modules, so keeping this gated doesn't give us the control that we might think it does. Personally I think this should be open to customers to adjust as their infrastructure requires.

dms1981:
https://github.com/ministryofjustice/modernisation-platform-environments/pull/8674

mikereiddigital:
I've reviewed this issue & moving to done.

PR #8488, opened 2024-11-14
Amend `provision-member-directories.sh`

dms1981:

#8272 

Removes `providers.tf` from the autogenerated `CODEOWNERS` file we populate [modernisation-platform-environments](https://github.com/ministryofjustice/modernisation-platform-environments) with.

This is a precursory step to renaming the `platform_providers.tf` file in our templates directory.

## How has this been tested?

Tested script locally and confirmed behaviour

Deploy through CI, review PR in modernisation-platform-environments

- [x] All checks have passed

{Please write here}

### DIFF
--- a/scripts/provision-member-directories.sh
+++ b/scripts/provision-member-directories.sh
@@ -167,7 +167,6 @@ EOL
   done
 
   cat >> $codeowners_file << EOL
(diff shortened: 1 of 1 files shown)


ep-93 (approved):
Approved, but if we make any new accounts before the change you are making is done gotta manually pop it in.

Issue #8078, opened 2024-09-26
🔐 Restrict Terraform Apply from Running on Branches/Pull Requests


connormaglynn:
### User Story

When creating a branch/pull request in a Modernisation Platform Repository. 
I expect to be unable to perform destructive operations through CI/CD, such as Terraform Apply. 
So that all destructive changes are required to go through reviews before being able to run.

### Value / Purpose

At the moment, users with write access to Modernisation Platform repositories can bypass the normal peer review process before running destructive actions due to open OIDC permissions given to the role which is assumed by the pipelines.

The value of this work is to ensure that peer reviews are enforced before any destructive changes are made to the platform.

### Additional Information

- Need to identify where we are creating OIDC roles and ensure that permissions are separated out into multiple roles: one for destructive actions (Terraform Apply) and one for non-destructive (Terraform Plan)

### Definition of Done

- [x] I am unable to run Terraform Apply on branches and pull requests by modifying the GitHub Workflow files
- [x] CI/CD processes continue to work as usual

kcbotsh:
This has been done in the root account already

sukeshreddyg:
Most of the workflows have been updated, but a few are still facing issues. I have raised follow-up stories for them.
https://github.com/ministryofjustice/modernisation-platform/issues/8613
https://github.com/ministryofjustice/modernisation-platform/issues/8590

PR #8530, opened 2024-11-19
Fix Missing Assume Roles in GitHub Actions OIDC Plan Role

sukeshreddyg:
This PR updates the GitHub Actions OIDC role to include the missing assume roles that were causing failures in the pipeline. Certain workflows were failing due to insufficient permissions for assuming the necessary roles, leading to execution errors. By adding the required assume roles to the OIDC role configuration, these failures will be resolved. #8078


Khatraf (approved):
lgtm

PR #8532, opened 2024-11-19
Edit SSM Inventory Sync bucket policy for matching source org ID

richgreen-moj:

Having created a central bucket to store inventory data in the Modernisation Platform account I subsequently tried to create a resource data sync from Sprinkler but encountered an access denied error: https://github.com/ministryofjustice/modernisation-platform/actions/runs/11895397311/job/33146521856?pr=8517#step:7:236

After troubleshooting with AWS support it seemed that my policy where I had deviated slightly from AWS's [suggested one](https://docs.aws.amazon.com/systems-manager/latest/userguide/inventory-create-resource-data-sync.html#systems-manager-inventory-resource-data-sync-AWS-Organizations) was causing the issue.

I can only restrict this down to the source org ID of the MoJ organization rather than a `ForAnyValue:StringLike` in `aws:PrincipalOrgPaths`.

Though this means MoJ accounts outside of MP could sync their inventory data into the bucket I would say the risk of malicious activity is acceptably low. I won't be creating a resource data sync for those accounts anyhow as they are out of scope.

## How has this been tested?

Manually tested in CLI/console.

{Please write here}

- [X] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modernisation-platform-account/s3.tf
+++ b/terraform/modernisation-platform-account/s3.tf
@@ -420,9 +420,9 @@ data "aws_iam_policy_document" "ssm_inventory_sync_bucket_policy" {
       values   = ["bucket-owner-full-control"]
     }
     condition {
-      test     = "ForAnyValue:StringLike"
-      variable = "aws:PrincipalOrgPaths"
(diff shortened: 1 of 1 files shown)


Khatraf (approved):
lgtm

PR #8604, opened 2024-11-26
Remove Conditional logic to switch OIDC role

sukeshreddyg:
This PR will remove the conditional logic for the OIDC role, as it is causing issues during Terratest runs due to the use of github-action-plan, which has only read permissions. By removing this condition, it will revert to using the previous GitHub Actions role, restoring the earlier functionality.


richgreen-moj (approved):
LGTM

Issue #8545, opened 2024-11-20
Core-Logging - lambda athena_table_update failing with Access to KMS is not allowed


mikereiddigital:
### Expected Behavior

The above lambda should run nightly without error.

### Actual Behavior

The lambda attempts to access Secrets Manager in the Modernisation Platform account and on doing so an AccessDenied error is thrown for a KMS key.

The following cloudtrail IDs in the Modernisation Platform account show the latest errors for the first run and the following two retries.

4e6d85d9-2765-3130-b86e-69c74cf19da8

f66c72e4-a0dd-3caa-a002-b69ee23582b7

2020a66f-05dd-3de8-9f5f-fc653c324ec2

### Steps to Reproduce the Problem

The lamda runs nightly in core-logging. The source for the lambda is https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/core-logging/athena.tf

### Account

core-logging

Khatraf:
The nightly lambda job is now running without issue after switching to the multi-region kms key.

mikereiddigital:
Have reviewed & moving to done.

PR #8618, opened 2024-11-28
Use multi-region KMS key for environment management

Khatraf:

#8545 

The AccessDenied error encountered when the athena_table_update Lambda function attempts to retrieve the environment_management secret from AWS Secrets Manager was caused by the Lambda function using an incorrect single-region KMS key for decryption, instead of the multi-region key associated with the secret.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


markgov (approved):
LGTM

Issue #7825, opened 2024-09-02
Monitor root user activity in MP accounts


richgreen-moj:
### User Story

As a MP engineer
I want to be alerted to when the AWS account root user is being used
So that I know that it is only used for genuine purposes (and there are no security breaches)

### Value / Purpose

This is a follow-on from [#7437](https://github.com/ministryofjustice/modernisation-platform/issues/7437) which looked at alerting for admin role usage and more generally the NCSC recommendations on mitigating the risk of an attacker gaining unauthorised access to code and production environments.

It may need to be looked at after this more general ticket that investigates how we will be looking after the root account/codebase... https://github.com/ministryofjustice/modernisation-platform/issues/7824 

### Definition of Done

- [x] Identify best solution for alerting and where to alert
- [x] Deploy solution
- [x] Update documentation
- [ ] Reviewed by another team member

SimonPPledger:
#7824 is now complete

richgreen-moj:
PR https://github.com/ministryofjustice/modernisation-platform/pull/8654 has enabled monitoring/alerting for root account usage across all MP accounts. The PR explains in detail what has been enabled including links to examples of root account usage alerts being triggered.

I don't think there is any need to update any particular documentation for this change. 

Over time we should review the amount of traffic being sent to the low priority alarms channel from member accounts and look to act on any issues or see if we need to tweak the alarm configuration etc.

Kudzai-moj:
everything looks good to me. Moving to done

PR #8654, opened 2024-12-02
Subscribe member account alarms into pagerduty/slack

richgreen-moj:

As part of https://github.com/ministryofjustice/modernisation-platform/issues/7825 I've been looking at the ability to monitor/alert for root account activity.

As part of this I've noticed that the baselines code has already deployed a set of alarms for certain scenarios to all MP accounts, including root usage activity: https://github.com/ministryofjustice/modernisation-platform-terraform-baselines/blob/main/modules/securityhub-alarms/main.tf#L133

These are labelled "security-hub alarms" and have been deployed as they are direct recommendations from the security hub controls e.g. AWS/CIS best practices.

Currently we only receive alerts for these issues where we have subscribed our "core" accounts (i.e. `core-shared-services`, `core-logging`, `core-network-services`, `core-security`, `core-vpc` and `modernisation-platform`) which then feed into our [core-alerts](https://moj-digital-tools.pagerduty.com/service-directory/PYDTWVM) pagerduty service which in turn feed the [low-priority](https://moj.enterprise.slack.com/archives/C02PFCG8M1R) or [high-priority](https://app.slack.com/client/E01066BNN9G/C03CY6451QT) Slack channels.

An example of me triggering the root account usage alarm for the MP account can be seen [here](https://mojdt.slack.com/archives/C02PFCG8M1R/p1733144952405839).

## NOTE:

This also fixes an issue described in [this thread](https://mojdt.slack.com/archives/C013RM6MFFW/p1733159697853769)

In the words of Sukesh:

_"The issue occurred because the `aws_caller_identity` data resource was unable to retrieve the `account_id` for the Modernisation Platform (MP). This happened because the provider was using the Sprinkler GitHub OIDC role now. Recently, Sprinkler introduced a separate GitHub OIDC role, and the workflow was updated to use that role from the Sprinkler account. Previously, the workflow was associated with the MP provider and assumed the role from the MP account. These changes caused the errors you were seeing.
I have now updated the reference to the MP account reference in single-sign-on and member-bootstrap in your PR, which should resolve the issue."_  

This PR subscribes the existing cloudwatch alarms/SNS topics deployed through the baseline to pagerduty in the same way as our core accounts so that we will start to receive alerts in our channels for member accounts also.

This will mean we can monitor any root account usage across the platform as well as for any of the other alarms defined in [ministryofjustice/modernisation-platform-terraform-baselines/modules/securityhub-alarms/](https://github.com/ministryofjustice/modernisation-platform-terraform-baselines/blob/main/modules/securityhub-alarms/)

This would include such alarms as `admin-role-usage`, where we have been tracking the MP teams usage of the `AdministratorAccess` role,  and a whole lot more.

I did consider creating a separate channel or service to split these although it seems like too much duplication initially - plus although these alarms are deployed to  member accounts they are for our (MP teams) benefit. and so should have equal importance to those already subscribed in the core account set.

Potential downsides to this could be that we see certain alarms triggering too frequently such as `unauthorised-api-calls` which we have recently been trying to filter out by addressing any shortcomings with our own code-base. If receiving these for member accounts it might not necessarily be within our gift to remedy these although we could notify them and enquire as to why it's occurring.

If the alerting seems to noisy we could pivot and create a new dedicated sns topic for some of the least frequently occuring events i.e. root account usage, admin role usage etc.

## How has this been tested?

Tested in Sprinkler 
See example alerts:
https://moj-digital-tools.pagerduty.com/incidents/Q1OY2KAUS4MELK?utm_campaign=channel&utm_source=slack
https://moj-digital-tools.pagerduty.com/incidents/Q2NZBEYMP5NPB8?utm_campaign=channel&utm_source=slack

{Please write here}

- [X] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/collaborators.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/collaborators.tf
@@ -13,7 +13,7 @@ module "collaborator_readonly_role" {
 
   # Allow created users to assume these roles
   trusted_role_arns = [
-    local.modernisation_platform_account.id
+    data.aws_ssm_parameter.modernisation_platform_account_id.value
   ]
 }
 
@@ -40,7 +40,7 @@ module "collaborator_security_audit_role" {
 
   # Allow created users to assume these roles
   trusted_role_arns = [
-    local.modernisation_platform_account.id
+    data.aws_ssm_parameter.modernisation_platform_account_id.value
   ]
 }
 
@@ -52,7 +52,7 @@ module "collaborator_developer_role" {
   source = "github.com/terraform-aws-modules/terraform-aws-iam//modules/iam-assumable-role?ref=de95e21a3bc51cd3a44b3b95a4c2f61000649ebb" # v5.39.1
 
   trusted_role_arns = [
-    local.modernisation_platform_account.id
+    data.aws_ssm_parameter.modernisation_platform_account_id.value
   ]
 
   create_role       = true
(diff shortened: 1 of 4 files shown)


ewastempel (approved):
LGTM

PR #8717, opened 2024-12-10
Remove lakeformation managed policy 

dms1981:

https://github.com/ministryofjustice/modernisation-platform/actions/runs/12236290660/job/34131158803#step:7:3385

Removes the resource that attaches lakeformation permissions. For what it's worth I can't actually see how this duplication occurs, but as the policy was only attached as a convenience we're removing it.

## How has this been tested?

Tested via local plan

Deploy through CI

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #8792, opened 2024-12-19
adding check for no issue number

markgov:

we have found an issue with the new post comment feature that breaks the plan when the reusable workflow is called outside of a PR. Because there is no issue number we are seeing errors like the following

Error: Unhandled error: TypeError: Cannot read properties of undefined (reading 'number')

This PR adds in a check test to see if issue_number is blank if it is it just puts a comment in the console and skips to the end

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


.github/workflows/reusable_terraform_plan_apply.yml
```diff
@@ -126,6 +126,10 @@ jobs:
             const identifier = workflowId ? `_${workflowId}_\n` : `_${{ inputs.workflow_id }}_\n`;
             const summary = `\`${{ steps.show.outputs.summary }}\``;
             const issue_number = context.payload.pull_request.number;
+            if (!issue_number) {
```
ewastempel, line 129: I don't see how this can work. You are comparing null to boolean and I cannot see the evidence of it working, as this run has a PR number.

markgov:
I have tested this update and now when the workflow is run without an issue number it now skips adding the comment see bellow workflow run for confirmation
https://github.com/ministryofjustice/modernisation-platform/actions/runs/12649212337/job/35245131099

.github/workflows/reusable_terraform_plan_apply.yml
markgov: added a check for null

ASTRobinson (approved):

PR #8856, opened 2025-01-06
Update cooker.json

ep-93:
Reverting the nuke, discussed here - https://moj.enterprise.slack.com/archives/C013RM6MFFW/p1736155012360379


dms1981 (approved):
Approved, but is it worth confirming what the behaviour ought to be for the `nuke` key? EG. Should it be read from all groups, or is it only relevant when the level is `sandbox`?

Issue #7900, opened 2024-09-11
Limit Deployment of Automation Roles to Non-Production Accounts to Enhance Security


sukeshreddyg:
### User Story

As a Modernisation Platform Engineer
I want to ensure that automation roles are deployed only to non-production accounts where they are actively needed, so that we can avoid unnecessary role deployments in production environments. This will help mitigate the security risk associated with having unused or irrelevant roles in production accounts.

### Value / Purpose

* The presence of automation roles in production accounts creates security vulnerabilities. Limiting them to non-prod accounts will reduce exposure to these risks.
* Focusing on non-prod accounts ensures that the roles are only deployed where needed, improving operational effectiveness.
* By limiting automation roles to only non-prod accounts, we reduce the risk of unnecessary resource allocation.

### Additional Information

* The deployment of user roles will not be limited in this scope, as they are necessary for both production and non-production accounts.
* we should limit the deployment of automation roles to non-production accounts. This action aligns with our goal to improve security by ensuring that these roles are not present in production accounts, where they could be exploited.

### Definition of Done

- [x] Explore options to filter and identify the specific non-prod accounts where automation roles are actively required.
- [x] Confirm the list of roles that should be limited to non-prod accounts (specifically automation roles, excluding user roles).
- [x] Confirm with the team that the proposed solution complies with organizational security and operational standards
- [x] implement the deployment limitations.

markgov:
Low priority

dms1981:
From a consistency standpoint I don't think this is something we should be progressing. Manipulating which roles/policies are available on a per-account basis adds uncertainty when investigating account issues and moves us away from a standardised approach.

davidkelliott:
I agree, I recommend this issue be closed

ewastempel:
I think this ticket would be useful if it was focusing on the standardised approach (e.g. member accounts use only A,B,C roles, core accounts use C,DE, unrestricted - A, B, F), but I understand it’s drifting away from it and it is proposing customised approach, which I do agree would be pain to manage in the future.

ewastempel:
Saying that, @sukeshreddyg did you have security in mind when creating it? I think roles that are used e.g. by instance scheduler, or nuke, should only be deployed to non-prod and in that case we could re-write it to be more specific?

ewastempel:
[...]

PR #8938, opened 2025-01-13
Ensure instance-scheduler role is created only for non-production environments

Khatraf:

#7900 

This PR updates the Terraform configuration to ensure roles are only created for non-production environments (dev,test,pre-prod). Role is not created unnecessarily for workspaces like testing-test, production.

## How has this been tested?

Tested it locally and works as expected.

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
Looks good to me. Checked the regex pattern which will apply true if the text is not "-production".

PR #8935, opened 2025-01-13
new TGW code

markgov:
## A reference to the issue / Description of it
As Part of issue TGW Ireland Creation
[#8584](https://github.com/ministryofjustice/modernisation-platform/issues/8584)
we need to create a new PR to implement a transit gateway for the Ireland region for DR 

## How does this PR fix the problem?

This PR creates new code for the creation of the transit gateway

## How has this been tested?

Please describe the tests that you ran and provide instructions to reproduce.

{Please write here}

## Deployment Plan / Instructions

Will this deployment impact the platform and / or services on it?

{Please write here}

## Checklist (check `x` in `[ ]` of list items)

- [ ] I have performed a self-review of my own code
- [ ] All checks have passed
- [ ] I have made corresponding changes to the documentation
- [ ] Plan and discussed how it should be deployed to PROD (If needed)

## Additional comments (if any)

{Please write here}

PR #8940, opened 2025-01-14
Correction to the provider

markgov:

With the implementation of the TGW from https://github.com/ministryofjustice/modernisation-platform/pull/8935 we found the that TGW did not get deployed in the correct account 

This PR changes the Provider so that it uses the correct account and also remove the TGW that was created in the wrong account

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

PR #8968, opened 2025-01-15
tm-871: Adding dso sso aad group developer access

davekentmoj:
https://dsdmoj.atlassian.net/browse/TM-871

Enable access to Mod Platform via Entra ID for DSO .

Previously test in NOMIS Development environment.

This is parallel access so is not an impactful change.

- [X] All checks have passed
- [X] I have made corresponding changes to the documentation


mikereiddigital (approved):
lg2m

Issue #8674, opened 2024-12-04
Instance Scheduler - Ensuring the skip-list of excluded accounts is current.


mikereiddigital:
### User Story

As a Modernisation Platform Engineer 
I need to ensure that the list of accounts that are excluded from the instance-scheduler jobs is current and up to date.
So that we avoid accounts either erroneously included or excluded from both the production list and that list used by unit tests.

The purpose of this story is to investigate & implement a means by which this list can be dynamically generated at runtime by checking the relevant field(s) in the environments files and adding the account name to an array/list that is then used. 

### Value / Purpose

This will avoid non-member accounts being included that results in the unauthorised-api-calls alarm triggering & causing further slack alerts in the low-priority channel.

### Context / Background

The number of AccessDenied errors caused by the scheduler calling non-member accounts results in the unauthorised-api-calls alarm triggering & alerts sent to the low priority slack channel. We want to reduce these occurrences. 

### Additional Information

Could be worth considering this as a spike to test first the feasibility. In addition it needs to be considered whether a new field is added to the json files as there is one non-prod member account that we want excluded.

### Definition of Done

- [x] Solution identified & feasibility tested.
- [ ] Solution implemented.
- [ ] Unit Tests refactored.
- [ ] Impact of the solution tested & confirmed that AccessDenied errors are no longer occurring.

mikereiddigital:
End of week update:

- This is a [link](https://github.com/mikereiddigital/test/blob/issue/8674/github_utils/github_utils.go to a new go file that includes two new functions - 1) FetchDirectory which returns from github the json contents of the environments files, checks whether it is a "member" and if so, adds the details to an array. 2) FetchJSON which is a helper function used by FetchDirectory, and 3) extractNames which iterates through the json array and, for each account (name), gets the account type if the instance_scheduler_skip value is not true.

Worth nothing that instead of it being a "skip" list, this is an "include", so the relevant getNonProductionAccounts function will be amended- see [here](https://github.com/ministryofjustice/modernisation-platform-instance-scheduler/blob/a9418e84910e682616e7f6c587b78b3ce7028b17/instance-scheduler/utils.go#L55C6-L55C30)

- The branch instance-scheduler-skip in modernisation-platform contains an amended cooker.json that includes the instance-scheduler-skip key.

[...]

PR #8971, opened 2025-01-15
Issue 8674 - Runbook for Excluding accounts from Instance Scheduler.

mikereiddigital:

https://github.com/ministryofjustice/modernisation-platform/issues/8674

Adds a runbook for excluding accounts from the instance-scheduler.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}


source/index.html.md.erb
```diff
@@ -145,6 +145,7 @@ This documentation is for anyone interested in the Modernisation Platform and it
 - [Revoking User Access](runbooks/revoking-user-access.html)
 - [Security Monitoring](runbooks/security-monitoring.html)
 - [Security Testing and ITHC](user-guide/security-testing-and-ithc.html)
+- [Excluding accounts from the Instance Scheduler](runbooks/instance-scheduler-excluding-accounts.html)
```
connormaglynn, line 148: May be worth consolidating this documentation with https://user-guide.modernisation-platform.service.justice.gov.uk/concepts/environments/instance-scheduling.html 😁 

It would also make sense for the documentation to live under the `Shared Services and Tools` section, I think of Runbooks as more for internal MP use rather than customer facing 📖

source/runbooks/instance-scheduler-excluding-accounts.html.md.erb
```diff
@@ -0,0 +1,88 @@
...
+
+- Non-member accounts which do not have the cross-account role & policies added during deployment. Any attempt by the scheduler to access these accounts would fail with an AccessDenied error and as such they are excluded.
+
+- Production member accounts. These are out of the scope of the scheduler to avoid loss of service availability.
+
+- Those non-production member accounts that users have requested are skipped by the scheduler.
+
+## How Instance Scheduler Determines which accounts are in scope
+
+Previously, the Instance Scheduler used a variable containing all of the account names that were to be skipped, including non-member accounts. That method has been changed so instead of a "skip list", this information is obtained form the /environments/*.json files in the Modernisation Platform repository.
```
connormaglynn, line 31: It is probably just worth stating what it currently does - if we need users to take some action, we should probable inform them via the update channel 🆙

source/index.html.md.erb
mikereiddigital: Thanks @connormaglynn I'll do that.

connormaglynn (approved):
🚀 Looks Good To Me!

PR #8993, opened 2025-01-17
Move RAM association job to reusable workflow and switch secrets source to AWS Secrets Manager

sukeshreddyg:

This PR refactors the current workflow by moving the `member-account-ram-association` job to a reusable workflow. In addition, it updates the source from which secrets are fetched: instead of using GitHub Secrets, we now fetch the secrets directly from AWS Secrets Manager. This change improves security and scalability by keeping sensitive secrets centralized in AWS.

**Centralized Secrets Management**: Previously, In workflows, secrets were fetched from GitHub Secrets. Now, secrets are fetched from AWS Secrets Manager, ensuring better security practices and centralization of secrets management.
**Reusable Workflow**: By moving the `member-account-ram-association` job to a reusable workflow, we enable better reuse and maintainability of the job across different workflows. The job now only runs when required, controlled by the `run_ram_association` input.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/reusable_terraform_plan_apply.yml
+++ b/.github/workflows/reusable_terraform_plan_apply.yml
@@ -24,31 +24,48 @@ on:
         description: 'Unique ID for the calling workflow'
         required: false
         type: string
+      run_ram_association:
+        description: "Whether to run the RAM association job"
+        required: false
+        default: false
+        type: boolean
     secrets:
-      modernisation_platform_environments:
+      MODERNISATION_PLATFORM_ACCOUNT_NUMBER:
         required: true
-      SLACK_WEBHOOK_URL:
+      PASSPHRASE:
         required: true
-      pagerduty_token:
-        required: false
-      pagerduty_userapi_token:
-        required: false
-      gh_workflow_token: 
-        required: false
 
 env:
-  ENVIRONMENT_MANAGEMENT: "${{ secrets.modernisation_platform_environments }}"
-  TF_VAR_github_token: "${{ secrets.gh_workflow_token }}"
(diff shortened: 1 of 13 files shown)


.github/workflows/reusable_terraform_plan_apply.yml
```diff
@@ -148,4 +165,50 @@ jobs:
           payload: |
             {"blocks":[{"type": "section","text": {"type": "mrkdwn","text": ":no_entry: Failed GitHub Action:"}},{"type": "section","fields":[{"type": "mrkdwn","text": "*Workflow:*\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|${{ github.workflow }}>"},{"type": "mrkdwn","text": "*Job:*\n${{ github.job }}"},{"type": "mrkdwn","text": "*Repo:*\n${{ github.repository }}"}]}]}
         env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_URL: ${{ env.SLACK_WEBHOOK_URL }}
+  member-account-ram-association:
```
ewastempel, line 169: Minor comment about jobs in the reusable workflow. If you extract common code from a number of workflow (in this case member-account-ram-association), but not all workflows, maybe it is OK for that code to be a separate reusable workflow, leaving the original reusable workflow only for the common to all. I know it is possible to feed inputs and turning it on and off that way, but the reusable workflow becomes not common anymore and it may be more difficult to read it, when it becomes a one monolithic reusable workflow.

PR #9032, opened 2025-01-22
add hosting-migrations team to their environments

sobostion:
We used to use the hmmps-migration team but have had a team restructure and want to use the hosting-migrations team now

It adds the hosting-migrations team to the envs we need access to

## How has this been tested?

The built in github checks are the only ones being used here

No as it is only adding access to a new team

- [x] All checks have passed

### DIFF
--- a/environments/delius-core.json
+++ b/environments/delius-core.json
@@ -12,6 +12,12 @@
           "github_action_reviewer": "true"
         },
         {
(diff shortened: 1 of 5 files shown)


dms1981 (approved):
LGTM - but could you populate the PR notes please?

PR #8917, opened 2025-01-10
Enable R53 DNS Firewall in ALERT mode

richgreen-moj:
## A reference to the issue / Description of it

https://github.com/ministryofjustice/modernisation-platform/issues/7474

## How does this PR fix the problem?

This PR adds a new module `r53-dns-firewall` which creates the following resources per VPC:

-  a R53 Resolver Firewall Rule Group 
- a custom list of allowed and blocked domains which can be defined via locals
- R53 Resolver Firewall Rules in the following priority:
  **1** - An allow rule for the custom allowed domains list
  **2-5** - a set of rules that alert any domains that match the [AWS-managed threat lists](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver-dns-firewall-managed-domain-lists.html)
  **6** - A Block rule for the custom blocked domains list
- Associates the rule group to the VPC
- R53 Resolver query logging config to be associated with a cloudwatch log group
- Cloudwatch alarm so any matching domains that the firewall acts on trigger a dedicated SNS topic which is hooked into PagerDuty (this is currently subscribed to the PagerDuty Core Alerts service - associated with the `#modernisation-patform-low-priority-alarms` Slack channel)

I've then called the module in the `core-vpc` code and have added a rule to ALERT on any matches to the AWS-managed lists (https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver-dns-firewall-managed-domain-lists.html), this should mean any matches are alerted on (not yet blocked as I'm imagining a period of a sprint or so where we deploy to prod on alert mode only before updating to BLOCK at a later date after a careful review of the logs).

I've also demonstrated how the custom allow/block domain lists can be populated to block or allow certain domains - with some random dummy values for now. 

## How has this been tested?

Please describe the tests that you ran and provide instructions to reproduce.

I've tested this in the garden and house sandbox VPCs by running the code locally and then running `nslookup` queries for the blocked domains on an ec2 instance in the VPCs. This has proved that the rules and logging/alerting are working.

## Deployment Plan / Instructions

Will this deployment impact the platform and / or services on it?

Not initially as we will leave it set to ALERT only but I plan to write a follow on issue to review the logs in a few weeks and then all being well turn on blocking the domains in the AWS managed lists.

## Checklist (check `x` in `[ ]` of list items)

- [X] I have performed a self-review of my own code
- [X] All checks have passed
- [ ] I have made corresponding changes to the documentation
[...]

Issue #7474, opened 2024-07-10
🔒️ Implement AWS Route53 Resolver DNS Firewall


dms1981:
### User Story

As a platform customer
I want to be better protected against potential malware situations 
So that my account is better protected

### Value / Purpose

As part of issue #6121 we looked at the NCSC PDNS. While we decided that it was not right for us, we can get an equivalent level of protection with AWS Route53 Resolver DNS Firewall.

This AWS service allows us to implement filtering of DNS requests against managed blocklists which should result in us being better secured against potential malware infections.

### Useful Contacts

@dms1981

### Additional Information

I would assume that this would be implemented through the same code we use to set up the Route53 Resolver logging.

I also assume we'd want to include alerting if a malware domain triggered the Route53 Resolver DNS Firewall.

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver-dns-firewall.html

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_resolver_firewall_config

### Definition of Done

- [x] Implementation discussed with team
- [x] Implemented & tested in sandbox account
- [x] Implemented across customer VPCs
- [x] Implemented across core VPCs
- [ ] Documentation updated

dms1981:
Do we need to tie this into any alerting - is it possible to retrieve an event and send it to our high priority alerts channel as part of this implementation?

ep-93:
Ive been looking into this, only ticket left so I shall start to tackle it, but probably will call in some packet wrangler experts to help

ep-93:
Was off sick for 2 days, and the day before focused on our security incident. Will start this ticket today

richgreen-moj:
I've been reading into the Route 53 Resolver DNS Firewall AWS docs and also completed the [aws-simulearn-dns-query-filtering](https://explore.skillbuilder.aws/learn/course/external/view/elearning/20066/aws-simulearn-dns-query-filtering)  AWS skill builder workshop which was really useful to see it in action. 

Next steps are to start drafting some code for implementation in the `core-vpc` codebase. I hope to be able to test this out in the sandpit VPC first so as not to cause any potential disruption.

richgreen-moj:
https://github.com/ministryofjustice/modernisation-platform/pull/8917 enables R53 DNS Firewall in ALERT mode for the AWS Managed domian lists.

[...]

PR #9049, opened 2025-01-23
Associate non-prod VPCs with cloudwatch rqlc-only and r53 dns firewall monitoring

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/7474

Having already enabled r53 DNS firewall for all MP VPCs in https://github.com/ministryofjustice/modernisation-platform/pull/8917 , I had to rollback the query logging/monitoring config as in https://github.com/ministryofjustice/modernisation-platform/pull/9041 as it conflicted with the existing logging setup in `core-logging`.

In this PR I have done two things:

1. Associated all dev/test/preprod VPCs with the cloudwatch r53 query logging config (shared from `core-logging`) so that we can capture when the r53 dns firewall captures traffic in the logs.
2. In `core-logging` I've created a Cloudwatch metric filter/alarm so any matching domains that the firewall acts on triggers a dedicated SNS topic which is hooked into PagerDuty (this is currently subscribed to the PagerDuty Core Alerts service - associated with the `#modernisation-platform-low-priority-alarms` Slack channel)

## How has this been tested?

Local Terraform plan only.

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-logging/logging.tf
+++ b/terraform/environments/core-logging/logging.tf
@@ -1,6 +1,6 @@
 locals {
   resolver_query_log_config_names = toset(["core-logging-rlq-cloudwatch", "core-logging-rlq-s3"])
-  vpc_ids                         = { for key, value in module.vpc : key => value["vpc_id"] if key == "live_data" }
+  vpc_ids                         = { for key, value in module.vpc : key => value["vpc_id"] if key == "live_data" || key == "non_live_data" }
   rlq_ids                         = { for name, config in data.aws_route53_resolver_query_log_config.core_logging : name => config.id }
   vpc_rlq_associations = merge([
     for vpc_key, vpc_id in local.vpc_ids : {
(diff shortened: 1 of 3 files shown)


terraform/environments/core-logging/logging.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   resolver_query_log_config_names = toset(["core-logging-rlq-cloudwatch", "core-logging-rlq-s3"])
-  vpc_ids                         = { for key, value in module.vpc : key => value["vpc_id"] if key == "live_data" }
```
dms1981, line 3: This feels superfluous; the `if` statement is meant to only extract the `live_data` VPC. You can remove it entirely if you want to capture both VPC IDs in the local value.

dms1981 (approved):

Issue #8345, opened 2024-10-25
Replace Terraform DynamoDB state locking with native state locking


dms1981:
### User Story

As a Modernisation Platform Engineer
I want to implement Terraform native state locking
So that I don't need to use DynamoDB for state locking

### Value / Purpose

In [Terraform 1.10.*](https://github.com/hashicorp/terraform/releases#:~:text=backend/s3%3A%20The%20s3%20backend%20now%20supports%20S3%20native%20state%20locking.%20When%20used%20with%20DynamoDB%2Dbased%20locking%2C%20locks%20will%20be%20acquired%20from%20both%20sources.%20In%20a%20future%20minor%20release%20of%20Terraform%20the%20DynamoDB%20locking%20mechanism%20and%20associated%20arguments%20will%20be%20deprecated.%20(%2335661)) Hashicorp will implement native state locking that is not reliant on DynamoDB tables.

The present use of DynamoDB for state locking introduces friction for our customers when stale state locks persist and require removal. Further to this we have to assume a role in the MP account in order to access the DynamoDB table.

By using native state locking we can reduce the amount of friction customers get by needing us to remove stale state locks, and we can simplify both our terraform configuration and our guidance on running local plans by removing the steps needed to assume a role in the main MP account during `terraform init`

### Definition of Done

- [x] Terraform native state locking tested
- [x] [Modernisation Platform](https://github.com/ministryofjustice/modernisation-platform) terraform updated to use native state locking
- [x] [Modernisation Platform Environments](https://github.com/ministryofjustice/modernisation-platform-environments) terraform updated to use native state locking
- [x] Confirm all statefiles using state locking (#8789 )
- [x] Old configuration for state locking removed
- [x] Documentation updated

SimonPPledger:
Believe this is now available

dms1981:
Prior issues on state locking for reference:
* https://github.com/ministryofjustice/modernisation-platform/issues/2467
* https://github.com/ministryofjustice/modernisation-platform/issues/5534

PR #9052, opened 2025-01-23
Allow Sprinkler to place native lockfiles

dms1981:

#8345 

Amends the S3 bucket policy so that Sprinkler can place native terraform locks into S3.

The bucket policy allows Sprinkler slightly broader wildcard permissions, but this is in line with other roles.

## How has this been tested?

Tested through CI pipeline

Deploy through CI

- [x] All checks have passed

{Please write here}


Kudzai-moj (approved):
LGTM

Issue #8960, opened 2025-01-15
Delete Unused Workspaces When Default Workspace is in Use


sukeshreddyg:
### User Story

As a Modernisation Platform Engineer
I want to ensure that only the `default` workspace is in use and any unused workspaces are deleted, to avoid confusion and prevent selecting workspaces that aren't being utilized.

### Value / Purpose

This will help streamline workspace management by eliminating unused or inactive workspaces. It ensures that resources are allocated to the correct workspace, improving clarity and preventing the selection of the wrong workspace.

### Context / Background

It has been observed that there is a workspace related to `github-resources` which is not being used. Resources have been created in the `default` workspace instead of the unused workspace, leading to potential confusion and inefficiencies. Deleting unused workspaces when the default workspace is being used will help prevent such issues.

### Definition of Done

- [x] All unused workspaces are identified and deleted.
- [x] If the default workspace is in use, no other workspace will be available for selection, ensuring users only use the default one.

Khatraf:
<img width="522" alt="Image" src="https://github.com/user-attachments/assets/a572531f-e4d0-41e8-ad52-e69ae9916359" />

these are the unused workspaces that I will delete.

Khatraf:
These are the workspace that I have deleted:

- modernisation-platform-account (terraform/modernisation-platform-account)
- github-resources (terraform/github)
- development (terraform/environments)

As the pipelines uses the default workspace, the above workspaces were deleted to reduces the potential for confusion or conflicts in state files.

mikereiddigital:
I'm happy this ticket has been completed. Moving to done.

PR #9109, opened 2025-01-28
update adr for TF workspaces

Khatraf:

#8960 - Updates the exceptions to include directories that are using the default workspace.


mikereiddigital (approved):
lg2m

PR #9121, opened 2025-01-30
Remove dynamodb table and legacy role for state access

dms1981:

#8345

Prior to this PR we offered a role in the Modernisation Platform repository for S3 state locking.

Our previous use of state locking required access to a DynamoDB table in order to place and release state locks. To access this DynamoDB table our users assumed a role in the same account as the table.

Now that we have implemented native state locking we no longer require this DynamoDB table, nor do we require the role that users would assume to access it.

This PR will require any customers who have not done so already to either delete their `.terraform` directory locally if they conduct local plans of their work, or alternative `terraform init --reconfigure` if appropriate.

## How has this been tested?

Tested through CI, tested locally with `developer` credentials to a small number of accounts in [modernisationplatform-environments](https://github.com/ministryofjustice/modernisation-platform-environments).

Deploy through CI

Potential impact to customers who do not undertake a deletion of any local `.terraform` directory created with legacy configuration to assume the now-deleted role.

- [x] All checks have passed

https://developer.hashicorp.com/terraform/cli/commands/init#backend-initialization


richgreen-moj (approved):
LGTM 🔥

PR #9145, opened 2025-01-31
Update template workflow to support components in environments

sukeshreddyg:
Updated the template workflow to incorporate reusable workflows that support both component-based and non-component environments. This change eliminates the need for maintaining two separate workflow versions, simplifying workflow management and ensuring consistency across environments. With this update, the workflow dynamically adapts based on whether components exist, reducing duplication and making future maintenance easier. The template is used in the environments repo when creating accounts, ensuring that all newly created accounts follow a unified and scalable workflow structure.

### DIFF
--- a/.github/workflows/templates/workflow-template.yml
+++ b/.github/workflows/templates/workflow-template.yml
@@ -32,7 +32,7 @@ permissions:
 
 jobs:
   strategy:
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

PR #9279, opened 2025-02-14
remove secret from reusable-signed-commit.yml

ASTRobinson:
Removing GitHub token as GitHub automatically provides secrets.GITHUB_TOKEN, so defining it inside the workflow_call is causing an error.

image


dms1981 (approved):
LGTM

PR #9353, opened 2025-02-24
Update Sandbox Nuke Exclude

ASTRobinson:
add exclude to sandbox for AP-ingestion following slack issue here: https://mojdt.slack.com/archives/C01A7QK5VM1/p1740393955485749

### DIFF
--- a/environments/analytical-platform-ingestion.json
+++ b/environments/analytical-platform-ingestion.json
@@ -14,7 +14,8 @@
         },
         {
           "sso_group_name": "analytical-platform-engineers",
-          "level": "sandbox"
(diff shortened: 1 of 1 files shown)


ewastempel (commented):
I believe this will sometimes work and sometimes it will not work. It is to do with which access item is read last and if it has `nuke` value specified.

I suggest to add `nuke` key for all roles, so also include the following:
```{
          "sso_group_name": "data-platform-audit-and-security",
          "level": "security-audit"
          "nuke": "exclude"
        }

ewastempel (approved):
LGTM

PR #9355, opened 2025-02-24
🤝 Analytical Platform Engineering access to Modernisation Platform Terraform state

jacobwoffenden:

We have administrative-like permission in our account, but we'd like to use the existing Terraform to make proper use of it.

Gives users to `platform-engineer-admin` role in Analytical Platform accounts access to the state location for Analytical Platform

## How has this been tested?

It hasn't, but is based on similar logic in this policy, and something Analytical Platform did for their Data Engineering colleagues ([source](https://github.com/ministryofjustice/analytical-platform/blob/main/terraform/aws/analytical-platform-management-production/terraform-state/s3-bucket.tf#L48-L76))

It's additive so shouldn't break anything, but is giving Analytical Platform access to somewhere only Modernisation Platform can write to

- [ ] All checks have passed

I'm sure there is probably a better way to reference our accounts 🤔 Maybe its worth moving our accounts into a child OU so we can reference that instead


ep-93 (approved):

ep-93:
Had a meeting yesterday where this was approved by Ewa who had some questions on it.

Issue #8605, opened 2024-11-26
Enable S3 Malware Protection for MP/Core Buckets and set up alerting


Khatraf:
### User Story

As a DevOps engineer
I want to identify the S3 buckets in our accounts that should have malware protection enabled and set up appropriate alerting
So that we ensure critical buckets are protected, and potential malware findings are promptly flagged for review.

### Value / Purpose

Automated malware detection- as it automatically scans objects uploaded to S3 buckets for malicious files, reducing the need for manual checks and interventions.
Real-Time Threat Detection - allowing teams to respond to malware or compromised files as soon as they are uploaded.

### Context / Background

Following on #8050, this issue is to identify and prioritise S3 buckets across our accounts to enable GuardDuty S3 malware protection and set up alerting for when Guardduty flags an object as malicious. AP already has alerting set up, so we can look at how they have achieved this.

### Definition of Done

- [x] S3 buckets across our accounts is reviewed and prioritised for enabling malware protection.
- [x] Malware protection is successfully enabled for the identified buckets.
- [x] Alerting is set up
- [x] Documentation is updated

Khatraf:
A potential bucket to implement this for could be mod-platform-image-artefact-bucket20230203091453221500000001 bucket in core-shared-services-production as it allows collaborator to upload artifacts.

Khatraf:
Created a new guardduty S3 malware protection role for member and core accounts.

- I tested enabling s3 malware protection in example using the role and it worked as expected : https://github.com/ministryofjustice/modernisation-platform-environments/actions/runs/13697622758/job/38303713799?pr=9857#step:9:20
- I've got a [PR](https://github.com/ministryofjustice/modernisation-platform/pull/9487) to enable S3 malware protection for buckets in core-shared services to help keep our stored data secure and prevent potential threats.

Next steps: I've messaged members who are referencing the memberinfrastructureaccess role to use GuardDutyS3MalwareProtectionRole, so I plan remove "malware-protection-plan.guardduty.amazonaws.com" from the trust policy. I'll also need to update the documentation to reference this new role.
Also, need to set up alerting.

Khatraf:
Alerts have been set up, and can be seen in our low priority channel here: https://mojdt.slack.com/archives/C02PFCG8M1R/p1741873738137539
Document has been updated to reflect new role created for enabling malware protection.

mikereiddigital:
I'm happy this ticket has been completed.

PR #9487, opened 2025-03-06
set up s3 malware protection 

Khatraf:

#8605 

This PR introduces GuardDuty S3 malware protection for S3 buckets in core-shared services to improve security and protect against threats.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


terraform/environments/core-shared-services/s3_malware_protection.tf
```diff
@@ -0,0 +1,43 @@
+# Variable to allow bucket selection
```
dms1981, line 1: Heya 👋 Why the use of a variable and a local here? Surely you could get the same from having a simple...
```
for_each = toset(aws_s3_bucket.example.id,aws_s3_bucket.another-example.id)
```

Khatraf: I chose the variable to allow for dynamic configuration, to enable users to specify bucket names flexibly both tf managed and manually created buckets.

Khatraf: for this case, your suggestion would probably work too

dms1981: I'd say that passing the values in directly to the `for_each` with `toset()` reduces the amount of complexity and is preferable - could you try that approach?

terraform/environments/core-shared-services/s3_malware_protection.tf
```diff
@@ -0,0 +1,29 @@
+# GuardDuty Malware Protection Plan
+resource "aws_guardduty_malware_protection_plan" "malware_protection_plan" {
+  for_each = toset([
+    "mod-platform-image-artefact-bucket20230203091453221500000001",
```
dms1981, line 4: Almost there! You're better off using the attributes of the Terraform resources than hardcoding their names as these would be subject to change.
`module.s3-bucket.bucket.id` for the image artefact bucket, and `module.s3-software-bucket.bucket.id` ought to be the correct references if I read the [module outputs](https://github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket/blob/main/outputs.tf) correctly.

ewastempel (approved):

Issue #8670, opened 2024-12-04
Refactor `Terraform: Github resources`


dms1981:
### User Story

As a Modernisation Platform Engineer
I want to refactor how we create GitHub resources
So that we can speed up our Terraform runs

### Value / Purpose

Our `Terraform: Github Resources` action takes around 10 minutes in total to run without changes. With changes, we have seen runs take upwards of 45 minutes. By reducing this we can improve our responsiveness to request for changes, updates, or deletions of Github resources.

Is there something about how our Terraform is structured that results in this slowness? For example, are there ways that we can retrieve data to speed things up? From a cursory look at a recent action the [contributor-access module](https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/github/modules/contributor/main.tf) appears to be the most-called resource during workflow runs.

### Context / Background

A [recent incident](https://status.modernisation-platform.service.justice.gov.uk/posts/details/PNH27HA) exposed the slowness of how we provision Github resources through Terraform.

### Definition of Done

- [x] Assess `terraform/github` code for refactoring opportunities
- [x] If appropriate, refactor code to improve runtime

dms1981:
We can use the [github_repository_collaborators](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_collaborators) as a replacement. It's a more recent addition than our current use of `github_team_repository` however the replacement is **authoritative** so I'll need to confirm how we add collaborators and ensure that they are given permissions through this new resource.

dms1981:
Ready for review! A bare plan now takes around 2 minutes, down from 10. An apply ought to be significantly faster too; the migration across to this resource was slowed down by api throttling, but following updates were more focused and faster.

PR #9508, opened 2025-03-10
Refactor GitHub repo permissions

dms1981:

#8670 

Replaces the use of `github_team_repository` with `github_repository_collaborators`.

To go into a little more detail, our original approach of applying permissions through the `github_team_repository` resource scaled poorly, requiring ~1600 or so API calls to iterate through our repositories and teams. The use of `all-org-members` allows all members of the Ministry of Justice GitHub Organization to contribute to our repositories directly, while this approach requires ~25 API calls.

The local value used to construct a map of teams/users and permissions ensures that we maintain the association of teams to the `modernisation-platform-environments` for the purposes of approving GitHub environment deployments.

This should dramatically reduce the time taken to apply permissions when changes are required, as well as removing our reliance on an API call to retrieve our list of repositories.

We did [discuss this on Slack](https://mojdt.slack.com/archives/C013RM6MFFW/p1741613916943099). We generally prefer code that's dynamic, but given the importance of a consistent list of repositories to apply permissions to and the use of statically defined lists elsewhere in this code we agreed that a static list of repositories was worthwhile. It should be easy enough to revisit.

## How has this been tested?

Tested through CI pipeline

We'll need to exercise care over approving this PR - it will take ~45 minutes to remove all permissions as the old resources are removed before the new permissions are applied with the new resource. As such, we'll likely want to advise our customers when we plan to do this - perhaps outside of normal office hours?

- [ ] All checks have passed

{Please write here}


terraform/github/locals.tf
```diff
@@ -95,11 +105,49 @@ locals {
     ]))
   )
 
-  # Create a list of repositories that we want our customers to be able to contribute to
+  # Define a list of repositories that we want to apply permissions to
   modernisation_platform_repositories = [
-    for s in data.github_repositories.modernisation-platform-repositories.names : s if startswith(s, "modernisation-platform-")
+    "modernisation-platform",
```
richgreen-moj, line 110: Is there a way we could get this list from the outputs of `github/modules/repository` ?
I guess we would need to come here and manually edit this list when adding/removing repos in future?

dms1981: The way that the module is used in on a per-repo basis, so yes I could replace these calls with something like `module.modernisation-platform.repository.name` if you'd prefer!

richgreen-moj: Yeah although just thinking that would not get around the need to feed this list each time we update repos. 

🤔 it might be a bit much but in theory we could change the way we feed the repository module, i.e. create a map of repo configs and use `for_each` to feed the module and then get the outputs in a single list to feed this. This would all just be for the benefit though of not having to update the code in multiple places when we manage repos - which we could document.

richgreen-moj (approved):
Content to try this out.

richgreen-moj (approved):
👍

PR #9519, opened 2025-03-11
Prevent GitHub Terraform from running during periods of degraded GitHub status

dms1981:

#8670

Adds a simple `curl` / `jq` evaluation that will set an exit code to stop the workflow when a query of the GitHub status api returns a result for incidents other than `none`.

## How has this been tested?

Tested locally to evaluate output

Deploy through CI

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #9551, opened 2025-03-14
TM-946: add secrets and SG in prep for HMPP domain file share

drobinson-moj:

Background see https://github.com/ministryofjustice/modernisation-platform/issues/5970

For business objects reporting systems (both nomis and oasys), a promotion management process is used to promote changes from preproduction into production. For example, configuration from preproduction is saved onto multiple file(s)  using a preprod web interface, and subsequently uploaded to production also using a prod web interface.

Putting file shares into the usual member environments won't work for this as the environments are network isolated, and a S3 based solution will be too cumbersome for the end users. A file share which sits alongside the DCs is the proposed solution. This can be mounted by jump servers in both preproduction and production.  A separate files share will be used for dev and test.

This PR is prep and just creates the required secrets and security groups.
The PR also adds some missing tags.

This is first of 2 PRs. This creates necessary secrets and SGs, before the second PR will stand up the file shares.

## How has this been tested?

Can't really be tested other than trying it, but we are adding new resources so not likely to break anything. FSX joining a self-hosted domain has been tested already in member accounts.

No

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -1216,6 +1216,8 @@ resource "aws_security_group_rule" "ad_fixngo" {
 }
 
 resource "aws_secretsmanager_secret" "ad_fixngo" {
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #9565, opened 2025-03-17
TM-946: add fileshare for preprod-prod hmpp AD

drobinson-moj:

Background see https://github.com/ministryofjustice/modernisation-platform/issues/5970
Follow up to https://github.com/ministryofjustice/modernisation-platform/pull/9551 and https://github.com/ministryofjustice/modernisation-platform/pull/9553

For business objects reporting systems (both nomis and oasys), a promotion management process is used to promote changes from preproduction into production. For example, configuration from preproduction is saved onto multiple file(s) using a preprod web interface, and subsequently uploaded to production also using a prod web interface.

Putting file shares into the usual member environments won't work for this as the environments are network isolated, and a S3 based solution will be too cumbersome for the end users. A file share which sits alongside the DCs is the proposed solution. This can be mounted by jump servers in both preproduction and production. A separate files share will be used for dev and test.

This PR adds the preprod-prod fileshare for azure.hmpp.root domain.

Adds a file share that can be access by both preproduction and production HMPPS accounts

## How has this been tested?

The devtest file share, stood up in previous PR, is working OK.

No

- [ ] All checks have passed

See slack thread on Friday re 1st PR https://mojdt.slack.com/archives/C01A7QK5VM1/p1741963642307969

### DIFF
--- a/terraform/environments/core-shared-services/ad-fixngo.tf
+++ b/terraform/environments/core-shared-services/ad-fixngo.tf
@@ -323,7 +323,7 @@ locals {
           module.ad_fixngo_ip_addresses.azure_fixngo_ip.PCMCW0012,
         ]
         ad_domain_name                      = "azure.hmpp.root"
-        ad_file_system_administrators_group = "Domain Join"
+        ad_file_system_administrators_group = "AWS FSx Admins"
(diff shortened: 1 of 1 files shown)


Kudzai-moj (approved):
LGTM

PR #9688, opened 2025-04-02
:dependabot: Update Generate Dependabot File script

ASTRobinson:
This PR updates the `generate-dependabot-file` script to use `directories` rather than `directory` as well as grouping the GitHub Actions dependencies into a single PR helping to reduce the number of :dependabot: reviews 

Script tested locally and produces the expected dependabot file successfully


Khatraf (approved):
LGTM

Issue #9556, opened 2025-03-17
Automate Pruning of Stale Branches


sukeshreddyg:
### User Story

As a Modernisation Platform Engineer, 
I want to automate the process of identifying and removing stale branches so that I can save time and ensure repository cleanliness without manual intervention.

Do for all repositories managed by MP team (excluding AMI and Config Mgt and Environments)

### Value / Purpose

- Improve repository maintenance efficiency by automating the removal of stale branches.
- Reduce manual effort and human error in identifying and deleting stale branches.
- Keep repositories clean and organised, ensuring only active and relevant branches are retained.

### Context / Background

At present, stale branches are removed manually, which is time-consuming and error-prone. With multiple repositories, it becomes challenging to consistently identify and remove branches that are no longer in use. Automating this process will ensure consistency, save time, and reduce the risk of overlooking stale branches.

### Definition of Done

- [x] The time period for considering a branch stale is agreed upon and configured
- [x] An automation job is set up to identify and prune stale branches across all repositories.
- [x] The automation job runs on a scheduled basis
- [x] Raise a ticket to cover this for all other environments

Kudzai-moj:
**Repositories**


- [ ] [modernisation-platform-terraform-bastion-linux](https://github.com/ministryofjustice/modernisation-platform-terraform-bastion-linux)
- [ ] [modernisation-platform-terraform-ecs-cluster](https://github.com/ministryofjustice/modernisation-platform-terraform-ecs-cluster)
- [ ] [modernisation-platform-terraform-s3-bucket](https://github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket)
- [ ] [modernisation-platform-terraform-aws-vm-import](https://github.com/ministryofjustice/modernisation-platform-terraform-aws-vm-import)
- [ ] [modernisation-platform-terraform-pagerduty-integration](https://github.com/ministryofjustice/modernisation-platform-terraform-pagerduty-integration)
- [ ] [modernisation-platform-terraform-loadbalancer](https://github.com/ministryofjustice/modernisation-platform-terraform-loadbalancer)
- [ ] [modernisation-platform-terraform-ssm-patching](https://github.com/ministryofjustice/modernisation-platform-terraform-ssm-patching)
- [ ] [modernisation-platform-terraform-ec2-instance](https://github.com/ministryofjustice/modernisation-platform-terraform-ec2-instance)
[...]

PR #9699, opened 2025-04-03
Feature 9556 automating pruning of stale branches

Kudzai-moj:

[#9556](https://github.com/ministryofjustice/modernisation-platform/issues/9556)

This PR is exactly the same as the below. My commits are not signing at moment. Need to merge this and then fix the signed commits.

https://github.com/ministryofjustice/modernisation-platform/pull/9724

### DIFF
--- a/.github/workflows/prune-stale-branches.yml
+++ b/.github/workflows/prune-stale-branches.yml
@@ -1,0 +1,113 @@
+name: Cleanup of Stale Branches (Deletes if branch and commits are 2+ months old)
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  cleanup:
+    name: Cleanup Stale Branches in Modernisation Platform Repos
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Ensure CLI tools are installed
+        run: |
+          command -v gh >/dev/null || (echo "❌ GitHub CLI (gh) is not installed" && exit 1)
+          command -v jq >/dev/null || (echo "❌ jq is not installed" && exit 1)
+
+      - name: Authenticate GitHub CLI
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
+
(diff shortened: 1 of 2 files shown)


.github/workflows/prune-stale-branches.yml
```diff
@@ -0,0 +1,113 @@
...
+
+    steps:
+      - name: Ensure CLI tools are installed
+        run: |
+          command -v gh >/dev/null || (echo "❌ GitHub CLI (gh) is not installed" && exit 1)
+          command -v jq >/dev/null || (echo "❌ jq is not installed" && exit 1)
+
+      - name: Authenticate GitHub CLI
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
```
richgreen-moj, line 25: This might not work when performing actions in other repos, you might need to use one of the tokens we have with wider permissions.

.github/workflows/prune-stale-branches.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Cleanup stale branches across repositories
+        run: |
+          set -eo pipefail
+          DEBUG=false
+
+          repos=(
```
richgreen-moj, line 32: Perhaps to start with you could test this with a single repo (unless you have already). Also I wonder if there is a programmatic way to find the repos rather than hard-coding? I know they might not change much but it saves issues in future if we add or archive repos. 
Does the script handle if the repo has been archived for instance?
You could definitely use the github api to find the list of repos with a filter. Perhaps ensure that t definitely ignores the ones you don't want to manage.

Kudzai-moj: i made the request changes, please see the output below,  from dry https://github.com/ministryofjustice/modernisation-platform/actions/runs/14283268538/job/40035296815

richgreen-moj: Not by design more likely just we haven't kept that list up to date. You could add them if you like?

Kudzai-moj: Thank you, ok, i will add them.

richgreen-moj (dismissed):
Nice work 👍

PR #9766, opened 2025-04-09
 Allow all branches in OIDC trust relationship for temporary workaround

sukeshreddyg:
The existing trust policy was too restrictive, expecting the `sub` claim to match only the exact branch `refs/heads/main`. However, the `sub` claim now includes additional fields (e.g., `workflow`, `job_workflow_ref`, etc.), which caused role assumption failures.

(https://mojdt.slack.com/archives/C013RM6MFFW/p1744198891546489?thread_ts=1744196862.994729&cid=C013RM6MFFW)

- The original issue stemmed from the fact that the sub claim in the OIDC token now includes additional workflow-related metadata, causing role assumption failures when the policy expected a specific branch format (refs/heads/main).

- By allowing all branches (*) in the IAM trust policy, the fix ensures that role assumption will succeed regardless of the extended sub claim format, effectively bypassing the issue in the short term.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

PR #9787, opened 2025-04-11
Permissions change temp - Do not merge 

ep-93:
Discussed with Ewa and Simon and Rich. 

Only Ewa to merge if she agrees to it.


ewastempel (approved):
Enabling sandbox access in preprod as a one time off and for a short period of time only (agreed with the team). This is going to be reverted after the time is agreed.

Issue #9571, opened 2025-03-18
Address Invalid Attribute Combination warning in Bastion & S3 Bucket TF Modules


richgreen-moj:
### Expected Behavior

No warnings in Terraform when deploying the bastion module.

### Actual Behavior

The following warning is displayed when deploying the bastion module:

```
╷
│ Warning: Invalid Attribute Combination
│ 
│   with module.baseline.module.bastion_linux.module.s3-bucket.aws_s3_bucket_lifecycle_configuration.replication,
│   on .terraform/modules/baseline.bastion_linux.s3-bucket/replication.tf line 62, in resource "aws_s3_bucket_lifecycle_configuration" "replication":
│   62:   rule {
│ 
│ No attribute specified when one (and only one) of [rule[0].prefix.<.filter]
│ is required
│ 
│ This will be an error in a future version of the provider
│ 
│ (and 3 more similar warnings elsewhere)
```

This was raised to our attention in this [slack thread](https://mojdt.slack.com/archives/C01A7QK5VM1/p1742295631879579)

This may also be affecting other module such as the patching module so we should check them all.

### Steps to Reproduce the Problem

This can be seen if you look at any recent run of the terratest job:
https://github.com/ministryofjustice/modernisation-platform-terraform-bastion-linux/actions/workflows/go-terratest.yml

It's also occurring in the s3 bucket module.. see this [thread](https://mojdt.slack.com/archives/C013RM6MFFW/p1742921071741119?thread_ts=1742918010.088809&cid=C013RM6MFFW)

richgreen-moj:
I added a fix to the module in https://github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket/pull/703

The `aws_s3_bucket_lifecycle_configuration.replication` resource had no filter block so this adds a block 
```
filter {
    prefix = ""
}
```
which covers the full contents of the bucket as a default setup.

This has resolved the deprecation warning. The new version of the module is [`v8.2.1`](https://github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket/releases/tag/v8.2.1)

richgreen-moj:
I've updated the reference to the s3 module in the bastion linux module https://github.com/ministryofjustice/modernisation-platform-terraform-bastion-linux/releases/tag/v4.4.2

[...]

PR #9788, opened 2025-04-11
Bump core-logging s3 modules to `v8.2.1`

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/9571

- Bumps s3 modules in `core-logging` to `v8.2.1` - doing this as dependabot doesn't seem to trigger PRs when pinning to commit SHAs. This included adding `prefix ""` to the lifecycle rules specified.

- Also I've fixed an issue with the athena lambda role which had an inline policy which is deprecated. This now has seperate policy defined and attachment.

- I've also added extra inputs to the `module.s3-bucket-cloudtrail` resource as when updating the module past v8.2.0 some changes meant that the logging bucket was being marked for deletion. This now doesn't show as a deletion in the plan.

- There is still a warning relating to the chatbot module which can be fixed in another PR.

## How has this been tested?

See checks below for TF plan.

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-logging/athena.tf
+++ b/terraform/environments/core-logging/athena.tf
@@ -10,7 +10,7 @@ data "aws_kms_alias" "environment_management" {
 #S3 Bucket for Athena temp SQL queries 
 
 module "s3-bucket-athena" {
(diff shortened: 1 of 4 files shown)


Khatraf (approved):
LGTM

Issue #9676, opened 2025-04-01
`terraform-member-environment` Workflow Fails to Assume `github-actions-apply` Role While Other Workflows Succeed


sukeshreddyg:
### Expected Behavior

All workflows in the repository using the IAM role `github-actions-apply` should:

Successfully assume the role via GitHub OIDC.

Execute Terraform deployments without authentication errors.

### Actual Behavior

- The [terraform-member-environment.yml ](https://github.com/ministryofjustice/modernisation-platform/blob/main/.github/workflows/terraform-member-environment.yml)workflow fails when trying to assume the `github-actions-apply` role.
- All Other workflows in MP repository succeed with the same role.

**Error observed:**

https://github.com/ministryofjustice/modernisation-platform/actions/runs/14196158364/job/39771714676#step:4:13

### Steps to Reproduce the Problem

1. Trigger the `terraform-member-environment.yml` workflow (via push to main or manually).
2. Observe the AWS credential step failing during role assumption.

sukeshreddyg:
This issue has been resolved. The member environments workflow can now assume the role in the preprod and prod environments.

https://github.com/ministryofjustice/modernisation-platform/actions/runs/14215736433/job/40564713929

PR #9801, opened 2025-04-14
Remove unnecessary role assumption permissions from `Plan OIDC` role

sukeshreddyg:

This PR removes the role assumption permissions from the Plan OIDC role. The Plan OIDC role was previously assuming roles that were unnecessary for its functionality, as it only needs to read the state file and match new resources against the existing ones. This change will help reduce unnecessary permissions and improve security. #9676 

By removing the role assumption permissions, we ensure that the Plan OIDC role no longer has access to assume roles that it does not require. This improves the overall security posture and simplifies the role configuration.

## How has this been tested?

Yes, I have tested this change on several workflows running at the branch level. All workflows executed successfully without any errors related to role assumption or permission issues. 

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modernisation-platform-account/iam.tf
+++ b/terraform/modernisation-platform-account/iam.tf
@@ -218,27 +218,6 @@ data "aws_iam_policy_document" "oidc_assume_plan_role_member" {
   }
 
   statement {
-    sid    = "AssumeRole"
-    effect = "Allow"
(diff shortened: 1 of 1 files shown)


ewastempel (approved):
nice one Sukesh

PR #9895, opened 2025-04-30
Member Access Permissions - Config:*

ep-93:

https://moj.enterprise.slack.com/archives/C01A7QK5VM1/p1746016006240269

Should add required permissions


mikereiddigital (approved):
lg2m. thanks for picking this up.

Issue #9700, opened 2025-04-03
signed-commits action missing files when used by new-member-environments


mikereiddigital:
### Expected Behavior

There should not be any files omitted from the signed commit created for the new-member-environment-files.yml workflow. However the last few runs have been missing files. 

### Actual Behavior

As evidence [here](https://github.com/ministryofjustice/modernisation-platform/actions/runs/14173956354/job/39704169161#step:6:51) the signed-commits action is omitting files from the git add, resulting in new environments that don't have all the required functionality.

### Steps to Reproduce the Problem

Any full run of the new environments workflow since 6th March 2025 could have this error.

### Modules

https://github.com/ministryofjustice/modernisation-platform-github-actions/blob/main/signed-commit/action.yml

ASTRobinson:
New release created in GitHub-Actions to resolve bug issue:  https://github.com/ministryofjustice/modernisation-platform-github-actions/releases/tag/v3.2.6

PR created with verified `signed-commit` action containing all 15 expected files: https://github.com/ministryofjustice/modernisation-platform-environments/pull/11531

moving for review

richgreen-moj:
Nice work @ASTRobinson! Especially as it was a pest to test/fix! Moving to done 👍

PR #9907, opened 2025-05-01
Switch New-Memeber-Environmnet-Files to signed commit

ASTRobinson:
As per ticket https://github.com/ministryofjustice/modernisation-platform/issues/9700 this PR switches the `New-Memeber-Environmnet-Files` workflow back to using the signed-commit action so further investigation can be carried out.


richgreen-moj (approved):
LGTM 👍

PR #9922, opened 2025-05-01
Create Dummy Account for testing - Step 1

ASTRobinson:
This PR creates a dummy-account that i will use to run through the account creation process for testing signed-commits, ticket ref https://github.com/ministryofjustice/modernisation-platform/issues/9700


Khatraf (approved):
LGTM

PR #9945, opened 2025-05-06
Create secret for list non mp account ids

ASTRobinson:
Create a secret for the list of non Modernisation-Platform Account IDs for use on the observability platform.


Kudzai-moj (approved):
LGTM

PR #9971, opened 2025-05-07
Add secret to limied read IAM policy

ASTRobinson:
Add secret to `modernisation-account-limited-read-member-access` role IAM policy so that the observability-platform workflow can read the secret from the Modernisation-Platform account


mikereiddigital (approved):
lg2m

Issue #9624, opened 2025-03-26
Connect Junipers in `youth-justice-networking-production` to MP TGW


richgreen-moj:
### User Story

As a MP Engineer
I need to connect the Juniper SRX boxes in `youth-justice-networking-production` via VPN to the MP transit Gateway
So that the current YJAF functionality is migrated to MP

### Value / Purpose

Required for the YJAF migration to MP 

### Context / Background

Will need to work with Stephen Hobden to agree the setup.

### Definition of Done

- [x] Create VPN
- [x] Test

dms1981:
You can see examples of configuring remote-end customer gateways and VPN connections to them in these PRs: #3824, #3942 

Our expectation is that we will create two VPNs; one for each SRX, and that each SRX will advertise routes to us over BGP that we will then propagate into our transit gateway inspection route table.

ep-93:
Stuck in blocked, reached out to Steve.

richgreen-moj:
Draft PR for building VPNs to connect Junipers to TGW... https://github.com/ministryofjustice/modernisation-platform/pull/9896

richgreen-moj:
VPN is deployed and static routes have been assigned to the MP TGW route table. Steve has confirmed he can ping the endpoints of the VPN so we are confident that it is set up successfully. He is now testing the connection onwards via the TGW and into the MP-hosted YJAF environments. We are waiting to here the results of the testing to see if we need to troubleshoot or do anything further.

ep-93:
Networking side test has been completed.

PR #9974, opened 2025-05-07
Add IKE versions input and specify ike2 for yjb vpn tunnels

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/9624

Adds IKE versions input to the vpn resource and specifies `ikev2` for the YJB VPN tunnels.

## How has this been tested?

Local plans and via the checks in this PR.

I've added a condition that checks the if the IKE versions have been set in the vpn configs in `vpn_attachments.json` and if not it will fall back on a default value that specifies BOTH options i.e. `["ikev1", "ikev2"]`. 

The other VPNs show as changes required but they already show as having default config via the console i.e. both ikev1 and ikev2 enabled. This is being triggered I believe as the versions were being ignored as a lifecycle which I've also had to remove as part of this change. 

As best I can tell this won't have any impact on the existing tunnels.

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-network-services/locals.tf
+++ b/terraform/environments/core-network-services/locals.tf
@@ -56,7 +56,7 @@ locals {
   )
   firewall_sets = fileexists("./firewall-rules/sets.json") ? jsondecode(templatefile("./firewall-rules/sets.json", local.all_cidr_ranges)) : {}
 
-  vpn_attachments = fileexists("./vpn_attachments.json") ? jsondecode(file("./vpn_attachments.json")) : {}
+  vpn_attachments = fileexists("./vpn_attachments.json") ? jsondecode(file("./vpn_attachments.json")) : tomap({})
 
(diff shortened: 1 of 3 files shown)


dms1981 (approved):

dms1981 (commented):
Heya @richgreen-moj - having looked at this I think there's a simple solution. Can you try adding something like `tunnel1_ike_versions = []` to the VPN configurations that are held in JSON?

richgreen-moj:
> Heya @richgreen-moj - having looked at this I think there's a simple solution. Can you try adding something like `tunnel1_ike_versions = []` to the VPN configurations that are held in JSON?

I might ditch this one @dms1981 - now I've managed to make it not complain anymore it says no changes necessary! I think by default the VPNs have been set up to be IKE1 or IKE2 compatible. I might just check this theory with support though.

dms1981 (approved):
LGTM!

PR #9992, opened 2025-05-08
Quicksight Admin for Collab

ep-93:
Required for YJAF.

Makes the quickight-admin role available for collaborators.


markgov (approved):
LGTM

PR #9996, opened 2025-05-08
Fixes the workflow. 

ep-93:
Raises every 6 months or when triggered. 

I can edit the text as required, will show in stand up tomorrow.


richgreen-moj (approved):
LGTM

PR #10016, opened 2025-05-12
Feature/static routes only yjaf

ep-93:

Required for YJB connectivity, changing to static routes. This means less resilience, however currently that has been accepted and will be addressed in the future.

- [ ] I have checked that all places that require the static route change have had it


richgreen-moj (approved):
LGTM 👍

PR #10060, opened 2025-05-16
PagerDuty integration for Performance Hub

jemnery:

As part of our app migration to dotnet core, we'd like to move away from our current error notification using GOV.UK Notify.

Add a new PagerDuty integration

## How has this been tested?

As yet untested

No


richgreen-moj (approved):
LGTM 👍

Issue #9970, opened 2025-05-07
S3-Upload KMS permissions on oasys development / core-shared-services


ep-93:
### Expected Behavior

https://moj.enterprise.slack.com/archives/C01A7QK5VM1/p1746615278840949

Capita - using the s3-upload role, need to upload artefacts to the mod-platform-image-artefact-bucket and encrypt with the general-hmpps key, however they are only able to view keys in the oasys development account not the core-shared-services account where the general-hmpps KMS key lives. I'm not sure why they can't. 

### Actual Behavior

Contact - 

Carl Last
Morgan Boniface

### Steps to Reproduce the Problem

Ask Carl Last or Morgan to show you the issue.

sukeshreddyg:
User confirmed that they are able to upload and access objects from the bucket.

PR #10122, opened 2025-05-21
Fix cross-account S3 upload failure by adding `kms:Decrypt` permission

sukeshreddyg:
Uploading files to the S3 bucket using the AWS Console was failing with an `Access Denied` error when using a CMK from another account. This PR adds the `kms:Decrypt` permission to the IAM role used for uploading objects to S3 with a cross-account customer managed KMS key. #9970


Khatraf (approved):
lgtm

PR #10138, opened 2025-05-22
Feaat/add soa r53 zone

andywelshmoj:

Adds new *INTERNAL* DNS zone `laa-ccms-soa.service.justice.gov.uk`

Currently this application has no DNS zone and one is needed

## How has this been tested?

New resources, cannot test

{Please write here}

Deploy zone then deploy records from `modernisation-platform-environment`. Starting in dev and promoting to higher envs


andywelshmoj:
not needed

PR #10187, opened 2025-05-30
Use modernisation platform account ID for role assumption

sukeshreddyg:

This PR updates the GitHub Actions workflow to use the `modernisation_platform_account_id` instead of the AWS Organizations `root_account_id` when assuming IAM roles.

Previously, the workflow attempted to assume a role using the root account ID, which is not the correct target. This change ensures the `github-actions-apply` role is assumed in the correct (Modernisation Platform) account.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

PR #10602, opened 2025-07-03
Add AWS_REGION env and OIDC permissions to GitHub Actions workflow

sukeshreddyg:
Added `AWS_REGION` env and OIDC permissions (id-token: write, contents: read) to the workflow.


ewastempel (approved):
LGTM

PR #10725, opened 2025-07-04
New Account - add alias skip for vcms-test

ASTRobinson:
As per new account documentation steps, adding skip alias for vcms-test `Error: creating account alias with name 'vcms-test'`


mikereiddigital (approved):
lg2m

PR #10745, opened 2025-07-08
chore: widen laa-ccms-soa ecr idam

andywelshmoj:
Widens scope of iam roles allowed to push to laa-ccms-soa ecr


Kudzai-moj (approved):
LGTM

Issue #10509, opened 2025-06-24
Update MP code to use v6.0 of the TF AWS Provider


richgreen-moj:
### User Story

As a MP Engineer
I want to upgrade our code to ensure compatibility with the latest provider
So that we continue to support the codebase and ensure future compatibility

### Value / Purpose

The TF AWS provider has been updated to v6 which has some potentially breaking changes. We need to test the impact of applying this against our code in the `modernisation-platform` repo to ensure we keep the code in a compatible and supportable state moving forward.

### Context / Background

The TF AWS provider has been updated to v6 which has some potentially breaking changes. We need to test the impact of applying this against our code in the `modernisation-platform` repo. 

There are lots of PRs that have been raised by Dependabot in the MP codebase. We need to test upgrading to v6 and see the potential impact it has on our code.

### Additional Information

https://github.com/hashicorp/terraform-provider-aws/releases/tag/v6.0.0

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-6-upgrade

### Definition of Done

- [x] Test all MP codebase with v6 of the provider
- [x] Fix any incompatibility issues
- [x] Merge in PRs to upgrade

Gary-H9:
We have started work in a similar vein in [this](https://github.com/ministryofjustice/analytical-platform/issues/8065) ticket. Most notable outcomes are captured in [this](https://github.com/ministryofjustice/modernisation-platform-environments/pull/11609) PR.

richgreen-moj:
TF configs that require update:

`/terraform`
- [X] `/single-sign-on` https://github.com/ministryofjustice/modernisation-platform/pull/11052
- [x] `/pagerduty` https://github.com/ministryofjustice/modernisation-platform/pull/10911 
- [x] `/modernisation-platform-account` https://github.com/ministryofjustice/modernisation-platform/pull/10920
- [x] `/github` https://github.com/ministryofjustice/modernisation-platform/pull/10913

`/terraform/environments`
- [x] `terraform/environments` (root folder) https://github.com/ministryofjustice/modernisation-platform/pull/10932

Bootstrap
- [x] `/bootstrap/delegate-access` https://github.com/ministryofjustice/modernisation-platform/pull/10943
- [x] `/bootstrap/member-bootstrap` https://github.com/ministryofjustice/modernisation-platform/pull/10943
- [x] `/bootstrap/secure-baselines` https://github.com/ministryofjustice/modernisation-platform/pull/10943 
[...]

PR #10920, opened 2025-07-14
Update mp account code to v6 aws TF provider

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/10509

Updates the `modernisation-platform-account` code to be compatible with v6 of the aws terraform provider.

I've also migrated the code away from using the local `core-monitoring` module (as it's just a local module that calls a remote module so seems like an uneccessary extra level of obfuscation). I can remove the `moved` blocks in a follow-up PR.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/modernisation-platform-account/iam.tf
+++ b/terraform/modernisation-platform-account/iam.tf
@@ -14,7 +14,7 @@ module "collaborators_group" {
   #checkov:skip=CKV_TF_1:Module registry does not support commit hashes for versions
   #checkov:skip=CKV_TF_2:Module registry does not support tags
   source  = "terraform-aws-modules/iam/aws//modules/iam-group-with-policies"
-  version = ">= 5.0"
+  version = "~> 5.0"
   name    = "collaborators"
 
   group_users = [for user in module.collaborators : user.username]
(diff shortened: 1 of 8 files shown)


terraform/modernisation-platform-account/iam.tf
```diff
@@ -14,7 +14,7 @@ module "collaborators_group" {
   #checkov:skip=CKV_TF_1:Module registry does not support commit hashes for versions
   #checkov:skip=CKV_TF_2:Module registry does not support tags
   source  = "terraform-aws-modules/iam/aws//modules/iam-group-with-policies"
-  version = "~> 5.0"
+  version = ">= 5.0"
```
ASTRobinson, line 17: Should this stay as `~> 5.0`

richgreen-moj: So I tried that to begin with and then got the error that it couldn't satisfy the version constraint because we're trying to update to v6 and that pins it only to `v5.x`
The terraform iam modules seem to all have a version constraint of `>= 4.0` e.g. https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-group-with-policies/versions.tf#L7

PR #11247, opened 2025-08-12
Removing Ewa's digital user from PD

ewastempel:
Now that Ewa's justice user is added to the team and the schedule, deleting Ewa's digital user.

The automation may not have the permissions needed to delete the user, in which case, the clean up will be done manually.
The code still needs updating regardless though.


Khatraf (approved):
LGTM

PR #11250, opened 2025-08-12
TM-1464: dso firewall rules improvements v1

drobinson-moj:

Some DSO firewall rules are too wide, e.g. to and from Azure estate. Other DSO firewall rules can be simplified by use of IP sets and port sets. A series of PRs will tighten the existing firewall rules and simplify using IP/port sets. This will make the platform more secure, and the firewall rules will be easier to manage.

This PR:
- fixes illegal CIDRs in locals file (the base IP is incorrect for the given subnet)
- adds missing UDP netbios ports from DC port set. This set is only used by DSO rules 
- adds outbound rules for Domain Controller traffic

This will help remove the Azure ANY rule plus a series of individual DC rules in a subsequent PR.

## How has this been tested?

n/a. New rules added. No rules being removed here.

No

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/development_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/development_rules.json
@@ -273,6 +273,20 @@
     "destination_port": "53",
     "protocol": "UDP"
   },
+  "mp_hmpps_development_to_mp_ad_azure_dcs_TCP": {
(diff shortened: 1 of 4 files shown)


mikereiddigital (approved):
lg2m

PR #11354, opened 2025-08-28
Removing problematic bracket from the scheduled baseline workflow

ewastempel:
Don't tell anyone, but this was a bad review on my side and it needed fixing :)


ewastempel:
This has been already implemented, so closing the PR.

Issue #11572, opened 2025-09-22
Create accounts for CATH application


ewastempel:
### Is this an existing AWS account?

No

### Environment details

PoC for Claude capabilities running on bedrock.

### Application Name

hmcts-claude-provider

### Description of application

Building Claude capabilities to be used as part of experiment assessing multi-agent part of working.

### SSO Group Name

to be provided later (most likely entra ID)

### GitHub code owner team slug

hw469

### GitHub actions reviewer team slug

hw469

### Environments

- [x] Development
- [ ] test
- [ ] Preproduction
- [ ] Production

### Environment access level Development

developer

### application

hmcts-claude-provider

### business-unit

HMCTS

### infrastructure-support

SoftwareExcellence@justice.gov.uk

### owner

SoftwareExcellence@justice.gov.uk

### Slack Channel

rse-dev-tools

### Do you require isolated networking?

No

### How do users connect to the application

Over the public internet

### Additional features

Please check any additional features required
- [ ] Additional VPC Endpoints
- [ ] Extended DNS Zones
- [ ] Other - please specify

### Other information

This is to enable the use of Claude Code in an experiment done by HMCTS to trial multi-agent ways of working in a single squad of c. 6 engineers.

### Definition of Done

Definition of Done
- [ ] account created
- [ ] customer informed

PR #11588, opened 2025-09-23
New account for HMCTS

markgov:

as part of Create accounts for CATH application #11572

i have created new files for a new dev account for HMTS

New account

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


environments/hmcts-claude-provider.json
```diff
@@ -0,0 +1,24 @@
+{
+  "account-type": "member",
+  "codeowners": ["hw469", "hw469"],
```
ewastempel, line 3: 1. This is a list of teams, but you have listed the same name twice.
2. I also noticed that this is a username and not the team, so please reach out to the requestor for the details (it was only used as a placeholder).
3. I think at the first instance, codeowners are actually not needed for this account and the same for GH actions approvers. It was an error filling the issue, so the default team should be used (but you will need the team slug first)

environments/hmcts-claude-provider.json
```diff
@@ -0,0 +1,24 @@
...
+  "environments": [
+    {
+      "name": "development",
+      "access": [
+        {
+          "sso_group_name": "modernisation-platform",
+          "level": "developer"
+        }
```
ewastempel, line 11: Based on what was filled out in the form, the `github_action_reviewer` field is missing under the access config.

environments/hmcts-claude-provider.json
markgov: not required for now

Kudzai-moj (approved):
LGTM

PR #11664, opened 2025-10-03
bug fix for firewall rules

markgov:

Traffic from the LAA VPC (`172.20.0.0/16`) to Cloud Platform services was being dropped, despite the destination being included in the FQDN allowlist. Investigation showed that the FQDN rule group wasn’t being applied to these flows because the LAA CIDR was not listed in `fw_home_net_ips`. As a result, the traffic bypassed FQDN inspection and was caught by default layer-4 drop rules.

Additionally, the firewall policy was configured with `stateless_fragment_default_actions = ["aws:drop"]`, which caused fragmented packets to be dropped before reaching the stateful engine for FQDN evaluation.

 Added `172.20.0.0/16` to the `fw_home_net_ips` list so FQDN rules apply to flows sourced from the LAA VPC.  
- Updated firewall policy to set `stateless_fragment_default_actions = ["aws:forward_to_sfe"]`, ensuring fragments are passed to the stateful engine for inspection.  
- Reordered stateful rule groups so the FQDN allowlist is evaluated before general stateful rules, preventing broad drop rules from shadowing domain matches.

These changes ensure that traffic from the LAA VPC to allowed FQDN destinations is correctly permitted.

## How has this been tested?

{Please write here}

- Deploy through the standard Terraform pipeline.  
- This change affects the firewall policy and rule groups only; no application downtime is expected.  
- After deployment, monitor CloudWatch Network Firewall logs to confirm that traffic from the LAA VPC is allowed as expected and that no unintended traffic is dropped.  
- **Rollback plan:** revert `fw_home_net_ips` to the previous version and restore the old fragment default action if needed.

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/fqdn_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/fqdn_rules.json
@@ -38,5 +38,5 @@
     "stats.microsoft.com",
     "wustat.windows.com"
   ],
-  "fw_home_net_ips": ["10.26.0.0/16", "10.27.0.0/16"]
+  "fw_home_net_ips": ["10.26.0.0/16", "10.27.0.0/16", "172.20.0.0/16"]
 }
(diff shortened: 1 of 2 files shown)


mikereiddigital (approved):
lg2m

PR #11676, opened 2025-10-07
Fix label formatting in issue creation for workflows

richgreen-moj:
Correct label formatting in the issue creation commands for secrets rotation and collaborator deletion workflows.

The list of labels is comma seperated and included a whitespace which was being read as part of the label name.


Kudzai-moj (approved):
lgtm

Issue #11637, opened 2025-09-29
Proof of Concept - Rework Scheduled Baselines to support greater than 256 accounts


mikereiddigital:
### User Story

As a Modernisation Platform Engineer I need certainty that our scheduled baselines and other workflows that use a matrix for each account will not break once the total number of accounts in scope goes above 256.

The scope of this ticket is the next step on from https://github.com/ministryofjustice/modernisation-platform/issues/10910 with needs to address the issues raised from that ticket:

1. How do the scheduled baselines jobs behave when each is split into two or more separate matrices.
2. How do we organise the depends-on between the various jobs so that the necessary dependencies between the stages are in place to avoid latter jobs failing.

Note - this is a large piece of work and may suit having two people working on it together.

### Value / Purpose

As it stands, our scheduled baselines job will fail if the number of accounts being fed into a job matrix goes above 256. This POC will confirm we have a solution that works.

### Context / Background

See ticket https://github.com/ministryofjustice/modernisation-platform/issues/10910 for the first part of this work.

### Definition of Done

- [ ] Determine the required dependencies for scheduled baselines if each job is split into 2 or more.
- [ ] Create new reusable workflows for each of the scheduled baselines stages.
- [ ] Build a working prototype that implements the method identified in the Spike (https://github.com/ministryofjustice/modernisation-platform/issues/10910) that confirms the solution will work. Note it may be necessary to disable to current live baselines whilst this is worked on to avoid duplicate running.
- [ ] Review outcomes with team and agree any next steps- one of which will be to identify other workflows that use the same matrix.

mikereiddigital:
Added myself and @richgreen-moj.

mikereiddigital:
@richgreen-moj and I had a very useful chat on how to proceed with this ticket and the general logistics of testing etc. We plan to start tomorrow.

mikereiddigital:
@richgreen-moj and I have got it working almost in full:

- New v2 of the scheduled-baseline workflow - https://github.com/ministryofjustice/modernisation-platform/blob/11637-new-calling-workflow/.github/workflows/scheduled-baseline-2.yml

- New single reusable template for each of the four stages - https://github.com/ministryofjustice/modernisation-platform/blob/11637-reusable-branch/.github/workflows/reusable-scheduled-baselines.yml

To-Do:

1. Run through with all accounts

2. Replace aws roles used with the correct one.

3. Fix the on-error slack message that failing to send

3. General review & polish

[...]

PR #11683, opened 2025-10-07
Issue 11637 - New Scheduled Baselines & Reusable Workflow.

mikereiddigital:

#11637 

New scheduled-baseline job that supports 1000+ accounts that uses a new reusable workflow for each of the four jobs in scope. Original workflow is kept should it be required but will be disabled and the schedule & on-push triggers disabled. Amended documentation that references the older version.

Note that the Lychee test fails due to the file in question not being found as it's not in the main branch.

## How has this been tested?

This has been tested against both a limited set of non-production accounts as well as a full-run against all accounts:

- https://github.com/ministryofjustice/modernisation-platform/actions/runs/18282760117

The existing baselines workflow - https://github.com/ministryofjustice/modernisation-platform/actions/workflows/scheduled-baseline.yml - should be DISABLED first before merging this PR.

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/.github/workflows/scheduled-baseline.yml
+++ b/.github/workflows/scheduled-baseline.yml
@@ -1,16 +1,16 @@
 name: "Terraform: scheduled baseline"
 
 on:
-  schedule:
-    - cron: "30 22 * * 0"
-  push:
-    branches:
-      - main
-    paths:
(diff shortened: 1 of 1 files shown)


richgreen-moj (approved):
LGTM 👍

Issue #11061, opened 2025-07-18
Discuss SOC and XSIAM alerts for modernisation platform accounts + member accounts


ep-93:
### User Story

I'm adding this here as a request came in about alerting for member accounts with SOC and XSIAM. More info will be given from Aumeer, Yaasseen, however this is a placeholder.

### Value / Purpose

A new layer of alerting has been offered. Whilst I believe security should work out what is required for their alerts,  they wanted to ask us about application log alerting also and other things.. sorry its vague 

### Context / Background

For XSIAM and SOC logs.

### Useful Contacts

Aumeer, Yaasseen

### Definition of Done

- [ ] First outcome
- [ ] Second outcome

SimonPPledger:
Having spoken to Yaasseen he only is after  DNS hosted in core-network services - ie these 

<img width="610" height="538" alt="Image" src="https://github.com/user-attachments/assets/305f060a-a42a-43e4-8d02-51926c4281a9" />

richgreen-moj:
Explanation of benefits from security team:

While reviewing Nomis DNS logs with Dominic Robinson, we identified that the Route 53 logs currently ingested from the Modernisation Platform (and Cloud Platform) are limited to DNS queries made within VPCs. We do not currently receive logs for DNS queries to public hosted zones.
 
Admittedly, the AWS engineers have followed [documentation](https://aws.amazon.com/blogs/aws/log-your-vpc-dns-queries-with-route-53-resolver-query-logs/) diligently and the documentation is specific Route 53 Resolver Query Logs, which we are receiving at the moment. So we see the dns queries made by resources within the VPCs.
 
So from a security perspective, at the moment, we can detect if VPC-based resources are making DNS queries to potentially malicious domains or resolving them to suspicious IPs.
What we cannot detect are activities such as external/malicious IPs making DNS queries for our publicly hosted AWS resources,  related activities such as DNS DDOS or have visibility for investigative purposes
 
We wanted to check if it is feasible to collect DNS query logs for public hosted zones at a platform level please?

richgreen-moj:
I've implemented public DNS query logging for all public hosted zones in `core-network-services` following this [guide](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/query-logs.html)

All logs are being captured in a CloudWatch log group in `us-east-1` which is the mandated region for these type of logs.

[...]

PR #11693, opened 2025-10-08
🔧  Build CW Log Destination in `us-east-1`

richgreen-moj:

https://github.com/ministryofjustice/modernisation-platform/issues/11061

I'm rebuilding the cloudwatch logs destination resources in `us-east-1` as the firehose lives in that region and they need to be in the same region to link up.


Khatraf (approved):
LGTM

PR #11715, opened 2025-10-13
adding rule to allow ports from laa-shared-services to ssogen

kowusu123:
adding rule to allow ports 7002 and 7003 from laa-shared-services to ssogen


markgov:
unable to merge due to a non signed commit PR will need to be recreated

PR #11818, opened 2025-11-04
Adding George Hill to the DSO Pagerduty schedule, and removing Robert Sweetman

ghill26:
## Adding myself (George Hill) to the DSO PagerDuty schedule, and removing Robert Sweetman

This is so that I can be on the concierge rota to provide support. Additionally, this removes Robert Sweetman from the DSO rota.

This references my PagerDuty user as appropriate, and is intended to honour the existing DSO schedule. It also removes the reference to Robert Sweetman, due to Robert's account being disabled


richgreen-moj (approved):
LGTM

PR #11911, opened 2025-11-19
Change of byte size and rule order

markgov:

after looking at the code i have found that if i change the byte size to 3 this will force a name change and i also found that i needed to change thr rule groups to strict order as well

code fix's

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


markgov:
also reversed a PR in the main firewall code that added new to the policy name

PR #11949, opened 2025-11-24
Updating how we communicate changes that have org wide impact

ewastempel:
This came out of the retro actions. This is so that our communication of the AWS organisation wide changes is documented.


ewastempel:
Stale branch, starting fresh

PR #12047, opened 2025-12-02
Fix secret rotation job raising duplicate issues

richgreen-moj:

The secrets rotation reminder workflow was creating duplicate GitHub issues daily for the same secrets that needed rotation. For example, the `xsoar` secret had 12 duplicate open issues created over consecutive days.

The issue was caused by the GitHub App token being unable to use the GitHub search API (`gh issue list --search`), which meant the duplicate check was always returning empty results. 

The fix:
- Changed from using `gh issue list --search` to `gh issue list --json --jq` which uses the direct API
- Increased the limit from 100 to 500 to ensure all open issues are checked (repo currently has 177 open issues)
- Added better logging to show when secrets are past threshold but already have open issues
- I've also added `nonmp-account-ids` to the list of secrets that don't need rotating.

## How has this been tested?

- Ran the workflow manually on the `fix/duplicate-secret-reminders` branch
- Confirmed it correctly detected 12 existing open issues for `xsoar` and `nonmp-account-ids`
- Verified it did NOT create duplicate issues when existing ones were found

SEE EXAMPLE RUN: https://github.com/ministryofjustice/modernisation-platform/actions/runs/19858423274/job/56902111975#step:8:1

**NB** The check on this PR will fail as the IAM role to assume only works on `main`, I had to edit this to test it.

No deployment impact. This is a fix to the secrets rotation reminder workflow itself. The workflow will run on its normal daily schedule and will stop creating duplicate issues.

- [x] All checks have passed

The 12 duplicate issues for `xsoar` and `nonmp-account-ids` should be manually closed, keeping only one open issue for each secret.

### DIFF
--- a/.github/workflows/secrets-rotation-reminder.yml
+++ b/.github/workflows/secrets-rotation-reminder.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - "main" 
     paths:
-      - ".github/workflows/secrets-rotation-reminder.yaml"
+      - ".github/workflows/secrets-rotation-reminder.yml"
   schedule: 
       - cron: '30 11 * * *' # run at 11:30 daily 
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mikereiddigital (dismissed):
lg2m

PR #12225, opened 2026-01-07
Update check_iam_users_and_keys.sh

ep-93:
Redacts username in script. Rest of redactions done via the redact-output.sh


ep-93:
Screenshot 2026-01-07 at 18 19 04

PR #12942, opened 2026-04-16
Checkout `modernisation-platform-environments` for nuke redeploy workflow

sukeshreddyg:
`terraform apply` steps operate on Terraform code that lives in `modernisation-platform-environments`, not in this repo. Without this change, the workflow was referencing `terraform/environments/` and `scripts/` paths that don't exist in 
`modernisation-platform`, causing the steps to fail.


markgov (approved):
LGTM

PR #12990, opened 2026-04-23
Revert CNI tag for AP and Data Platform

ASTRobinson:
Reverting `critical-national-infrastructure` tag and adding temporary bypass of CNI OPA check for AP and Data environments.


richgreen-moj (approved):
LGTM

PR #12993, opened 2026-04-23
app-id has been deprecated

ASTRobinson:
Warning: Input 'app-id' has been deprecated with message: Use 'client-id' instead.

image

https://github.com/ministryofjustice/modernisation-platform/actions/runs/24833548010


mikereiddigital (approved):
lg2m

PR #13049, opened 2026-05-01
adding annesa mariyam to the dso pagerduty schedule

annesa-m:
## Adding myself (Annesa Mariyam) to the DSO PagerDuty schedule

This is so that I can be on the concierge rota to provide support.

This references my PagerDuty user as appropriate, and is intended to honour the existing DSO schedule

### DIFF
--- a/terraform/pagerduty/member-services-integrations.tf
+++ b/terraform/pagerduty/member-services-integrations.tf
@@ -1855,6 +1855,7 @@ locals {
     "antony.gowland"   = local.digital_email_suffix
     "dominic.robinson" = local.digital_email_suffix
     "george.hill2"     = local.justice_email_suffix
+    "annesa.mariyam"   = local.justice_email_suffix
   }
   # repeat users, e.g. for a 3 day stint of concierge
   dso_schedule_user_order = [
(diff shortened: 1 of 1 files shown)


terraform/pagerduty/member-services-integrations.tf
```diff
@@ -1967,8 +1971,8 @@ resource "pagerduty_schedule" "dso" {
 
   layer {
     name                         = "Primary Schedule"
-    start                        = "2025-11-04T00:00:00Z"
-    rotation_virtual_start       = "2025-11-04T00:00:00Z"
+    start                        = "2026-05-01T00:00:00Z"
```
drobinson-moj: I think you'll need 2026-05-12T00:00:00Z here - that's when Tony is next on rota. Otherwise it will put Tony on Rota starting from today

drobinson-moj (changes requested):

terraform/pagerduty/member-services-integrations.tf
annesa-m: That's been amended, thanks

drobinson-moj (approved):

ASTRobinson (approved):

Issue #12217, opened 2026-01-07
Protect Baseline Resources from Automated Nukes and Manual Deletions


sukeshreddyg:
### User Story

As an Infrastructure team member, 
I want to ensure that our baseline resources are protected from automated nukes and accidental deletions, so that the environment remains compliant and stable.

### Value / Purpose

- Prevents loss of critical baseline resources due to automated pipelines or manual actions.
- Reduces manual effort to recreate baseline resources after accidental deletion.
- Provides a repeatable and controlled approach to account cleanup and resource protection.

### Context / Background

https://mojdt.slack.com/archives/C013RM6MFFW/p1767779520191869

### Definition of Done

- [ ] Baseline resources are tagged consistently (e.g., Baseline=true).
- [ ] aws-nuke config updated to exclude baseline resources by name, type, or tag.
- [ ] SCPs reviewed and applied to protect baseline resources, with a documented break-glass procedure for account closure.
- [ ] Team members notified of the changes on how to handle exceptions.

sukeshreddyg:
AWS nuke config updated and SCP created to protect baselines

ASTRobinson:
Definition of Done completed, nuke config updated and SCP created. Moving to done.

PR #13073, opened 2026-05-06
protect AWSBackupSelection resources from aws-nuke

sukeshreddyg:
`aws_backup_selection` resources created by the [modernisation-platform-terraform-baselines](https://github.com/ministryofjustice/modernisation-platform-terraform-baselines) module are not protected by the existing `secure-baselines` tag filters in the nuke config presets.

This is because `aws_backup_selection` **does not support tagging** as a resource attribute in Terraform, making tag-based filtering in the preset impossible for this resource type.

Without this change, running aws-nuke on a member account risks deleting the platform-managed backup selections, which would break the backup strategy for non-production workloads.
#12217

### DIFF
--- a/scripts/nuke-config-template.txt
+++ b/scripts/nuke-config-template.txt
@@ -18,6 +18,7 @@ resource-types:
   # don't nuke IAM users
   excludes:
     - ACMCertificate
+    - AWSBackupSelection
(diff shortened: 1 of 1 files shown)


richgreen-moj (approved):
LGTM

PR #13086, opened 2026-05-07
fix: add get obj

matt-heery:

Adds get object back so read only can read the results of the query.

- [x] All checks have passed


markgov (approved):
LGTM

PR #13103, opened 2026-05-10
Revert "Module Update for ec2 image builder logs bucket"

Kudzai-moj:
Reverts ministryofjustice/modernisation-platform#13074

Reverted as module will be updated to deal with buckets like this which are customer-managed KMS encrypted at rest, but not request-header compliant.


richgreen-moj (approved):
LGTM

PR #13109, opened 2026-05-11
Mod Platform Terraform state bucket update for s3 module

Kudzai-moj:

https://github.com/orgs/ministryofjustice/projects/17/views/4?pane=issue&itemId=156958371&issue=ministryofjustice%7Cmodernisation-platform-security%7C102

Follow-up to: https://github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket/pull/890

This PR updates the Terraform state bucket to the latest S3 bucket module version while retaining AES256 compatibility.

During testing, it was confirmed that Terraform/OpenTofu state uploads explicitly use `AES256` rather than SSE-KMS request headers.

We are updating our existing buckets first before making a release. Once we are satisfied the changes work correctly with our buckets, we will release v10 and update remaining buckets to use v10 as well.

---

This PR updates the Terraform state S3 bucket to use the latest version of the `modernisation-platform-terraform-s3-bucket` module.

The bucket already has customer-managed KMS bucket default encryption configured, however Terraform/OpenTofu uploads explicitly request:

```json
{
  "SSE": "AES256"
}
```

This means strict SSE-KMS request-header enforcement would break Terraform/OpenTofu state uploads.

To avoid introducing a breaking change, the bucket remains configured with:

```hcl
sse_algorithm = "AES256"
```

This preserves compatibility with existing Terraform/OpenTofu backend behaviour.

---

### Updated bucket

- Terraform state bucket

---

## How has this been tested?

- Terraform plan reviewed successfully
- Bucket default encryption verified
- Terraform/OpenTofu state object uploads verified
- `head-object` testing confirmed uploaded state objects are explicitly stored using `AES256`

---

### Will this deployment impact the platform and / or services on it?

- No

This change preserves compatibility with existing Terraform/OpenTofu state uploads while upgrading the bucket to the latest module version.

---

- [x] All checks have passed

---


terraform/modernisation-platform-account/s3.tf
```diff
@@ -116,7 +116,7 @@ resource "aws_kms_alias" "s3_state_bucket_eu-west-1_replication" {
 }
 
 module "state-bucket" {
-  source = "github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket?ref=9facf9fc8f8b8e3f93ffbda822028534b9a75399" # v9.0.0
+  source = "github.com/ministryofjustice/modernisation-platform-terraform-s3-bucket?ref=c67758cd6d263bec9c225b99b6f76d6074514159"
```
ewastempel, line 119: Could you please preserve the version information.

terraform/modernisation-platform-account/s3.tf
```diff
@@ -125,6 +125,7 @@ module "state-bucket" {
   bucket_name                = "modernisation-platform-terraform-state"
   replication_bucket         = "modernisation-platform-terraform-state-replication"
   suffix_name                = "-terraform-state"
+  sse_algorithm              = "AES256"
```
richgreen-moj: Instead of this did you mean to put...
`enforce_kms_request_headers = false` .. i.e. keeping the KMS encryption with CMK, but not enforcing encrypted headers in bucket policy?

terraform/modernisation-platform-account/s3.tf
Kudzai-moj: Hi @ewastempel .We are currently testing our own buckets against the latest commit. I have not yet created a release/new version. I will do this once im happy everything works for our buckets. So this will be updated again to v10 and not a specific commit/merge

terraform/modernisation-platform-account/s3.tf
Kudzai-moj: This one is a bit awkward. The Buckets default encryption is custom KMS, but Actual Terraform/OpenTofu state objects are being uploaded/stored as AES256, so the uploader is overriding/not using the bucket default KMS path. Uploads are not using KMS at all, if we add kms it will be a breaking change

```
aws s3api head-object \
  --bucket modernisation-platform-terraform-state \
  --key "environments/terraform.tfstate" \
  --query '{SSE:ServerSideEncryption,KMS:SSEKMSKeyId,BucketKey:BucketKeyEnabled}'
```

produces

```
{
    "SSE": "AES256",
    "KMS": null,
    "BucketKey": null
}
```

richgreen-moj: That's interesting actually I've learnt something new there. Even though we have a default SSE encryption status of `aws:kms`, TF is not using that and is uploading as AES256 (AWS default).

OK so I don't think we should "downgrade" the default encryption type based on the current behaviour. In a future ticket we could explore sending the TF state with KMS like in this [guide](https://oneuptime.com/blog/post/2026-02-23-how-to-encrypt-terraform-state-with-aws-kms/view) in which case we'd want to keep it like it is, we could even possibly use the most secure default mode (enforce kms upload headers) although obvs that would need testing.

My overall point is, let's not downgrade the default encryption setting based on the behaviour today. Happy to have a chat if it doesn't make sense.

Kudzai-moj: Ahh makes sense, we can leave this on v9 and i raise a new ticket to explore this and other buckets which may need a bit more time and testing. You happy for this one to stay as v9 for now?

terraform/modernisation-platform-account/s3.tf
ewastempel: got you :)

terraform/modernisation-platform-account/s3.tf
richgreen-moj: Can we not upgrade to your new version and set `enforce_kms_request_headers = false` - that way it continues as it was, i.e. defaults to kms encryption (despite TF not actually uploading items with kms encryption)

richgreen-moj (approved):
LGTM

PR #13153, opened 2026-05-18
feat: add sts:AssumeRole permission for Redshift Serverless access in reporting-operations policy

ASTRobinson:
Enables `reporting-operations` users to assume Redshift Serverless service roles and view scheduled query history. Uses dynamic account ID resolution for multi-environment compatibility.

The corresponding trust policy on the target Redshift service role will need to be updated separately to add the reporting-operations role as a trusted principal.

(Slack request: https://mojdt.slack.com/archives/C01A7QK5VM1/p1779102227692199)


Khatraf (approved):
LGTM

PR #13164, opened 2026-05-19
Remove SLSA step

ASTRobinson:
Removing SLSA steps as devSecOps-Actions is soon to be retired as per https://mojdt.slack.com/archives/C52LPHPK8/p1778673548700149


mikereiddigital (approved):
lg2m

PR #13300, opened 2026-06-09
Handle container-platform-octo account naming in member delegation role assumptions

ASTRobinson:
This PR updates IAM OIDC assume-role policy generation for member bootstrap to treat `container-platform-octo` the same way as `cloud-platform`.

Both accounts use a different naming convention and do not require the `member-delegation-<business-unit>-<environment> role ARN built from core-vpc-<environment>`.
Without this exception, Terraform attempts to index a non-existent `core-vpc-live` key and fails with `Invalid index`.

**Changes made**
- Updated the conditional in oidc_assume_role_member to skip member-delegation role ARN creation for:
   - `cloud-platform`
   - `container-platform-octo`
- Updated the same conditional in `oidc_assume_nuke_role_member` for consistency.

**Why**
Fixes failing workflow/plan for the new `container-platform-octo` account by avoiding invalid account ID map lookup for `core-vpc-live`.

**Impact**
- Prevents Terraform failure for `container-platform-octo`.
- No behavior change for existing standard member accounts.
- Existing `cloud-platform` behavior remains unchanged.

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/iam.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/iam.tf
@@ -1086,8 +1086,8 @@ data "aws_iam_policy_document" "oidc_assume_role_member" {
     sid    = "AllowOIDCToAssumeRoles"
     effect = "Allow"
     resources = compact([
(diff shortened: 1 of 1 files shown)


Kudzai-moj (approved):
lgtm

PR #13346, opened 2026-06-12
Retag CP TGW attachment and associate with nonlive route table

davidkelliott:

https://github.com/ministryofjustice/cloud-platform/issues/8292

We want to attach the container platform cloud-development-vpc to the MP TGW so that traffic can route out of the VPC using the MP centralised egress.

Added routes and propagations for CP attachment

TGW route table association (ties the attachment to the TGW route table)

Retag the attachment (gives you a sensible name you can read in core-network-services )

Turn on route-table propagation

Return routes in the relevant egress VPC

Firewall rules in the relevant inline inspection network firewall (in the egress VPC)

## How has this been tested?

Plan is green, terratests are clear

No impact to existing services

- [ ] All checks have passed

{Please write here}


dms1981 (approved):
LGTM; covers the content of a discussion I had with some MP and CP engineers earlier today.

1. The attachment exists, but needs to be associated with a route table.
2. The route table needs to learn the route back via the attachment
3. The inline inspection firewall needs to allow the traffic out
4. The VPC with the inline inspection firewall needs to route back via the transit gateway
5. Retagging the attachment given that it's created through a RAM share is helpful for anyone working on the TGW.

PR #13351, opened 2026-06-15
Terraform state kms backend remediation

Kudzai-moj:

https://github.com/orgs/ministryofjustice/projects/17/views/4?pane=issue&itemId=187213756&issue=ministryofjustice%7Cmodernisation-platform%7C13136

This PR updates Terraform/OpenTofu backend writes for `modernisation-platform-terraform-state` to send SSE-KMS backend config instead of relying on SSE-S3/AES256 behaviour.

Compatibility mode remains enabled, so existing member uploads should continue while the real bucket is validated.

Terraform init now passes the state bucket KMS key via backend config:

`kms_key_id = arn:aws:kms:eu-west-2:<modernisation-platform-account-id>:alias/s3-state-bucket`

The state bucket KMS key policy is also updated to allow MP OU principals to use the key for Terraform state reads/writes, restricted to S3 and the state bucket encryption context.

`enforce_kms_request_headers = false` is intentionally unchanged.

## How has this been tested?

A temporary cross-account mimic test was run in `modernisation-platform-environments`:

- `example-development` mimicked the platform-owned state bucket/KMS key
- `cooker-development` mimicked a member environment writing Terraform backend state
- the example bucket enforced strict SSE-KMS request headers

The cooker canary successfully wrote state to the example-owned bucket.

Confirmed state object metadata:

`ServerSideEncryption = aws:kms`

`SSEKMSKeyId = arn:aws:kms:eu-west-2:083957762049:key/xxxxxx`

This validates the cross-account backend write pattern in a strict mimic test. The real `modernisation-platform-terraform-state` bucket still needs validation after this PR is approved and applied.

1. Deploy the KMS key policy update.
2. Run the `modernisation-platform-environments` cooker branch against the real state bucket.
3. Confirm the real state object writes use SSE-KMS.
4. Monitor member Terraform plan/apply workflows before any follow-up strict enforcement change.

`enforce_kms_request_headers = false` remains enabled, so existing uploads without explicit SSE-KMS headers should continue to work.

Expected service impact is low, but the state bucket is platform-critical.

- [x] All checks have passed


markgov (approved):
LGTM

PR #13365, opened 2026-06-16
Terraform state kms backend combined remediation

Kudzai-moj:

https://github.com/orgs/ministryofjustice/projects/17/views/4?pane=issue&itemId=187213756&issue=ministryofjustice%7Cmodernisation-platform%7C13136&reload=1

This PR prepares the `modernisation-platform-terraform-state` bucket for Terraform/OpenTofu backend state and lock writes to use the platform customer-managed KMS key.

Backend initialisation needs to change in a follow-up PR because Terraform currently writes state with SSE-S3/AES256. Passing `kms_key_id` during `terraform init` makes Terraform send SSE-KMS headers for state and lock writes.

This PR adds the required KMS permissions before changing backend initialisation.

It updates the state bucket KMS key policy to allow state bucket use from:

- principals in the Modernisation Platform OU
- existing `sprinkler-development` bootstrap GitHub Actions roles

Access is restricted to S3 in `eu-west-2` and the Terraform state bucket encryption context.

It also adds the matching KMS permissions to member GitHub Actions plan/apply roles.

## How has this been tested?

A temporary cross-account test in `modernisation-platform-environments` confirmed that a member-style writer can write Terraform state to a strict SSE-KMS bucket using a customer-managed KMS key.

Confirmed state object metadata:

`ServerSideEncryption = aws:kms`

`SSEKMSKeyId = arn:aws:kms:eu-west-2:083957762049:key/xxxxxx`

This PR only deploys the required KMS/IAM permissions. The real state bucket write test will be run after this PR is applied, when the backend init change is raised separately.

- [x] All checks have passed

This PR is the permission-preparation step. The backend initialisation change is deliberately separated to avoid requiring new KMS permissions before they are live.


mikereiddigital (approved):
LG2M.

PR #13377, opened 2026-06-18
Update transit_gateway_connections.tf

ep-93:

https://github.com/ministryofjustice/modernisation-platform/actions/runs/27758970735

A try is added


mikereiddigital (approved):
lg2m

PR #13379, opened 2026-06-18
Pass KMS key for Terraform state backend

Kudzai-moj:

https://github.com/orgs/ministryofjustice/projects/17/views/4?pane=issue&itemId=187213756&issue=ministryofjustice%7Cmodernisation-platform%7C13136&reload=1

This PR updates Terraform/OpenTofu backend initialisation so state and lock writes to `modernisation-platform-terraform-state` use the platform customer-managed KMS key.

The required KMS key policy and member role permissions were deployed in the [previous PR](https://github.com/ministryofjustice/modernisation-platform/pull/13365).

Terraform was previously initialising the S3 backend with encryption enabled, but without a `kms_key_id`, so backend writes used SSE-S3/AES256.

This PR updates `scripts/terraform-init.sh` to pass:

`kms_key_id = arn:aws:kms:eu-west-2:<modernisation-platform-account-id>:alias/s3-state-bucket`

The account ID is resolved from existing environment/secrets sources, with `TERRAFORM_STATE_KMS_KEY_ID` available as an override for testing.

## How has this been tested?

A previous cross-account mimic test confirmed a member-style writer can write Terraform state to a strict SSE-KMS bucket using a customer-managed KMS key.

After this PR is applied, the [cooker canary](https://github.com/ministryofjustice/modernisation-platform-environments/pull/17213/changes) will be run against the real `modernisation-platform-terraform-state` bucket to confirm state writes use SSE-KMS.

- [x] All checks have passed

`enforce_kms_request_headers = false` remains enabled, so this does not yet enforce strict SSE-KMS headers.
Strict request-header enforcement should only be considered after the real state bucket canary confirms SSE-KMS writes are working as expected.


mikereiddigital (approved):
lg2m

PR #13435, opened 2026-06-25
handle container-platform member-delegation skips in bootstrap IAM

ASTRobinson:
**What**
This updates the bootstrap IAM OIDC assume-role policies to treat `container-platform-*` accounts the same way as the existing `container-platform-octo` exception.

**Why**
New container platform accounts such as `container-platform-hmpps` do not follow the standard `core-vpc-*` member-delegation lookup path. Without a skip, Terraform attempts to index a non-existent `core-vpc-live` entry and fails with an `Invalid index error`.

**Changes**
- replace the hard-coded `container-platform-octo` check with a `container-platform-*` prefix check
- apply the change in both member OIDC policy documents:
  - apply role policy
  - nuke role policy
- keep the defensive `account_ids` key existence check around the `core-vpc-*` lookup

**Outcome**
This allows creation of new `container-platform-*` accounts without breaking bootstrap IAM policy generation, while preserving the existing behaviour for `cloud-platform` and the already-created `container-platform-octo`.

### DIFF
--- a/terraform/environments/bootstrap/member-bootstrap/iam.tf
+++ b/terraform/environments/bootstrap/member-bootstrap/iam.tf
@@ -1100,8 +1100,8 @@ data "aws_iam_policy_document" "oidc_assume_role_member" {
     sid    = "AllowOIDCToAssumeRoles"
     effect = "Allow"
     resources = compact([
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #13498, opened 2026-07-02
Fix KMS permissions for state bucket replication role

Khatraf:

While testing Object Lock on the Terraform state replication bucket, noticed that the replication bucket had not received recent replicated state objects. The replication role already had S3 replication permissions, but it did not have explicit KMS permissions to decrypt objects from the source state bucket key and encrypt replicated objects using the destination replica KMS key.

This PR adds an inline IAM policy to the Terraform state bucket replication role.

The policy allows the replication role to:

- decrypt source state bucket objects using the source KMS key in `eu-west-2`
- use the replica KMS key in `eu-west-1` when writing replicated objects to `modernisation-platform-terraform-state-replication`

The permissions are scoped with `kms:ViaService` and S3 encryption context conditions for the source and replication bucket ARNs.

## How has this been tested?

Tested this in the Example account.

{Please write here}

- [ ] All checks have passed

{Please write here}


mikereiddigital (approved):
lg2m

Issue #13221, opened 2026-06-01
Centralise VPC endpoints


davidkelliott:
### User Story

As an MP engineer
I want traffic to VPC endpoints to be centralised
So that hosting can save money
Initially this is being being built in MP but as a separate thing for CP
### Value / Purpose

Cost savings and networking simplification

### Context / Background

Currently VPC endpoints are located in the shared VPCs, it is best practice to use VPC endpoints to access AWS services rather than going over the public internet. However they cost per endpoint, by centralising them we save money.  Although this cost is not too extreme in MP, CP are rebuilding and will have many additional VPCs, AWS estimates potential savings of up to $180K by centralising these for the CP endpoints alone. Solution should use Route 53 profiles.

### Useful Contacts

David Elliott, Ky Truong, David Sibley

### Additional Information

Architectural pattern here - https://docs.aws.amazon.com/whitepapers/latest/building-scalable-secure-multi-vpc-network-infrastructure/centralized-access-to-vpc-private-endpoints.html

### Definition of Done

- [ ] New VPC with VPC endpoints created
- [ ] Routing in place
- [ ] create new tickets to migrate existing endpoints in MP

ASTRobinson:
Latest Diagram: 

<img width="3358" height="1591" alt="Image" src="https://github.com/user-attachments/assets/6cb785c7-be51-4357-8844-4365d0b404de" />

ASTRobinson:
Progress update on centralised VPC endpoints (for CP-first implementation in MP):
- Aligned with Cloud Platform / Container Platform on the cross-account sharing approach using AWS RAM for central endpoint consumption.
- Confirmed target design direction is centralised interface endpoints + Route 53 Profiles for private DNS resolution from consuming VPCs.
- Current workstream is focused on establishing the shared networking foundations first (RAM resource sharing, associations, and DNS/routing pattern), before endpoint migration.
- Next delivery steps are:
  1. Build new central VPC hosting the initial endpoint set.
  2. Put Route 53 profile associations and routing in place for consuming VPCs.
  3. Validate CP connectivity pattern end-to-end.
  4. Raise follow-on MP migration tickets for existing shared-VPC endpoints.

ASTRobinson:
ADR plan (draft)

**Title**: Centralised VPC Endpoints with Route 53 Profiles
**Status**: Proposed

**Context**
Endpoint-per-VPC model increases fixed cost and operational overhead.
CP scale-out materially increases projected spend.
[...]

PR #13499, opened 2026-07-02
Add route53profiles:* to MemberInfrastructureAccess role

ASTRobinson:
Adds `route53profiles:*` to the member bootstrap IAM policy to allow member accounts to associate Route53 Profiles with their VPCs.

This is required for onboarding to the centralised VPC endpoints hub (#13221). The centralised-endpoints Route53 Profile is shared via RAM to consumer accounts; without this permission, AssociateProfile calls fail with AccessDeniedException.


Kudzai-moj (approved):
lgtm

PR #13504, opened 2026-07-02
Fix Config logs S3 replication KMS permissions

Khatraf:

The `modernisation-platform-logs-config` bucket was failing replication to `modernisation-platform-logs-config-replication` bucket. Other core logging buckets, such as CloudTrail and Route53 public DNS logs, already allow the replication role the wider set of KMS actions needed for encrypted replication.

This PR updates the Config logs source KMS key policy so the S3 replication role can use the same KMS action set as the working replicated logging buckets.

The policy now allows the replication role to use:

- `kms:Decrypt`
- `kms:Encrypt`
- `kms:ReEncrypt*`
- `kms:GenerateDataKey*`
- `kms:DescribeKey`

## How has this been tested?

Tested in Example account. 

{Please write here}

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-logging/logs_config_s3_kms.tf
+++ b/terraform/environments/core-logging/logs_config_s3_kms.tf
@@ -51,9 +51,15 @@ data "aws_iam_policy_document" "kms_config_logs" {
   }
 
   statement {
-    sid       = "AllowReplicationAccess"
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #13516, opened 2026-07-03
Add HMPPS and LAA accounts to centralised endpoint DNS share

ASTRobinson:
## What
Adds the HMPPS and LAA container-platform live/non-live accounts to the Route53 Profile RAM share for the centralised VPC endpoints DNS profile.

## Why
Ky requested that these additional container platform accounts receive the shared Route53 Profile so they can associate it with their VPCs and use the centralised interface endpoints.

Accounts added:
- container-platform-hmpps-nonlive
- container-platform-hmpps-live
- container-platform-laa-nonlive
- container-platform-laa-live

## Change
Updates `centralised_endpoint_consumer_accounts` in `terraform/environments/core-network-services/vpc-endpoints.tf` to include the four additional account IDs.

## Notes
This only expands the Route53 Profile RAM share principals; it does not change endpoint configuration or TGW routing.

### DIFF
--- a/terraform/environments/core-network-services/vpc-endpoints.tf
+++ b/terraform/environments/core-network-services/vpc-endpoints.tf
@@ -2,11 +2,15 @@ locals {
   centralised_endpoint_vpc_cidr = "10.20.240.0/20"
 
   centralised_endpoint_consumer_accounts = {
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #13533, opened 2026-07-06
Handle non-standard DNS names for centralised interface endpoints

ASTRobinson:
## What
Adds explicit Route53 private zone name overrides for centralised interface endpoint services that do not follow the default `<service>.<region>.amazonaws.com` pattern.

## Why
Ky's endpoint validation showed several services resolving publicly or failing DNS because the current Terraform assumes all endpoint DNS names can be derived from the endpoint service name.

Affected services:
- Detective: `api.detective.<region>.amazonaws.com`
- ECR API: `api.ecr.<region>.amazonaws.com`
- ECR Docker: `dkr.ecr.<region>.amazonaws.com`
- EKS Auth: `eks-auth.<region>.api.aws`
- Kinesis Firehose: `firehose.<region>.amazonaws.com`

Also, ECR Docker image pulls use account-specific hostnames under `*.dkr.ecr.<region>.amazonaws.com`, so a wildcard alias record is required.

## Change
In `terraform/environments/core-network-services/vpc-endpoints.tf`:
- Add `centralised_endpoint_service_private_dns_zone_overrides` map for the non-standard services above
- Apply these overrides when constructing `centralised_endpoint_service_private_dns_zones`
- Add `aws_route53_record.centralised_endpoint_wildcard_alias_records` for `*.dkr.ecr.<region>.amazonaws.com` (ECR Docker)

## Validation
- Ran `terraform fmt` on the file
- Verified override keys match configured endpoint service keys in `centralised-vpc-endpoints.json`

## Expected outcome
These services should now resolve to private `10.20.24x.x` endpoint IPs via the shared Route53 Profile and be reachable over the centralised endpoint hub.

### DIFF
--- a/terraform/environments/core-network-services/vpc-endpoints.tf
+++ b/terraform/environments/core-network-services/vpc-endpoints.tf
@@ -20,9 +20,31 @@ locals {
     replace(endpoint_name, ".", "_") => "com.amazonaws.${data.aws_region.current.region}.${endpoint_name}"
   }
 
+  centralised_endpoint_service_private_dns_zone_overrides = {
+    detective          = "api.detective.${data.aws_region.current.region}.amazonaws.com"
+    ecr_api            = "api.ecr.${data.aws_region.current.region}.amazonaws.com"
+    ecr_dkr            = "dkr.ecr.${data.aws_region.current.region}.amazonaws.com"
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

PR #13572, opened 2026-07-15
Pilot Route53 Profile-native DNS for selected interface endpoints

ASTRobinson:
## What
Implements a phased migration path to the new Route53 Profiles + interface VPC endpoint integration in `core-network-services`.

This PR introduces a pilot cohort of endpoints that now use native profile endpoint associations (with endpoint Private DNS enabled), while the remaining endpoints continue on the existing self-managed PHZ/alias model.

## Why
Ky's testing showed a set of services with non-standard DNS naming behavior requiring bespoke PHZ exceptions and wildcard records under the legacy approach.

AWS now supports associating interface VPC endpoints directly with Route53 Profiles, which can eliminate much of that custom DNS management.

We are piloting with a small initial set agreed with Ky before expanding further.

## Pilot services moved to profile-native mode
- **SQS**
- **SNS**

### `terraform/environments/core-network-services/vpc-endpoints.tf`
- Split endpoints into two cohorts:
  - `centralised_interface_endpoint_services_legacy_dns`
  - `centralised_interface_endpoint_services_profile_native`
- Keep legacy endpoints on:
  - `private_dns_enabled = false`
  - self-managed Route53 PHZ + alias records
- Move pilot endpoints to:
  - `private_dns_enabled = true`
  - direct `aws_route53profiles_resource_association` to endpoint ARNs
- Scope PHZ/alias resources to legacy cohort only.

### `terraform/environments/core-network-services/output.tf`
- Update `centralised_interface_endpoint_arns` output to merge ARNs from both endpoint cohorts.

## Safety / migration behavior
- This is a phased, non-big-bang migration — starting with just SQS and SNS.
- Legacy behavior remains for all non-pilot services.
- Pilot services can be expanded incrementally by adding to `centralised_endpoint_profile_native_service_keys`.
- If the pilot validates successfully, remaining services will be migrated in follow-up PRs.

## Validation
- Ran `terraform fmt` on changed files.
- Coordinated pilot scope with Ky (Cloud Platform team).

### DIFF
--- a/terraform/environments/core-network-services/output.tf
+++ b/terraform/environments/core-network-services/output.tf
@@ -86,10 +86,16 @@ output "centralised_endpoint_vpc_id" {
 }
 
 output "centralised_interface_endpoint_arns" {
-  value = {
-    for name, endpoint in aws_vpc_endpoint.centralised_interface_endpoints :
-    name => endpoint.arn
-  }
+  value = merge(
+    {
+      for name, endpoint in aws_vpc_endpoint.centralised_interface_endpoints :
+      name => endpoint.arn
+    },
+    {
+      for name, endpoint in aws_vpc_endpoint.centralised_interface_endpoints_profile_native :
+      name => endpoint.arn
+    }
+  )
 }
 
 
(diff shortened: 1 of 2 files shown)


Khatraf (approved):
LGTM

PR #13584, opened 2026-07-17
Refactor to single endpoint resource and roll out Route53 profile-native DNS

ASTRobinson:
## What
Implements option 1 by consolidating centralised interface endpoints into a single Terraform resource map and rolling out profile-native DNS for all configured services.

## Changes
- Replaced split endpoint resources with one `aws_vpc_endpoint.centralised_interface_endpoints` resource keyed by service
- `private_dns_enabled` is now controlled per service via `centralised_endpoint_profile_native_service_keys`
- Set `centralised_endpoint_profile_native_service_keys` to all configured endpoint services for full rollout
- Removed separate `centralised_interface_endpoints_profile_native` resource block
- Added `moved` blocks for `sns` and `sqs` so Terraform migrates state addresses without replacing those already-migrated endpoints
- Updated Route53 profile endpoint associations to reference the unified endpoint resource
- Updated `centralised_interface_endpoint_arns` output to use unified resource
- Added explicit dependency on legacy PHZ resources before endpoint creation to avoid `ConflictingDomainExists` race during replacement

## Why this avoids the previous race
Previously, legacy and native endpoints lived in separate resources, allowing Terraform to attempt native endpoint create while conflicting PHZ still existed.

With a single resource address per service:
- Terraform performs in-place replacement semantics for `private_dns_enabled` transitions
- legacy PHZ resources are torn down
- endpoint creation is ordered behind PHZ resource completion via dependency

## Validation
- `terraform fmt` run on changed files
- `terraform validate` attempted; blocked in local shell due missing AWS credentials (same as prior runs)

### DIFF
--- a/terraform/environments/core-network-services/output.tf
+++ b/terraform/environments/core-network-services/output.tf
@@ -86,16 +86,10 @@ output "centralised_endpoint_vpc_id" {
 }
 
 output "centralised_interface_endpoint_arns" {
-  value = merge(
-    {
-      for name, endpoint in aws_vpc_endpoint.centralised_interface_endpoints :
-      name => endpoint.arn
-    },
-    {
-      for name, endpoint in aws_vpc_endpoint.centralised_interface_endpoints_profile_native :
-      name => endpoint.arn
-    }
-  )
+  value = {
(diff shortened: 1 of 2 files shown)


ewastempel (approved):
LGTM

PR #13595, opened 2026-07-20
Fix: Add collaborator email input field to New Collaborator issue template

ASTRobinson:
The `Collaborator email address` field in the New Collaborator issue template was a Yes/No dropdown, which only asked whether an email could be provided rather than capturing the actual email address.

This PR replaces the dropdown with a free-text input field so requesters can enter the collaborator's email address directly when raising the issue.


mikereiddigital (approved):
lg2m

Issue #13581, opened 2026-07-17
AWS Transform as a service


SimonPPledger:
### User Story

We have a user that wants to use AWS Transform to upgrade their application. They have used it previously with great success.We want to make this available as a service.
Firstly we want to decide the best location of AWS Transform - a new account outside of MP or within Shared Services (or we might make it available in CDPT-IFS as a starting point).
Then we need to make it available - see notes below

### Value / Purpose

This has been proven to speed up the upgrade of applications - and bringing them into vendor support, improving security

### Context / Background

This is to upgrade,  initially, the cdpt-ifs app

### Useful Contacts

Ewa, Damien Wilson (who will be testing access etc)

### Additional Information

Notes from Damien:

After reaching out and discussing the requirement with [@Dean Layton-James (DeanLJ)](https://moj.enterprise.slack.com/team/U06CUHXKU9Z) and their AWS Transform team, we now have a process and a set of permissions in place to meet the requirement.

From the AWS Transform team:
1. SCP / Service Access
•	Org SCPs must not block transform:* actions
•	Org SCPs must not block codeconnections:* actions (needed for the GitHub connection)
•	AWS Transform is available in us-east-1, us-west-2, and eu-west-1 — confirm which region works under their org policies

2. IAM Permissions for Transform
•	Someone with administrator access needs to complete the one-time setup (AWS Transform console → "Get started" → create profile)
•	Users accessing the web app need transform:AccessTransformProfile on the profile resource
•	During setup you choose an auth method (IAM-only, IAM Identity Center, or third-party IdP) — this cannot be changed after setup, so decide upfront

3. IAM Permissions for GitHub Code Connection
Transform connects to repos via Amazon CodeConnections (GitHub App installation — username/password is not supported).
The IAM principal creating the connection needs the following permissions:
codeconnections:CreateConnection
codeconnections:GetConnection
codeconnections:ListConnections
codeconnections:DeleteConnection
codeconnections:GetInstallationUrl
codeconnections:GetIndividualAccessToken
codeconnections:ListInstallationTargets
codeconnections:StartOAuthHandshake
codeconnections:UpdateConnectionInstallation
codeconnections:UseConnection
Once the connection is created, the Transform service itself uses codeconnections:UseConnection to pull code during jobs.
4. Enable the Service
1.	Admin navigates to the Transform console → "Get started"
2.	Completes profile setup (choose auth method, encryption key, etc.)
[...]

PR #13702, opened 2026-08-04
Adding secret for transform that holds the URL endpoint. This is required for bounding CORS actions against the S3.

mikereiddigital:

#13581 

This adds a secrets manager resource that will hold the Transform application endpoint URL. This is required for use by CORS-related resource to limit https access requests to the bucket only from this endpoint & so avoids it being held in clear text in the repo.

## How has this been tested?

{Please write here}

{Please write here}

- [ ] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #13726, opened 2026-08-06
PLAT-2: PlatOps Rota update

ghill26:
- Remove Tony G from the schedule
- Honour current rotations
- Update escalation policy


Khatraf (approved):
LGTM

PR #13741, opened 2026-08-07
Add ssmmessages and ec2messages centralised VPC interface endpoints

ASTRobinson:
## What
Adds `ssmmessages` and `ec2messages` to the centralised VPC interface endpoint list.

## Why
The Cloud Platform team uses SSM Session Manager port forwarding for `kubectl` access to private EKS clusters. `ssm` is already a centralised endpoint, but Session Manager also requires `ssmmessages` and `ec2messages` (these carry the session data channel).

Without these, SSM session traffic for private cluster access still egresses to the public internet, and relay security groups cannot be tightened to private ranges only.

Requested by Mohammad Patel / Cloud Platform team, related to cloud-platform issue #8370.

## Change
One-line JSON change: adds two entries to `terraform/environments/core-network-services/centralised-vpc-endpoints.json`

## Expected plan
- **2 to add**: new `aws_vpc_endpoint` for `ec2messages` and `ssmmessages` (profile-native, `private_dns_enabled = true`)
- **2 to add**: new `aws_route53profiles_resource_association` for each endpoint → profile
- No destroys, no changes to existing resources


mikereiddigital (approved):
lg2m

PR #13749, opened 2026-08-10
User request - adds support for aws_vpc_endpoint_security_group_association resources

mikereiddigital:
Adds terraform to support the creation of aws_vpc_endpoint_security_group_association resources for specific shared vpcs and security group names as defined in the local.

See slack thread https://mojdt.slack.com/archives/C01A7QK5VM1/p1786115779911719

Creates the above resource based on values provided in a local. It is sufficiently extensible in that it will allow for further associations to be added for different VPCs if the security group exists in the shared vpc.

## How has this been tested?

See plan output below. Have checked the vpce and sg resource IDs for preprod and prod and these are valid.

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-vpc/vpc.tf
+++ b/terraform/environments/core-vpc/vpc.tf
@@ -415,3 +415,58 @@ module "r53_dns_firewall" {
   tags_prefix = each.key
   tags_common = local.tags
 }
+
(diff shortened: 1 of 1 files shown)


Khatraf (approved):
LGTM

PR #13751, opened 2026-08-10
Revert "User request - adds support for aws_vpc_endpoint_security_group_association resources"

mikereiddigital:
Reverts ministryofjustice/modernisation-platform#13749

PR #13755, opened 2026-08-10
Support for VPC Endpoint & Sec Group Associations.

mikereiddigital:

Follows on from user request to add a security group to an existing api-gateway endpoint.

This change creates security groups, ingress rules and associations with vpc endpoints. It is required in core-vpc as to create the association the owner must be the same for the sec group and the endpoint.

The earlier PR (https://github.com/ministryofjustice/modernisation-platform/pull/13751) failed as core-vpc doesn't have ownership of the security groups referenced.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}


ewastempel (approved):
LGTM

PR #13768, opened 2026-08-12
Adds support for associating the endpoint with a private subnet to allow access from outside of the vpc via the network firewall.

mikereiddigital:

See slack thread https://mojdt.slack.com/archives/C01A7QK5VM1/p1786464966392309

By default, endpoints are associated with the protected subnets that have no routing from outside of the vpc. This allows the endpoint in question to be associated with the private subnets that have external routing in place.

The subnet associations added do not alter the existing associations.

This will only affect the endpoints listed in the local and so by default all other endpoints are unchanged.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/core-vpc/vpc.tf
+++ b/terraform/environments/core-vpc/vpc.tf
@@ -429,22 +429,24 @@ module "r53_dns_firewall" {
 locals {
   vpc_endpoint_access = [
     {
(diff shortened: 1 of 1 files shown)


Khatraf (approved):
LGTM

PR #13771, opened 2026-08-12
Revert "Adds support for associating the endpoint with a private subnet to allow access from outside of the vpc via the network firewall."

mikereiddigital:
Reverts ministryofjustice/modernisation-platform#13768


Khatraf (approved):
LGTM

PR #13812, opened 2026-08-19
First of Two Changes to Support Endpoints with Custom Subnet Associations.

mikereiddigital:

Electronic Monitoring Data Preproduction and Production require an endpoint for their api-gateway that adds subnet associations to general-private subnets, not protected. This is the first of two changes to add that support. 

The deleted terraform will be moved a new file supporting custom endpoint & subnet associations. Deletion required so api-gateway endpoint can be removed from electronic-monitoring-data.json and so avoid the need to delete the security group for which the ENI removal can be problematic and cause the apply to timeout and error.

Once that endpoint is deleted, a following PR will add terraform to custom_endpoints.tf to add the endpoints and subnet associations.

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}


markgov (approved):
LGTM

PR #13828, opened 2026-08-20
New Tag Pair for Secure Baselines to Prevent AWSBackup role being wiped in analytical-platform sandbox accounts.,

mikereiddigital:

Added a tag to secure-baselines resources to ensure that nukes run in analytical platform sandbox accounts don't wipe MP baselines resources.

Currently the nuke job that AP runs in analytical-platform-data-engineering-sandboxa is wiping the AWSBackup role which in turn causes the scheduled baselines job to fail as it tries to delete protected backups.

Whilst AP excluded baselines resources, this role is clearly being missed so this is an additional protection to ensure the role is not removed.

Tag pair as defined here:

https://github.com/ministryofjustice/data-engineering-cleanup/blob/main/housekeeping_sandbox.md#for-new-infrastructure-use-custom-tagging

## How has this been tested?

{Please write here}

{Please write here}

- [x] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/bootstrap/secure-baselines/locals.tf
+++ b/terraform/environments/bootstrap/secure-baselines/locals.tf
@@ -18,13 +18,14 @@ locals {
     "us-east-1",    # US East (N. Virginia) (for global services)
   ]
   environments = {
-    business-unit = "Platforms"
-    service-area  = "Hosting"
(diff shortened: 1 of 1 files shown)


Kudzai-moj (approved):
lgtm

PR #13829, opened 2026-08-20
Fix Terratest to run with read-only plan role

sukeshreddyg:

[#187](https://github.com/ministryofjustice/modernisation-platform-security/issues/187)

terraform refresh is not needed because terraform plan already refreshes the infrastructure state as part of the planning process. The explicit refresh was also trying to write the refreshed state back to S3, which required s3:PutObject permission that the read-only PR plan role should not have.


Kudzai-moj (approved):
lgtm

PR #13833, opened 2026-08-20
Allow HMPPS development HTTPS access to Cloud Platform

Kudzai-moj:

https://mojdt.slack.com/archives/C01A7QK5VM1/p1787227881885269

An EC2 instance in `electronic-monitoring-data-development` cannot reach an internal Cloud Platform non-production API over HTTPS. The request times out because there is no central firewall rule allowing TCP 443 from the HMPPS development VPC to the Cloud Platform network.

Adds a central firewall `PASS` rule allowing `hmpps-development` (`10.26.24.0/21`) to reach the Cloud Platform range (`172.20.0.0/16`) on TCP 443.

The existing routing and NACL configuration already permits this traffic, so no changes to those components are required.

## How has this been tested?

See tests below

- [x] All checks have passed

### DIFF
--- a/terraform/environments/core-network-services/firewall-rules/development_rules.json
+++ b/terraform/environments/core-network-services/firewall-rules/development_rules.json
@@ -265,6 +265,13 @@
     "destination_port": "1521",
     "protocol": "TCP"
   },
+  "hmpps_development_to_cp_https": {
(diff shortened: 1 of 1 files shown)


mikereiddigital (approved):
lg2m

PR #13851, opened 2026-08-24
Add Integration Hub API hosting platform component

harshvasudev29:
## Summary

- register the new api-hosting-platform component for Integration Hub
- allow the integration-hub-api-platform repository to use the team GitHub OIDC provider
- keep the orchestration API isolated from the legacy api-platform state now used by the file-transfer component

## Why

The benefit-checker orchestration layer needs an independently scoped Modernisation Platform component and Terraform state. Reusing api-platform would conflict with the migrated file-transfer stack.

## Follow-up

After merge, the generated component files will be used by the infrastructure change in modernisation-platform-environments.

### DIFF
--- a/environments/integration-hub.json
+++ b/environments/integration-hub.json
@@ -6,6 +6,9 @@
       "name": "api-platform"
     },
     {
+      "name": "api-hosting-platform"
+    },
+    {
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

PR #13856, opened 2026-08-25
🔧 Rename Integration Hub zones to file-transfer

dms1981:
:copilot: _This pull request body was generated by GitHub Copilot_

## What is changing

The production-shaped Integration Hub service is being named `file-transfer` rather than `managed-file-transfer`. This updates the hosted-zone names and Terraform keys for the parent zone and its development, test, and preproduction zones.

The delegation module is renamed to match and continues to delegate the child zones from `file-transfer.service.justice.gov.uk`.

## Why the ordering matters

The parent hosted zone must exist before Route 53 can look it up for delegation. The explicit dependency keeps that ordering in the same apply. The delegation records then use the name servers from the child zones, so those delegations are in place before records are added to the zones by the account Terraform.

## Testing

The branch contains the Terraform-only rename and has been pushed for CI validation.

Signed-off-by: dms1981 <10881569+dms1981@users.noreply.github.com>

### DIFF
--- a/terraform/environments/core-network-services/route53.tf
+++ b/terraform/environments/core-network-services/route53.tf
@@ -11,10 +11,10 @@ locals {
     dacp                              = "divorce-section-search.service.justice.gov.uk",
     delius-jitbit                     = "jitbit.cr.probation.service.justice.gov.uk",
     equip                             = "equip.service.justice.gov.uk",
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

Issue #12794, opened 2026-03-26
Decommission shared-services-dev (aka core-sandbox)


ewastempel:
### User Story

Verify if there are no weird dependencies on this account and get rid of it.

### Value / Purpose

Cost saving, reducing tech debt, reducing security and operational risk.

### Context / Background

Although it would be good to have a test environment for our core accounts, [this account name](https://github.com/ministryofjustice/modernisation-platform/pull/155/changes) is confusing and does not follow our naming convention. Renaming the account alias is not possible either.

If we decide to have a test account for core accounts, we can use `core-shared-services-development` or create a new one.

### Useful Contacts

Ewa, Connor, Simon

### Definition of Done

- [ ] Check for references and dependencies this account may carry.
- [ ] Once you are sure it is safe to delete the account, decommission it.

ASTRobinson:
Account successfully decommissioned - moving to closed.

PR #13885, opened 2026-08-28
Remove share-services-dev environment

ASTRobinson:
Removing unused environment as per issue https://github.com/ministryofjustice/modernisation-platform/issues/12794


ewastempel (approved):
nice one!

PR #13896, opened 2026-09-01
Remove retired Core Sandbox documentation

ASTRobinson:
**Summary**
Removes stale documentation for the retired Core Sandbox AWS account.

- Removes the `shared-services-dev` / Core Sandbox entry from the sandbox and testing environments catalogue.
- Replaces the deleted account-specific secure-baselines workspace example with a generic `<workspace_name>` placeholder.


mikereiddigital (approved):
lg2m

PR #13919, opened 2026-09-04
prisons: add buckets to the AE SSO role s3 permissions

andrewc-moj:
## A reference to the issue / Description of it

Related to this PR: https://github.com/ministryofjustice/modernisation-platform/pull/13412/changes

I need the users of the AE role to be able to access the added paths.

## How does this PR fix the problem?

Users need to be able to read and write from the locations added.

## How has this been tested?

The role already exists and the s3 permission changes are minor.

## Deployment Plan / Instructions

Impacts one role used by a small set of users

## Checklist (check `x` in `[ ]` of list items)

- [X] I have performed a self-review of my own code
- [ ] All checks have passed
- [ ] I have made corresponding changes to the documentation
- [ ] Plan and discussed how it should be deployed to PROD (If needed)

## Additional comments (if any)

{Please write here}

PR #13412, opened 2026-06-23
prisons-de - remove delete permissions AE SSO role

andrewc-moj:
## A reference to the issue / Description of it

Related to this PR https://github.com/ministryofjustice/modernisation-platform/pull/12753/changes#r3442043269
Also here: https://justiceuk.sharepoint.com/:w:/s/DataEngineeringDataScience/IQDsbHjlvbCITrTwbeYm7fjqAbVmoCzac2o3WYw2rHjLcBM?e=tHO43Z

## How does this PR fix the problem?

The PR reduces the scope of delete permissions

## Deployment Plan / Instructions

The role has limited use (currently) there should be limited users or services affected.

## Checklist (check `x` in `[ ]` of list items)

- [X] I have performed a self-review of my own code
- [X] All checks have passed
- [X] I have made corresponding changes to the documentation
- [X] Plan and discussed how it should be deployed to PROD (If needed)

PR #13932, opened 2026-09-04
Ae sso role updates

andrewc-moj:

Users still can't query data using the new SSO role we created.  

I've separated out bucket and object level permissions. Separately added some KMS key permissions - tightly scoped to DPR keys (locals)

Related to these old PRs:
- https://github.com/ministryofjustice/modernisation-platform/pull/13919
- https://github.com/ministryofjustice/modernisation-platform/pull/13412
- https://github.com/ministryofjustice/modernisation-platform/pull/13334

Provides bucket and object level permissions.  Provides users with permissions to use KMS key and decrypt.

Deployment affects an SSO role which is currently not working.  Affects a small numbers of users, but this role will increase in use once we get past all these bugs!

- [ ] All checks have passed

{Please write here}

### DIFF
--- a/terraform/environments/bootstrap/single-sign-on/locals.tf
+++ b/terraform/environments/bootstrap/single-sign-on/locals.tf
@@ -20,6 +20,22 @@ locals {
     if(data.name == local.env_name)
   }
 
+  analytics_engineering_kms_keys = {
+    dpr-prod = {
+      account_name = "digital-prison-reporting-production"
+      key_id       = "a9d617dd-1399-4116-bd2a-16ad43f710c9"
+    }
+    dpr-preprod = {
+      account_name = "digital-prison-reporting-preproduction"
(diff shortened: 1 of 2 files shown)


terraform/environments/bootstrap/single-sign-on/locals.tf
```diff
@@ -20,6 +20,22 @@ locals {
     if(data.name == local.env_name)
   }
 
+  analytics_engineering_kms_keys = {
```
lalithanagarur, line 23: Is the plan that we'll add more kms keys here if other buckets/other accounts get added? I guess in a sense this analytics-engineer role isn't fully 'generic', as it needs modifying if a new AWS account wants to use it to give it the required bucket and kms key permissions?

andrewc-moj: Discussed and we'll revsit

Kudzai-moj (approved):
lgtm

PR #13954, opened 2026-09-10
Fix actionlint findings in workflows

ASTRobinson:
## What

Resolves the scheduled Code Quality `ACTION/actionlint` findings across the top-level GitHub Actions workflows.

The changes are focused on actionlint/ShellCheck remediation, including:

- quoting shell variable expansions and `GITHUB_ENV` / `GITHUB_OUTPUT` writes
- grouping repeated output redirects where ShellCheck requested it
- replacing fragile generated shell execution when assuming STS credentials
- preserving string-typed reusable workflow inputs using explicit string expressions
- removing unsupported `paths` filtering from the `merge_group` event in the OPA policy workflow
- fixing a `PR_TITLE` / `pr_title` variable mismatch in the environment request workflow

## Why

The scheduled Code Quality workflow reported a large number of `ACTION/actionlint` findings. This PR addresses that linter class separately so the wider Code Quality rollout can be reviewed and validated in smaller, clearer chunks.

## Validation

- `actionlint` against all top-level `.github/workflows/*.yml` and `.github/workflows/*.yaml`: pass
- `git diff --check`: pass

### DIFF
--- a/.github/workflows/access-keys-monitoring.yml
+++ b/.github/workflows/access-keys-monitoring.yml
@@ -82,9 +82,9 @@ jobs:
       # Step 5: Set Account Number
       - name: Set Account Number
         run: |
-          ACCOUNT_NUMBER=$(jq -r -e '.modernisation_platform_account_id' <<< $ENVIRONMENT_MANAGEMENT)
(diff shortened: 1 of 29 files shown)


markgov (approved):
LGTM

PR #13956, opened 2026-09-10
Fix Zizmor findings in workflows

ASTRobinson:
## What

Resolves the scheduled Code Quality Zizmor findings across the selected GitHub Actions workflows.

Changes include:

- moving GitHub context, matrix values, workflow inputs, and step outputs into step environment variables before shell or Python use
- moving broad workflow-level permissions down to the jobs that need them
- scoping GitHub App tokens with explicit repositories and permissions in a form compatible with MegaLinter's Zizmor v1.25.0
- deriving the DORA GitHub App repository scope from the same JSON definition files used by the DORA scripts, so new repositories are not missed when the JSON is updated
- adding `persist-credentials: false` to checkout steps
- adding a narrow `.github/zizmor.yml` exception for `self-repository`, because Zizmor recommends `$/...` syntax but the current actionlint/GitHub workflow schema rejects it
- preserving the existing local reusable workflow calls and GitHub-compatible syntax

## Validation

- `actionlint` across all top-level workflow YAML files: pass
- `git diff --check`: pass
- `uvx zizmor==1.25.0` against the 28 CI-listed workflows: `No findings to report`
- DORA repository scope generation validated with `jq`

The remaining Zizmor output is suppressed findings from the repository configuration; no active findings remain in the CI-listed workflow set.

### DIFF
--- a/.github/workflows/access-keys-monitoring.yml
+++ b/.github/workflows/access-keys-monitoring.yml
@@ -10,15 +10,15 @@ env:
     TEMPLATE_ID: "1f0f5ccc-0f67-4ee2-942f-6e48804828ea"
     EXPECTED_TEMPLATE_VERSION: "1"  # Update with your expected template version
 
-permissions:
-    id-token: write # This is required for requesting the JWT
-    contents: read # This is required for actions/checkout
-    issues: write # This is required to create issues
+permissions: {}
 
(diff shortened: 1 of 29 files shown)


markgov (approved):
LGTM

PR #13960, opened 2026-09-11
Fix secrets-rotation-reminder workflow: filter empty secret IDs

ASTRobinson:
## What
Filters out empty entries left in the `secrets` array after excluding exempt secrets in `secrets-rotation-reminder.yml`.

## Why
`secrets=("${secrets[@]/$del}")` replaces matching elements with an empty string but doesn't remove them from the array. The subsequent loop then called `aws secretsmanager list-secret-version-ids --secret-id ""` for these empty entries, failing with:
```
aws: [ERROR]: An error occurred (ParamValidation): Parameter validation failed:
Invalid length for parameter SecretId, value: 0, valid min length: 1
```

## Fix
After building the exclusion-filtered array, drop any blank elements before iterating over secrets.

### DIFF
--- a/.github/workflows/secrets-rotation-reminder.yml
+++ b/.github/workflows/secrets-rotation-reminder.yml
@@ -97,6 +97,13 @@ jobs:
             do
               secrets=("${secrets[@]/$del}")
             done
+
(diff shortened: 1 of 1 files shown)


markgov (approved):
LGTM

PR #13965, opened 2026-09-11
Fix invalid workflow: add permissions to reusable-scheduled-baselines callers

ASTRobinson:
## Summary

Fixes an invalid workflow file in `scheduled-baseline-extended-matrix.yml` that has prevented "Terraform: scheduled baseline - Extended Matrix" from starting since yesterday's Zizmor/actionlint changes.

## Why

```
The workflow is not valid. .github/workflows/scheduled-baseline-extended-matrix.yml (Line: 150, Col: 3):
Error calling workflow 'ministryofjustice/modernisation-platform/.github/workflows/reusable-scheduled-baselines.yml@...'.
The nested job 'main' is requesting 'contents: read, id-token: write', but is only allowed 'contents: none, id-token: none'.
```

`reusable-scheduled-baselines.yml` sets `permissions: {}` at the workflow level and its `main` job requests `id-token: write` and `contents: read`. All 16 caller jobs in this file (`delegate-access-1..4`, `secure-baselines-1..4`, `single-sign-on-1..4`, `member-bootstrap-1..4`) invoke that reusable workflow via `uses:` but none of them granted these permissions, so GitHub rejects the workflow before it can run.

## Changes

- Add
  ```yaml
  permissions:
    id-token: write
    contents: read
  ```
  to each of the 16 caller jobs, matching the permissions already granted on `fetch-secrets` and `setup-prerequisites` in this same file.
- No changes to job logic, `with:`/`secrets:` inputs, or the reusable workflow itself.

## Validation

- `actionlint .github/workflows/scheduled-baseline-extended-matrix.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/scheduled-baseline-extended-matrix.yml
+++ b/.github/workflows/scheduled-baseline-extended-matrix.yml
@@ -1,16 +1,17 @@
+---
 name: "Terraform: scheduled baseline - Extended Matrix"
 
-on:
+"on":
   schedule:
-     - cron: "30 22 * * 0"
+    - cron: "30 22 * * 0"
   push:
     branches:
       - main
     paths:
-       - '.github/workflows/scheduled-baseline.yml'
-       - 'terraform/environments/bootstrap/**'
-       - '!**.md'
-       - 'scripts/update-sso-permission-sets.sh'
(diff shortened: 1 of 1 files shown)


Khatraf (approved):
LGTM

PR #13989, opened 2026-09-15
feat: allow Container Platform nonlive to reach Cloud Platform on 5432

FolarinOyenuga:
Adds a PASS rule for container-platform-nonlive to cloud-platform on TCP 5432, matching the existing per-tenant rules (laa_test_to_cloud_platform_postgres, platforms_test_to_cloud_platform_pgres, hmpps_*_to_pgres_tcp).

Evidence: with routing in place on both sides, live-1 VPC flow logs show zero packets from 10.195.x while 5103 ingress ACCEPT flows from MP 10.26.x arrive in the same window. The path is identical for both sources apart from the source CIDR, and every source that currently reaches Cloud Platform has an explicit PASS rule.

Reachability remains constrained by routing (only 10.195.48.0/20 is routed) and by the target security group.

Ref: https://github.com/ministryofjustice/cloud-platform/issues/8371


FolarinOyenuga:
Why: routing is now in place both sides, but live-1 flow logs show **0** packets from `10.195.x` against **5,103** ingress ACCEPT from MP `10.26.x` in the same window. Every source that currently reaches Cloud Platform has an explicit PASS rule; `container-platform-nonlive` does not.

Two notes for review:
- Placed in `non_live_data_rules.json` since CP3 nonlive associates with the `non_live_data` TGW route table. The [user guide](https://user-guide.modernisation-platform.service.justice.gov.uk/concepts/networking/network-firewall.html#amending-rules) lists only the four env files, but this one is merged into the same rule set in `locals.tf`. Happy to move it if you prefer.
- `sid` is index-based (`index(keys(var.rules), ..........)`), so the plan shows renumbering across many rules.

Scope is temporary, for a handful of CP3 alpha participants. Reachability stays constrained by routing (`10.195.48.0/20` only) and by the target security group.

ASTRobinson (approved):

Issue #12904, opened 2026-04-14
EPIC: ⛳️ Create a Feature Policy Document to Enable Feature Flagging Capabilities of our Services


connormaglynn:
### User Story

As a Modernisation Platform Engineering / End User
I need/want/expect to be able to easily identify and toggle what Modernisation Platform services an account is opted in / out of 
So that I can quickly enable or disable services for certain accounts and easily debug issues / assess the impact of issues with a service

### Value / Purpose

As the Modernisation Platform has grown to accommodate different types of accounts, it has become clear that our current low-level implementation of some services doesn't adapt well to new requirements. The symptom of this is having multiple areas within our Terraform that rely on complex if statements to determine if a resource should be created. One example is how we create the [resources for the instance scheduler service](https://github.com/ministryofjustice/modernisation-platform/blob/main/terraform/environments/bootstrap/member-bootstrap/instance-scheduler.tf#L4-L4) i.e. `count = (local.account_data.account-type == "member" && terraform.workspace != "testing-test" && !can(regex(".*-production$", terraform.workspace))) ? 1 : 0` - ideally, this logic should be extracted outside of the Terraform so it can be applied consistently across multiple modules. 

The value of this work is to reduce the complexity of our Terraform code by extracting the complex business logic into a single policy document, which will also help us quickly understand what services we are deploying into which accounts. This structure also allows the capability of account-specific overrides of features, which can be held in the members environments.json files, rather than embedded in the Terraform code.

### Context / Background

It has been raised a few times in retrospectives and other meetings that it is becoming difficult to manage new account types and their exceptions i.e. what services we deploy into their accounts, different naming conventions (cloud-platform-live instead of the standards -production).

### Useful Contacts

@connormaglynn 

### Additional Information

We also want a more concrete way of identifying the account lifecycle, instead of relying on the account name. We should also add in a "lifecycle" attribute to the environments.json files so we can decouple account names with lifecycle stage.

We should pick one service/feature to trial this new approach with so we can assess how complex migration is and the benefits of the new approach. (instance scheduler might be a good one!)

and example of what a policy document might look like, using the lifecycle and account profile as two distinct decision drivers 👇 

```
{
  "lifecycle": {
    "sandbox": {
      "features": {
        "weekly_nuke": true,
        "instance_scheduler": true
      }
    },
    "nonprod": {
      "features": {
        "instance_scheduler": true
      }
    },
    "prod": {
      "features": {}
    }
  },
  "profile": {
[...]

PR #13993, opened 2026-09-16
Add instance_scheduler feature override to testing environment 

sukeshreddyg:

#12904

Adds feature override configuration to the testing environment's policy document, implementing the feature flagging capability defined in #12904. This allows account-specific feature toggles as per the policy document approach


markgov (approved):
LGTM

PR #13996, opened 2026-09-16
implement feature policy document with instance scheduler as pilot service

sukeshreddyg:

#12904 

This PR implements feature flagging by centralising service enablement logic into `feature-policy.json` instead of scattered Terraform conditionals. It simplifies instance scheduler resource creation from complex conditional to `count = local.feature_flags.instance_scheduler ? 1 : 0`. Proves the pattern works as pilot for migrating other services.


markgov (approved):
LGTM

